PR #1, opened 2024-06-26
Onboarding

parfeniuklits:


Makefile
```diff
@@ -31,4 +34,11 @@ clean:
 	rm -rf .mypy_cache
 	rm -rf .pytest_cache
 
-.PHONY: install install-dev quality style test test-unit test-integration test-e2e test-smoke test-sanity test-regression build clean
+
+fix:
```
markurtz: @parfeniukink this is what style is supposed to be for. Can you migrate these pieces under style so we can unify that?

Also, black and ruff we're disagreeing on how to format some edge cases around indexing lists, so let's remove the call into black completely

Makefile
```diff
@@ -1,3 +1,6 @@
+run:
```
markurtz: Let's remove this as well. Ideally, our entry point should be through python environment with a command line interface rather than going through makefile

Makefile
```diff
@@ -31,4 +34,11 @@ clean:
...
 
-.PHONY: install install-dev quality style test test-unit test-integration test-e2e test-smoke test-sanity test-regression build clean
+
+fix:
+	python -m black src tests
+	python -m isort src tests
+	python -m ruff format src tests
+
+
+.PHONY: run install install-dev quality style test test-unit test-integration test-e2e test-smoke test-sanity test-regression build clean fix
```
markurtz: I missed this originally, but we need to update the .PHONY here for what's contained in the file. It's leftover from another repo and contains unsupported functions

markurtz (changes requested):

Makefile
parfeniuklits: I was thinking that the Makefile is created for developers. I use it instead of aliases. I will remove it

Makefile
parfeniuklits: @markurtz 
My TLDR: consider using `isort+black` instead of `ruff`. Just some of my thought. But I will remove it as you wish.

## Detailed
sure, but are you sure about the ruff? I mean.. I know that it is a great tool in terms of linting / formatting and even suppose to be an LSP _( a bit worst than pyright nowadays, but anyways )_

Maybe we should consider using ruff only for linting ( since it is faster than flake8, pylint, etc ) and black for formatting. The `black` in known as uncompromising formatter and they specialized only on formatting when `ruff` is actually was designed as a linter. I just mean that the formatting is a secondary concern, which might mean it doesn't enforce formatting as strictly or comprehensively as `black`.

Another reason behind is that the `ruff` also sorts imports but `isort` makes it a bit better.

`ruff` latest release version is `0.4.10` for today and just wanted to notice that they only copied most of the code from other linters like flake8 so the core functionality is a bit stolen 😄

What about authors of all these tools?
* Python Code Quality Authority for `isort`
* Python Software Foundation for `black`

markurtz: @parfeniukink is definitely a good point to bring up. According to their docs, Ruff is supposed to have full feature parity with flake8, black, and isort. 

The big push behind ruff from my side is ideally to standardize to a single tool across our full stack and ML applications that is backed by the top libraries we use / will integrate with, which are specifically Hugging Face Transformers and FastAPI. 

If we can get black working with Ruff, I'm in favor of keeping it, but I do want to ensure the base authority is Ruff due to its performance, the increasing adoption, and the downstream effects of that choice.

PR #2, opened 2024-06-26
Multiple request rates

DaltheCow:


src/guidellm/executor/profile_generator.py
```diff
@@ -61,25 +61,24 @@ def next_profile(
         pass
 
 
-@ProfileGenerator.register_generator(ProfileGenerationModes.SINGLE)
-class SingleProfileGenerator(ProfileGenerator):
-    def __init__(self, rate: float, rate_type: str, **kwargs):
-        super().__init__(ProfileGenerationModes.SINGLE)
-        self._rate = rate
+@ProfileGenerator.register_generator(ProfileGenerationModes.MULTI)
+class MultiProfileGenerator(ProfileGenerator):
```
markurtz: Not sure if I'm a fan of the Multi naming for this, specifically for two reasons:
1. The Sweep profile runs multiple as well, so it doesn't necessarily differentiate it from that functionality in name
2. This runs single as well as multi, so ideally would like a name to encapsulate this as well

Maybe something along the lines of FixedRateProfileGenerator? Let me know if you have any further ideas towards that line of thinking

setup.py
```diff
@@ -25,6 +25,7 @@ def _setup_long_description() -> Tuple[str, str]:
         'openai',
         'requests',
         'transformers',
+        'click'
```
markurtz: Can we get click and the other base fixes broken out as a separate diff so we can land that quickly and populate it across the rest of the open diffs?

src/guidellm/executor/profile_generator.py
```diff
@@ -61,25 +61,24 @@ def next_profile(
...
     def next_profile(
         self, current_report: TextGenerationBenchmarkReport
     ) -> Optional[Profile]:
-        if self._generated:
+        if self._rate_index >= len(self._rates):
             return None
 
-        self._generated = True
-
         if self._rate_type == "constant":
```
markurtz: Can we simplify the constant and poisson since the only thing changing between these now is the load_gen_mode?

src/guidellm/executor/profile_generator.py
```diff
@@ -61,25 +61,24 @@ def next_profile(
...
+        self._rates = rate
+        self._rate_index = 0
         self._rate_type = rate_type
         self._generated = False
 
     def next_profile(
         self, current_report: TextGenerationBenchmarkReport
     ) -> Optional[Profile]:
-        if self._generated:
+        if self._rate_index >= len(self._rates):
```
markurtz: I don't believe we're incrementing the rate index anywhere, need to do that after we've grabbed the next profiles rate

src/guidellm/executor/profile_generator.py
```diff
@@ -61,25 +61,24 @@ def next_profile(
...
 
-        self._generated = True
-
         if self._rate_type == "constant":
             return Profile(
-                load_gen_mode=LoadGenerationModes.CONSTANT, load_gen_rate=self._rate
+                load_gen_mode=LoadGenerationModes.CONSTANT, load_gen_rate=self._rates[self._rate_index]
             )
 
         if self._rate_type == "synchronous":
```
markurtz: Let's ensure we have a proper error here for sync if a rate is passed in so that the caller is aware that isn't supported

src/guidellm/main.py
```diff
@@ -106,10 +107,11 @@ def main(
         raise ValueError(f"Unknown data type: {data_type}")
 
     # Create executor
+    profile_mode = rate_type if rate_type == "sweep" else "multi"
```
markurtz: Is there anyway for us to pass through the rate_type arg to the executor or better tie that in? I'm worried about potential additions later on where we are defaulting to multi in an if else based on matching the rate_type, so additions could be lost in that

markurtz (changes requested):

src/guidellm/executor/profile_generator.py
DaltheCow: Yeah I was not a fan of calling it multi as well, but I didn't have a good sense of a better name. Something like FixedRate or CustomRate seems good. Custom rate focuses too much on the fact that a user can pass values in which feels weird, fixed rate feels closer to the mark given that a list is being passed in, seems obvious what its purpose would be

src/guidellm/main.py
DaltheCow: I'm trying to understand the potential future goal of rate type:

Seems like a primary concern is deriving the correct profile_mode so we can create the right profile generator (sweep/fixed_rate), and the secondary concern is to set up the Profile with the right load_gen_mode and rate.

I doubt this addresses your concerns very well, maybe a step in the right direction, maybe not:

https://github.com/neuralmagic/guidellm/pull/2/commits/fb1ebcb55bb564c775cd8358d104779af02a1c43

src/guidellm/executor/profile_generator.py
DaltheCow: I think that was already the case between these two (they differed in load_gen_mode and had the same load_gen_rate as they still do), this is an attempt but I can't say I think it is definitely an improvement:
https://github.com/neuralmagic/guidellm/pull/2/commits/eb03706bb48e1a815405a4979777d8d66b223a98

src/guidellm/executor/profile_generator.py
```diff
@@ -61,35 +67,41 @@ def next_profile(
...
-
         if self._rate_type == "synchronous":
+            if self._generated:
+                return None
+        
+            self._generated = True
+
             return Profile(
                 load_gen_mode=LoadGenerationModes.SYNCHRONOUS, load_gen_rate=None
             )
```
DaltheCow: Part of me feels like this class is now a little encumbered handling both the synchronous and async profile generator rate types since they don't benefit from being in the same class here.

I have half a mind to split this out into a SynchronousProfileGenerator but I don't have a good sense of that is actually a good idea or only seems more convenient now. Is the synchronous rate type totally distinct like it seems? If so I think that'd be a cleaner way to handle this.

tests/unit/executor/test_executor.py
```diff
@@ -0,0 +1,70 @@
...
+  mock_backend = MagicMock(spec=Backend)
+  rate_type = "sweep"
+  profile_args = None
+  max_requests = None,
+  max_duration = None,
+  executor = Executor(mock_request_generator, mock_backend, rate_type, profile_args, max_requests, max_duration);
+  assert executor.request_generator == mock_request_generator
+  assert executor.backend == mock_backend
+  assert executor.max_requests == max_requests
+  assert executor.max_duration == max_duration
```
parfeniukink: something went wrong with indention

parfeniukink (changes requested):

src/guidellm/main.py
DaltheCow: I don't think this is going to happen anymore, everything is heavily typed out and there isn't flexibility for ambiguous types anymore, we would need to have some bigger concept of the direction or vision for rate_type I guess

parfeniukink (approved):

pyproject.toml
```diff
@@ -102,9 +102,9 @@ lint.select = ["E", "F", "W"]
...
-     "smoke: quick tests to check basic functionality",
-     "sanity: detailed tests to ensure major functions work correctly",
-     "regression: tests to ensure that new changes do not break existing functionality"
- ]
+addopts = '-s -vvv --cache-clear --cov-report=term-missing --cov --cov-fail-under=75'
+markers = [
+    "smoke: quick tests to check basic functionality",
+    "sanity: detailed tests to ensure major functions work correctly",
+    "regression: tests to ensure that new changes do not break existing functionality"
+]
```
DaltheCow: My bad leaving this in! Thanks for removing, I was just trying to make working on tests easier

PR #3, opened 2024-06-28
Refactor RequestGenerator to use threading and update test suite

markurtz:
Refactored RequestGenerator class:
- Replaced asyncio.Queue with Queue from the queue module for thread safety.
- Utilized threading for background queue population to ensure non-blocking request generation.
- Removed the start method as threading automatically starts the background task in async mode.
- Ensured that the _populate_queue method runs in a background thread to keep the queue populated.
- Implemented clean shutdown with the stop method joining the background thread. Updated unit tests:
- Added test_request_generator_sync_constructor and test_request_generator_async_constructor to verify constructor behavior.
- Added tests for __repr__ and __iter__ methods.
- Added tests to ensure create_item raises NotImplementedError if not overridden.
- Added tests to verify __iter__ calls create_item the expected number of times.

Separated test files:
- Created tests/unit/request/test_base.py for unit tests.
- Created tests/integration/request/test_base.py for integration tests.

Unit tests:
- Verified the construction of the class with different input parameters.
- Mocked AutoTokenizer for testing tokenizer initialization with both a class implementation and a string alias.
- Ensured that the __iter__ method works correctly in both sync and async modes.
- Verified that create_item is called the expected number of times.

Integration tests:
- Tested tokenizer construction with both a Hugging Face tokenizer and a string alias, ensuring the correct tokenizer is created.


src/guidellm/request/base.py
```diff
@@ -31,20 +33,29 @@ def __init__(
...
             )
-            logger.info(f"Tokenizer initialized: {self._tokenizer}")
+            logger.info("Tokenizer initialized: {}", self._tokenizer)
         else:
             self._tokenizer = None
             logger.debug("No tokenizer provided")
 
+        if self._mode == "async":
+            self._thread = threading.Thread(target=self._populate_queue)
+            self._thread.daemon = True
```
parfeniuklits: What about using 
```suggestion
            self._thread = threading.Thread(target=self._populate_queue, daemon=True)
```

tests/unit/request/test_base.py
```diff
@@ -57,3 +74,55 @@ def test_request_generator_repr():
...
+
+    items = []
+    for item in generator:
+        items.append(item)
+        if len(items) == 5:
+            break
+
+    generator.stop()
+    assert len(items) == 5
+    generator.create_item.assert_called()
```
parfeniuklits: Just out of curiosity. Do we need to test the length of the `items` list?
The statement: `if len(items) == 5` already defines that, kinda.

_tldr: the code in the test function itself already guaranty that the length will be 5 and we don't have to test it more_

I feel that it could be checked if the length of the returned items is mocked somewhere inside the generator. But this assertion looks useless to me.

What do you think?

src/guidellm/request/base.py
```diff
@@ -118,46 +129,31 @@ def create_item(self) -> TextGenerationRequest:
...
+                    item = self.create_item()
+                    self._queue.put(item, timeout=0.1)
+                    logger.debug(
+                        "Item added to queue. Current queue size: {}",
+                        self._queue.qsize(),
+                    )
+                else:
+                    time.sleep(0.1)
+            except Full:
+                continue
```
parfeniuklits: We are going to use these constants for sleep and timeout?

parfeniuklits (approved):
Great job!

tests/unit/request/test_base.py
markurtz: good call on this, this was supposed to be checking that the internal state did not change after calling generaator.stop rather than the items generated. I'll git up a fix shortly

src/guidellm/request/base.py
markurtz: Good call, I'll migrate these to constants for general usage

parfeniukink:
@markurtz Might be not related to this PR, but actually the `schedule` package also contains dependencies on `asyncio` in its `Task`

PR #7, opened 2024-07-03
Parfeniuk/tests backfill

parfeniukink:


tests/unit/backend/test_base.py
```diff
@@ -0,0 +1,59 @@
...
+def test_backend_registry():
+    """
+    Ensure that all registered classes exist in the registry
+    after they
+    """
+
+    assert Backend._registry == {
+        BackendEngine.TEST: TestBackend,
+        BackendEngine.OPENAI_SERVER: OpenAIBackend,
+    }
```
DaltheCow, line 44: I was wondering about this when I was writing some of my own test. It makes sense that the most effective way to test the registry would be to just directly check the values like this, but I thought because registry is defined as a private variable (_registry vs registry), that it was maybe not the intended way to test this? But I couldn't think of any way to test it without directly accessing the private variable. I think this is more of a philosophical quandary, curious of your thoughts. If this is the best way I'll probably do this as well.

tests/unit/backend/test_base.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def model_tokenizer(self, model: str) -> Optional[str]:
+        raise NotImplementedError
+
+
+@pytest.mark.smoke
+def test_backend_registry():
+    """
+    Ensure that all registered classes exist in the registry
+    after they
```
DaltheCow:
```suggestion
    after they are registered
```
maybe this?

parfeniukink: Ohh. Thanks! It should be: "Ensure that all registered classes exist in the Backend._registry"

tests/unit/backend/test_base.py
parfeniukink: @DaltheCow good point here!

I was preparing a huge answer on that but removed the line 😄
We had a discussion on that one with Mark and this `register` decorator will demonstrate its power a bit later.

But here is the topic: if you look for existing tests you will find that they are pretty much straightforward.
I mean the core philosophical question for me here is whether should that be tested or if we can trust Python developers that the Python `__init__` dunder methods will work perfectly in all the cases. And, we should test only the business logic.

So yeah it is more about: are data models ( _aka entities, values objects, aggregates, whatever_ ) should be tested in general.

Alacritty 2024-07-04


Probably @markurtz may take a hand for both of us here 😁

parfeniukink: @DaltheCow btw, check `/tests/unit/backend/test_openai.py`

tests/unit/backend/test_openai.py
```diff
@@ -0,0 +1,78 @@
...
+from guidellm.request.test import TestRequestGenerator
+
+
+@pytest.mark.sanity
+def test_openai_backend_creation_with_default_model(openai_backend_factory: Callable):
+    """
+    Test whether the OpenAI Backend service is created correctly
+    with all default parameters.
+
+    Also checks wheather the `default_models` parameter does not abuse the OpenAI API.
```
DaltheCow:
```suggestion
    Also checks whether the `default_models` parameter does not abuse the OpenAI API.
```

tests/unit/backend/test_base.py
DaltheCow: Yeah makes sense, I guess to me those tests are pretty low cost to maintain, but also a bit silly to write in the first place. But still I don't mind them.

src/domain/executor/profile_generator.py
```diff
@@ -0,0 +1,152 @@
+import functools
```
DaltheCow, line 1: Just as a call out, my PR has inflight work on the profile generator that needs review. I also am considering the notion of creating another Profile Generator class and separating out the synchronous case from constant and poisson, since they are pretty exclusive. It makes a bit more sense in the context of the multiple rates that can be passed in since we iterate through the profiles based off of how many rates are passed in which doesn't really align with the synchronous use case. I don't understand if there are any other considerations that would impact that decision so it is sort of an open question on my PR

parfeniukink: Could you please send a link to that PR with the code? Can't find it

DaltheCow: Sure!
https://github.com/neuralmagic/guidellm/pull/2/files

parfeniukink:
Depricated onboarding PR

PR #9, opened 2024-07-11
Add serialization and deserialization classes and methods with pydantic

markurtz:


setup.py
```diff
@@ -23,6 +23,7 @@ def _setup_long_description() -> Tuple[str, str]:
         'loguru',
         'numpy',
         'openai',
+        'pydantic>=2.0',
```
parfeniukink: What do you think if we will route the correct call in the `Serializable` class depending on the pydantic version and remove the `>=2.0` from the setup.py file?

src/guidellm/core/distribution.py
```diff
@@ -1,202 +1,159 @@
...
-        distribution.
-    :type data: List[Union[int, float]], optional
+    A class to represent a statistical distribution and perform various
+    statistical analyses.
     """
 
-    def __init__(self, data: List[Union[int, float]] = None):
-        """
-        Initialize the Distribution with optional data.
+    data: Optional[List[Union[int, float]]] = []
```
parfeniukink: that one is similar to dataclasses in Python
```suggestion
    from pydantic import Field
    data: List[Union[int, float]] = Field(default_factory=list)
```

src/guidellm/core/serializable.py
```diff
@@ -0,0 +1,72 @@
...
+    deserialization.
+    """
+
+    def __init__(self, /, **data: Any) -> None:
+        super().__init__(**data)
+        logger.debug(
+            "Initialized new instance of {} with data: {}",
+            self.__class__.__name__,
+            data,
+        )
```
parfeniukink: What about adding the configuration object to the shared entity class?
```
from pydantic import AnyHttpUrl, BaseModel, ConfigDict
class Serializable(BaseModel):
    model_config = ConfigDict(
        extra="forbid",
        use_enum_values=True,
        validate_assignment=True,
        arbitrary_types_allowed=True,
        from_attributes=True,
    )
```

src/guidellm/core/serializable.py
```diff
@@ -0,0 +1,72 @@
...
+        Deserialize a JSON string to a model instance.
+
+        :param data: JSON string to deserialize.
+        :return: An instance of the model.
+        """
+        logger.debug("Deserializing from JSON... {}", data)
+        obj = cls.model_validate_json(data)
+        logger.debug("Deserialized from JSON: {}", obj)
+
+        return obj
```
parfeniukink: Don't you think that we are going to have logs pollution? Is it a requirement having such many loggers for each transformation step?

parfeniukink (commented):
Nice job!

setup.py
markurtz: I'd prefer not to unless we have a specific use case/reason or active conflict we know we'll hit to support Pydantic < 2.0. This is to keep our test matrix smaller

src/guidellm/core/distribution.py
markurtz: yep, thanks Dmytro, forgot about the issue with mutable defaults. Will update across with the Field for factory creation

src/guidellm/core/serializable.py
markurtz: Removing the final state logs to cut the number in half. This is at debug level and isn't frequently called, so I'm less worried about log pollution there

src/guidellm/core/result.py
```diff
@@ -22,19 +23,43 @@ class TextGenerationResult(Serializable):
...
+    )
+    end_time: float = Field(
+        default=None, description="The end time of the text generation."
+    )
+    first_token_time: float = Field(
+        default=None, description="The time taken to decode the first token."
+    )
+    decode_times: Distribution = Field(
+        default_factory=Distribution, description="The distribution of decode times."
+    )
```
parfeniukink, line 62:
```py
    request: TextGenerationRequest = Field(
        description="The text generation request used to generate the result."
    )
    prompt: str = Field(
        default="", description="The input prompt for the text generation."
    )
    prompt_word_count: int = Field(
        default=0, description="The number of words in the input prompt."
    )
    prompt_token_count: int = Field(
        default=0, description="The number of tokens in the input prompt."
    )
    output: str = Field(
        default="", description="The generated output for the text generation."
    )
    output_word_count: int = Field(
        default=0, description="The number of words in the output."
    )
    output_token_count: int = Field(
        default=0, description="The number of tokens in the output."
    )
    last_time: Optional[float] = Field(default=None, description="The last time recorded.")
    first_token_set: bool = Field(
        default=False, description="Whether the first token time is set."
    )
    start_time: Optional[float] = Field(
        default=None, description="The start time of the text generation."
    )
    end_time: Optional[float] = Field(
        default=None, description="The end time of the text generation."
    )
    # ...
```
 
Since you didn't specify many `Optional` fields I have to ask you whether you refer to the `pydantic`'s ability to set the default = `None` but restrict the initializer argument with `float` type. If yes, what do you think about using pydantic validators instead?

src/guidellm/core/result.py
```diff
@@ -150,19 +184,28 @@ class TextGenerationBenchmark(Serializable):
...
+    results: List[TextGenerationResult] = Field(
+        default_factory=list, description="The results of the text generation requests."
+    )
+    errors: List[TextGenerationError] = Field(
+        default_factory=list, description="The errors of the text generation requests."
+    )
+    concurrencies: List[RequestConcurrencyMeasurement] = Field(
+        default_factory=list,
+        description="The concurrency measurements of the requests.",
+    )
```
parfeniukink, line 200: Should we make `mode` an `Enum`?

src/guidellm/core/result.py
```diff
@@ -185,20 +228,45 @@ def error_count(self) -> int:
...
+            return True
+
+        # rate comparisons did not give a clear signal,
+        # let's double check that we aren't overloaded by comparing the
+        # compute throughput for the benchmark with the latency for the requests.
+        # overall this means that a relatively flat or decreasing throughput curve
+        # over time in addition to a growing processing queue is a sign of overload
+
+        # TODO
+        return False
```
parfeniukink, line 268: oh, this is what I was looking for 😄

parfeniukink (approved):
Nice

src/guidellm/core/result.py
markurtz: Yep, so the issue I was hitting with this was mypy raising typing errors when setting none for non optional fields.

src/guidellm/core/result.py
markurtz: Yes, I don't see any issue with that. Let's move that to another enhancement PR, though

PR #11, opened 2024-07-15
✨ Executor & Scheduler packages preparation

parfeniukink:
# TLDR

The `Executor` and `Scheduler` are prepared to work with.
Sweep profile generation mode and Poisson load generation modes are not tested ( _out of scope from the feature's perspective_ )

# Detailed

* ⚡️The overall stability of the application is improved.
* 🔥 Some deprecations are removed
* 🔥 The `task` module is completely replaced with the basics of the Python `asyncio` module
* ✅ `executor` package integration and unit tests are added
* ✅ `scheduler` package integration and unit tests are added
* ✅ Benchmarks report generations are tested in all combinations of modes
* ♻️ `scheduler.Scheduler` now works with `asyncio` instead of the custom `Task` representation.
* 🚚 `LoadGenerationModes` -> `LoadGenerationMode`


parfeniukink:
@markurtz Please check whether the Execution+Scheduler works as you expect. Once I have that intermediate approval I will add some more integration tests _( they are stashed locally but are ready to be pushed. Just wanted to make the first review cleaner because of some other minor changes )_

src/guidellm/core/result.py
```diff
@@ -148,6 +148,7 @@ def end_time(self) -> float:
         :rtype: float
         """
 
+        self._recording_started()
```
markurtz: Can we rename this function to check_recording_started or something along those lines? I was initially very confused why we were recording started during the end_time call

src/guidellm/core/result.py
```diff
@@ -148,6 +148,7 @@ def end_time(self) -> float:
...
         assert self._end_time
```
markurtz: I'm not a fan of burying asserts in the code as they don't give context and are not catchable. Can we update this to log at a message at a min with the assert, but ideally raise a targetable exception that is appropriate?

markurtz: Ditto towards the starat_time property and the assert contained in there as well

src/guidellm/core/result.py
```diff
@@ -270,7 +271,11 @@ class TextGenerationError:
     :type error: Exception
     """
 
-    def __init__(self, request: TextGenerationRequest, error: Exception):
+    def __init__(
+        self,
+        request: TextGenerationRequest,
+        error_class: BaseException,
```
markurtz: Good catch here for the generator exit exception

markurtz (changes requested):
## Scheduler Package
### Scheduler Module
**Constructor Arguments:**
- max_requests and max_duration: Ensure these arguments are not only non-None but also positive values during initialization.
**Instance Properties:**
- self._tasks: Consider removing this property. Tasks are local to the run functions, and they are not being removed upon completion. Managing this list would require adding thread safety.
- self.event_loop: Why is this a public property? Would a utility function be more appropriate?
**Methods:**
- self.run: Since we are populating the async down to the backend, we should make this method async. This will simplify the calling function, eliminating the need to manage an event loop before calling into this function.
**Load Generator:**
- Propose moving this to an instance property and set it to None if in sync mode. These values will remain unchanged after construction, and the invocation times on the generator should reflect this.
**Future Considerations:**
- Develop a mechanism to pipe progress from the scheduler, reporting the number of tasks completed so far. This should propagate to the executor and further up to the caller, ensuring clean progress updates for the CLI and future UI.
### Task Module
**Module Removal:**
- This module may be obsolete since we have async running end to end. Previously, it was needed to wrap a sync method into an async one and add cancellation capability. This functionality is now fully encapsulated by asyncio tasks.
### Load Generator
**Review:**
- The current implementation looks good.
## Executor Package
**Instance Properties:**
- self.scheduler: Similar to the tasks mentioned above, is there a reason to make this a class property? It’s scoped within the local run function, and making it thread-safe for external access during the run would require additional work.
- Public Setter for Scheduler: There seems to be no need for a public setter. The run function should be the sole user and setter of the scheduler.
**Methods:**
- self.run: As discussed above, we should transition this to an async function. This change aligns with the async flow throughout the library and simplifies the caller’s process.
**Future Considerations:**
- Determine a method to pipe progress from the scheduler, covering the number of profiles needed to run, and up to the caller. This should facilitate reporting through the CLI and UI.

src/guidellm/core/result.py
parfeniukink: Yes, you're right. I was more about `recording_is_started` but probably this is because of my language barrier 😄

src/guidellm/core/result.py
parfeniukink: So yeah. This assert statement was added to catch your attention 😅

From the philosophical perspective, I remember that we agreed that we were going to use a `pydantic` core for our Entities. So the assertion is just a quick hint for the mypy ( since the GitHub Actions is already there ) and they will probably be removed after you merge your *Serializers* Pull Request.

parfeniukink: What do you think?
It will be easier for me to resolve conflicts after rebasing

markurtz: Ah gotcha, makes sense

markurtz: Yeah, let me finalize the pydantic diff today and then can rebase on that

parfeniukink:
> ## Scheduler Package
> ### Scheduler Module
> **Constructor Arguments:**
> 
> * max_requests and max_duration: Ensure these arguments are not only non-None but also positive values during initialization.
> 
> ...

_NOTE: the original message is cut._

@markurtz  here are some comments for 🚧 points. ✅ - means I agree and this is already done in that PR
Scheduler Module
Constructor Arguments:
* max_requests and max_duration positive initialization step ✅
Instance Properties:
* self._tasks: Consider removing this property. ✅
* self.event_loop: ✅ 
Methods:
* self.run: ... we should make this method async... I don't know if I get the idea here. The Scheduler supports both: sync and async executions. This method just dispatches the execution code. The `_async_run` is a coroutine like you are saying. Or do you talk about having 2 types of Scheduler for sync and async jobs? 🚧
Load Generator:
* Propose moving this to an instance property and set it to None if in sync mode. I believe that we can even raise the error in that case to avoid union of types. ✅

Future Considerations:
* Develop a mechanism to pipe progress from the scheduler, reporting the number of tasks completed so far. This should propagate to the executor and further up to the caller, ensuring clean progress updates for the CLI and future UI. 🚧 Postponed to another PR

Task Module
Module Removal: ✅ 100% agree with you

Executor Package
Instance Properties:
* self.scheduler: Agree ✅ TLDR: overengineered for better testing experience. found another way to check the executor...
Methods:
* self.run: ✅ Alright. Are we going to move to the async engine on the OpenAI's side as well? I mean we still use the `Stream[Completion]` instead of a `AsyncStream`, right? 🚧
Future Considerations:
* Determine a method to pipe progress from the scheduler, covering the number of profiles needed to run, and up to the caller. This should facilitate reporting through the CLI and UI.  🚧 Postponed to another PR

parfeniukink:
@markurtz The failed CI is because of the changes in the `core` module. What about postponing small fixes ( _95% are resolved_ ) for later?

markurtz:
> @markurtz The failed CI is because of the changes in the `core` module. What about postponing small fixes ( _95% are resolved_ ) for later?

Since we don't have active CI on this for tests and haven't finalized a working version yet, I think it's fine to get this landed provided we quickly follow up with a diff the outstanding issues and ensure prechecks across all files, make quality, and tests are all passing

parfeniukink:
@markurtz resolved all the serialization issues and tests. Could you approve that one? It does not allow me to merge

markurtz (approved):

PR #16, opened 2024-07-23
Add CI/CD workflows with tox and GitHub Actions for comprehensive testing, styling, and quality checks

markurtz:
## Summary
Implements a series of GitHub Actions workflows and tox environments to streamline the CI/CD processes for the project. It replaces the previous Makefile-based pathways and focuses on modernizing the development, testing, and deployment practices.

## Details
- CI/CD Workflows: Introduced multiple GitHub Actions workflows for development, nightly builds, quality checks, and releases.
  - Development Workflow: Runs on pull requests to any branch. Includes unit tests and integration tests for Python versions 3.8 to 3.12.
  - Nightly Workflow: Scheduled to run nightly, ensuring continuous integration health with the unit, integration, and end-to-end tests.
  - Quality Workflow: This runs on pushes to the main branch and pull requests. It ensures code quality with linting, formatting, and type checking.
  - Release Workflow: Triggers on push to release branches, running comprehensive tests to prepare for new releases.
- tox Configuration: Updated tox.ini to define environments for unit tests, integration tests, end-to-end tests, quality checks, style checks, type checks, build, and clean operations.
- Developer Documentation: Expanded DEVELOPING.md to guide contributors in setting up the development environment, running tests, and maintaining code quality.
- README Updates: Updated project documentation to reflect the new CI/CD practices and removed outdated Makefile instructions.

## Test Plan
Pending running the new GitHub Actions workflows. If any issues pop up, those will be fixed either in this diff or in subsequent ones.


.github/workflows/development.yml
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python }}
+      - name: Install dependencies
+        run: pip install tox
+      - name: Run integration tests
+        run: tox -e test-integration -- -m smoke
```
parfeniukink, line 43: What do you think about leaving the `tox` only for local runnings and **_"python versions matrix"_** for GitHub Actions?

parfeniukink: Like how it is done in main right now

Makefile
parfeniukink, line 1: Since you decided to remove the Makefile from the project (probably because of the `tox`), What do you think about adding it to the `.gitignore` so we are still available to use it locally

pyproject.toml
```diff
@@ -44,6 +44,7 @@ dev = [
 
     # code quality
     "black~=24.4.2",
+    "flake8~=7.1.0",
```
parfeniukink: why do we still use it along with `ruff`?

parfeniukink (approved):

.github/workflows/development.yml
markurtz: I'm not sure I'm following. tox is there for local runs and GitHub actions to solidify all pieces behind the same tool rather than having separations for Makefile and tox. The tox config itself has the python environments still within it, GitHub actions recommends the version matrix, though, to run on each python version and ensure the setup action installs everything correctly. The other benefit of this pathway that GitHub Actions recommends is that we have separate job reports coming in for each Python version that is being tested and, therefore, can quickly figure out what is broken and in what environment.

For more info, see GitHub's docs here: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python#running-tests-with-tox

Makefile
markurtz: What's the benefit of utilizing the Makefile locally? I migrated everything over other than the install commands which are now recorded in the developer docs. Mainly calling out that I'd like to ensure that everyone is going through the same tox commands to handle the development and testing pathways, as this is one of the core benefits tox advertises.

pyproject.toml
markurtz: Good point, removing flake8. This was leftover from earlier where we still had flake8 running in the commands, but had dropped the requirement

.github/workflows/development.yml
parfeniukink: Oh, thanks. Didn't know that Github works concurrently with tox. This is why I wonder :)

Makefile
parfeniukink: So yeah, I am pretty agree with you and `tox` makes everything for us. But locally it is phenomenally easier to run everything with the `make` command instead of `tox -e ...` for all the cases.

Usually, I just run locally `make test.unit` or `make fix` to get things done, since you do that pretty often.
From my understanding, since we are going to develop the public version of the guidellm it has to be supported by Python 3.8, and `tox` just automates all the checks if you need them.

Probably `tox`'s commands set includes the set of `Makefile`'s commands. Anyway, the `make` command is only for developers who run the project. Commands inside could be dynamic, depending on those developers' needs IMHO.

But still, I can bypass that `make` with a ton of things, so no worries about it. Just out of curiosity.

tox.ini
```diff
@@ -2,20 +2,82 @@
...
+    mypy --check-untyped-defs
+
+
+[testenv:build]
+description = Build the project
+deps =
+    setuptools
+    wheel
+commands =
+    python setup.py sdist bdist_wheel
```
parfeniukink: python also supports `python -m build`. Does it make the same?

markurtz: Good question, I believe the only difference is the setup.py pathway doesn't isolate the build environment. python -m is the newer recommended way, so I'll update to that

PR #17, opened 2024-07-24
👔 Project settings is now encapsulated into `guidellm.config` module

parfeniukink:
* The module is currently has only `__init__` file
* All `os.getenv` replaced with `pydantic_settings` facade
* `GUIDELLM__` env prefix is set

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -10,6 +10,7 @@ This document aims to provide developers with all the necessary information to c
 ### Prerequisites
 
 Before you begin, ensure you have the following installed:
+
 - Python 3.8 or higher
 - `pip` (Python package installer)
 - `git` (version control system)
@@ -17,19 +18,21 @@ Before you begin, ensure you have the following installed:
 ### Installation
 
 1. Clone the repository:
-    ```bash
-    git clone https://github.com/neuralmagic/guidellm.git
-    cd guidellm
-    ```
+
+   ```bash
+   git clone https://github.com/neuralmagic/guidellm.git
+   cd guidellm
+   ```
 
 2. Install the required dependencies:
-    ```bash
-    pip install -e .[dev]
-    ```
+   ```bash
+   pip install -e .[dev]
+   ```
 
 ## Project Structure
 
 The project follows a standard Python project structure:
+
 ```plaintext
 guidellm/
 ├── src/
@@ -51,14 +54,15 @@ guidellm/
 To set up your development environment, follow these steps:
 
 1. Install pre-commit hooks:
-    ```bash
-    pre-commit install
-    ```
+
+   ```bash
+   pre-commit install
+   ```
 
 2. Ensure all dependencies are installed:
-    ```bash
-    pip install -e .[dev]
-    ```
+   ```bash
+   pip install -e .[dev]
+   ```
 
 ## Code Quality and Style Guidelines
 
@@ -70,9 +74,9 @@ Ruff is used for linting and formatting checks.
 
 - Configuration is in `pyproject.toml`.
 - To run Ruff:
-    ```bash
-    ruff check src tests
-    ```
+  ```bash
+  ruff check src tests
+  ```
 
 ### Isort
 
@@ -80,9 +84,9 @@ Isort is used for sorting imports.
 
 - Configuration is in `pyproject.toml`.
 - To sort imports:
-    ```bash
-    isort src tests
-    ```
+  ```bash
+  isort src tests
+  ```
 
 ### Flake8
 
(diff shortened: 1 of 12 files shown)


DEVELOPING.md
```diff
@@ -207,3 +213,35 @@ Please refer to the MAINTAINERS file for maintenance guidelines and contact info
 ## Contact and Support
 
 If you need help or have any questions, please open an issue on GitHub or contact us at support@neuralmagic.com.
+
+This will run automatic code styling using `black` and `isort` and test that the
```
markurtz: This text is out of place here, was this a miscopy?

DEVELOPING.md
```diff
@@ -207,3 +213,35 @@ Please refer to the MAINTAINERS file for maintenance guidelines and contact info
...
+repository's code matches its standards.
+
+## Project configuration
```
markurtz, line 213: Can we move this section above contact and support and just after the formatting, linting, and type checking section?

DEVELOPING.md
```diff
@@ -207,3 +213,35 @@ Please refer to the MAINTAINERS file for maintenance guidelines and contact info
...
+
+The project configuartion is powered by _[`🔗 pydantic-settings`](https://docs.pydantic.dev/latest/concepts/pydantic_settings/)_
+
+The project configuration entrypoint is represented by lazy-loaded `settigns` singleton object ( `src/config/__init__` )
+
+The project is fully configurable with environment variables. All the default values and
+
+```py
```
markurtz: I'm not sure if I'm following the purpose of having the sample code here. The code within the repo invokes immediately, right? So, we should likely have a full list of the env variables here that people would want to change alongside a link to the source like above in place of code showing them how it works

DEVELOPING.md
```diff
@@ -207,3 +213,35 @@ Please refer to the MAINTAINERS file for maintenance guidelines and contact info
...
 If you need help or have any questions, please open an issue on GitHub or contact us at support@neuralmagic.com.
+
+This will run automatic code styling using `black` and `isort` and test that the
+repository's code matches its standards.
+
+## Project configuration
+
+The project configuartion is powered by _[`🔗 pydantic-settings`](https://docs.pydantic.dev/latest/concepts/pydantic_settings/)_
+
+The project configuration entrypoint is represented by lazy-loaded `settigns` singleton object ( `src/config/__init__` )
```
markurtz:
```suggestion
The project configuration entry point is represented by lazy-loaded `settings` singleton object ( `src/config/__init__` )
```

src/guidellm/logger.py
```diff
@@ -64,56 +55,25 @@ def configure_logger(config: Optional[LoggerConfig] = None):
...
+    if config.log_file or config.log_file_level:
+        log_file = config.log_file or "guidellm.log"
+        log_file_level = config.log_file_level or "INFO"
         # log as json to the file for easier parsing
         logger.add(log_file, level=log_file_level.upper(), serialize=True)
-
-
-# invoke logger setup on import with default values enabling console logging with INFO
-# and disabling file logging
-configure_logger(
```
markurtz, line 111: We should keep the root configuration here in the event someone uses the repo as a library rather than going through the main CLI so we always have a base configuration setup and will log out properly. In this way, both CLIs and the library will log the same while enabling users to either override through code or through env variables

markurtz (changes requested):
Looks good and I like how simple the env loading is with pydantic. A few small changes requested

DEVELOPING.md
parfeniukink: Ohh. Yes probably

DEVELOPING.md
parfeniukink: It just reflects the idea of how this works. It might be helpful for developers who gonna work on the project. I believe that the sample of a code represents the idea better than words in most cases but I can remove it.

parfeniukink: This is kinda you have a code and a small description below that explains the idea on that concrete code 😄
this the _"the purpose of having the sample code here"_

src/guidellm/logger.py
parfeniukink: I don't understand what you mean by the **_root configuration_**.
Do you mean moving the `LoggingSettings` class there?

markurtz: Not quite, I mean that we should invoke configure_logger() in this base logger file like it had before so that on import into the guidellm's __init__.py (and therefore any imports from guidellm) it will trigger the setup of the logger

parfeniukink: ahh, now I got the idea

markurtz (approved):

PR #18, opened 2024-07-24
💄CLI interface improvements

parfeniukink:
# Summary

- The `click` CLI interface is tested with a bunch of unit tests
- The `main` function validation is added.
- `OpenAIBackend` initializer parameters are optimized
    - target, host, and port parameters usage is simplified
- `openai.NotFound` **_available models_** error is handled
- `SerializableFileType` renamed to `SerializableFileExtension`
- `SerializableFileExtension` now inherits `str` to simplify usage, since this Enum class is mostly used to work with strings.
- `rate_type_to_load_gen_mode` renamed to `RATE_TYPE_TO_LOAD_GEN_MODE_MAPPER`
- `rate_type_to_profile_mode` renamed to `RATE_TYPE_TO_PROFILE_MODE_MAPPER`
- CLI parameters are renamed:
    - `--num-seconds` -> `--max-seconds`
    - `--num-requests` -> `--max-requests`
- `path` removed from CLI arguments since it is not used
- .env `GUIDELLM` prefix is fixed
- Unused comments, settings, and code are removed
- Logger default unit test uses the injected logging settings object
- Module `backend.openai` has `_base_url` renamed to the `base_url`
- In `OpenAIBackend.make_request`, the `GenerativeResponse` always counts `output_tokens` with `self._token_count`
- `SerializableFileExtensions` is replaced with pure Python strings

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -231,7 +231,7 @@ The project is fully configurable with environment variables. With that configur
 ## Contact and Support
 
 If you need help or have any questions, please open an issue on GitHub or contact us at support@neuralmagic.com.
-=======
+
 This will run automatic code styling using `black` and `isort` and test that the
 repository's code matches its standards.
 
@@ -263,4 +263,3 @@ settings = Settings()
 ```
 
 With that configuration set you can load parameters to `LoggingSettings()` by using environment variables. Just run `export GUIDELLM__LOGGING__DISABLED=true` or `export GUIDELLM__LOGGING__NESTED=another_value` respectfully. The nesting delimiter is `__`
->>>>>>> d8244f6 (📝 Project `configuration management` is documented)
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -231,3 +231,36 @@ The project is fully configurable with environment variables. With that configur
 ## Contact and Support
 
 If you need help or have any questions, please open an issue on GitHub or contact us at support@neuralmagic.com.
+=======
+This will run automatic code styling using `black` and `isort` and test that the
+repository's code matches its standards.
+
+## Project configuration
+
+The project configuartion is powered by _[`🔗 pydantic-settings`](https://docs.pydantic.dev/latest/concepts/pydantic_settings/)_
+
+The project configuration entrypoint is represented by lazy-loaded `settigns` singleton object ( `src/config/__init__` )
+
+The project is fully configurable with environment variables. All the default values and
+
+```py
+class NestedIntoLogging(BaseModel):
+    nested: str = "default value"
+
+class LoggingSettings(BaseModel):
+    # ...
+    disabled: bool = False
+
+
+class Settings(BaseSettings):
+    """The entrypoint to settings."""
+
+    # ...
+    logging: LoggingSettings = LoggingSettings()
+
+
+settings = Settings()
+```
+
+With that configuration set you can load parameters to `LoggingSettings()` by using environment variables. Just run `export GUIDELLM__LOGGING__DISABLED=true` or `export GUIDELLM__LOGGING__NESTED=another_value` respectfully. The nesting delimiter is `__`
+>>>>>>> d8244f6 (📝 Project `configuration management` is documented)
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -228,12 +228,7 @@ The project is fully configurable with environment variables. With that configur
 | `GUIDELLM__OPENAI__BASE_URL`   | `http://localhost:8080` | The address to the **OpenAI-compatible** server.<br><br>OpenAI live base url is `https://api.openai.com/v1`                                             |
 | `GUIDELLM__OPENAI__API_KEY`    | `invalid`               | Corresponds to the **OpenAI-compatible** server API key.<br><br>If you look for the live key - check [this link](https://platform.openai.com/api-keys). |
 
-## Contact and Support
-
-If you need help or have any questions, please open an issue on GitHub or contact us at support@neuralmagic.com.
-
-This will run automatic code styling using `black` and `isort` and test that the
-repository's code matches its standards.
+<br>
 
 ## Project configuration
 
@@ -263,3 +258,7 @@ settings = Settings()
 ```
 
 With that configuration set you can load parameters to `LoggingSettings()` by using environment variables. Just run `export GUIDELLM__LOGGING__DISABLED=true` or `export GUIDELLM__LOGGING__NESTED=another_value` respectfully. The nesting delimiter is `__`
+
+## Contact and Support
+
+If you need help or have any questions, please open an issue on GitHub or contact us at support@neuralmagic.com.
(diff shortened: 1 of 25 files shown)


DEVELOPING.md
```diff
@@ -231,3 +231,35 @@ The project is fully configurable with environment variables. With that configur
 ## Contact and Support
 
 If you need help or have any questions, please open an issue on GitHub or contact us at support@neuralmagic.com.
+
+This will run automatic code styling using `black` and `isort` and test that the
```
markurtz: Need to remove this section since it should have landed with the config diff

examplemodel.yaml
```diff
@@ -0,0 +1,2 @@
+age: 30
```
markurtz: Why is this at the root?

src/config/__init__.py
```diff
@@ -45,8 +46,18 @@ class Settings(BaseSettings):
         extra="ignore",
     )
 
+    root_dir: Path
+
+    # TODO: add to the DEVELOPING.md after
+    # https://github.com/neuralmagic/guidellm/pull/17
+    # is merged
+    debug: bool = False
```
markurtz: I'm not following the need for a debug statement that only controls the model returned from OpenAI server.

src/config/__init__.py
```diff
@@ -45,8 +46,18 @@ class Settings(BaseSettings):
...
+
     logging: LoggingSettings = LoggingSettings()
     openai: OpenAISettings = OpenAISettings()
 
 
-settings = Settings()
+settings = Settings(
+    # NOTE: hardcoded since should not be changed in a runtime
+    root_dir=Path(__file__).parent.parent.parent,
```
markurtz: This is very fragile especially as we look at the install pathways for this project. Let's come up with a better way of setting the root_dir path. Specifically, the main.py should control that since it has access to the current working directory as well as contextual knowledge that this was run in the CLI.

Along these lines, I'm not following setting a root directory here for something that should run through the CLI file ideally

src/guidellm/__init__.py
```diff
@@ -2,7 +2,3 @@
 Guidellm is a package that provides an easy and intuitive interface for
 evaluating and benchmarking large language models (LLMs).
 """
-
-from .logger import configure_logger, logger
```
markurtz, line 6: We need to keep these here so that on import from any place within guidellm, it will properly configure the logger

src/guidellm/backend/openai.py
```diff
@@ -54,19 +58,22 @@ def __init__(
                 "must be specify for the OpenAI backend"
             )
 
-        if not (_base_url := (internal_callback_url or settings.openai.base_url)):
+        if target is not None:
+            _base_url = target
```
markurtz: can we rename this without the underscore prefix, this will keep it consistent with the rest of the repo for variable names within a function under local scope not having a prefix

src/guidellm/backend/openai.py
```diff
@@ -86,7 +93,12 @@ def make_request(
         request_args: Dict = {"n": 1}
 
         if (num_gen_tokens := request.params.get("generated_tokens", None)) is not None:
```
markurtz: We should still set a safety here of max tokens that can be generated if generated_tokens is not set, 4096 is a good starting value

src/guidellm/backend/openai.py
```diff
@@ -128,12 +140,19 @@ def available_models(self) -> List[str]:
...
-
-        return models
+        try:
+            models: List[str] = [
+                model.id for model in self.openai_client.models.list().data
+            ]
+            logger.info(f"Available models: {models}")
+        except openai.NotFoundError as error:
+            logger.error(error)
+            if settings.debug is True:
```
markurtz: I'm not following the rationale for this, especially since OpenAI and GPT-4o are not our main targets for benchmarking

src/guidellm/core/serializable.py
```diff
@@ -42,6 +43,16 @@ def __init__(self, /, **data: Any) -> None:
             data,
         )
 
+    @staticmethod
+    @functools.lru_cache(maxsize=1)
+    def available_file_extensions() -> List[str]:
```
markurtz: this is going to inject memory leak issues since it will hold onto the self reference. Additionally, this is very lightweight code and I don't see what we'll gain by adding the added complexity of caching this

src/guidellm/core/serializable.py
```diff
@@ -103,44 +120,34 @@ def save_file(self, path: str, type_: Optional[SerializableFileType] = None) ->
             it will save in YAML format.
         :return: The path to the saved file.
         """
-        logger.debug("Saving to file... {} with format: {}", path, type_)
```
markurtz, line 106: By removing the starting line, we have logic that can raise an error after and no way for the user to see what the inputs were that led to that error

markurtz (changes requested):
Left a few comments. Additionally, we need to add in integration and e2e tests for the CLI as discussed yesterday

src/guidellm/core/serializable.py
parfeniukink: Alright. I will remove this, no problem. I remember you told me about the performance :)
By the way, I thought that the `staticmethod` is going to remove the `self` object, isn't it?
https://docs.python.org/3.11/library/functions.html#staticmethod

src/guidellm/core/serializable.py
parfeniukink: Oops. My bad

src/guidellm/backend/openai.py
parfeniukink: Great! Setting it up.

src/guidellm/backend/openai.py
parfeniukink: sure

src/guidellm/__init__.py
parfeniukink: sorry. not sure why this is removed. my bad

src/config/__init__.py
parfeniukink: yeah. you are right. I will fix it

src/config/__init__.py
parfeniukink: Well. Yes, you are right. Probably this is my past following me

examplemodel.yaml
parfeniukink: wow, didn't see that tests created this

src/guidellm/backend/openai.py
parfeniukink: So since we choose the llama.cpp as our primary openai serving engine it doesn't have anything for models.list().

Also if you serve the model with your llama.cpp it does not matter which model you request. And this is the actual reason why I added the settings.debug parameter. So we at least can run it locally.

If the concern is only about the gpt-4o we can rename it.

examplemodel.yaml
parfeniukink: btw this is the one from your `ExampleModel` class

markurtz (approved):

markurtz (approved):

PR #19, opened 2024-07-24
Adapt build system for dynamic versioning and package naming based on build settings

markurtz:
## Summary
Enhances the build system by introducing dynamic versioning and package naming based on the build type and build number, which are loaded from env variables.

## Details
- Dynamic Versioning: The build version is now determined dynamically based on the environment variables GUIDELLM_BUILD_TYPE and GUIDELLM_BUILD_NUMBER. This allows for flexible versioning strategies such as dev, nightly, and release builds.
- Package Naming: The package name is adapted according to the build type, supporting different names for dev, nightly, and release versions.
- Script Addition: Added utils/inject_build_props.py to automate the injection of build properties into the pyproject.toml file.
- Dependencies: Updated dependencies in pyproject.toml to enable pre script builds.
- CI/CD Integration: Updated tox.ini to integrate the new build script and ensure all checks are aligned with the updated build process.
- Code Cleanup: Removed unused dependencies from .pre-commit-config.yaml..

## Test Plan
- Manually tested the build process for dev, nightly, and release builds to ensure proper versioning and naming.


parfeniukink (approved):
Why is the `black` removed from the `pre-commit` configuration?

parfeniukink:
> > Why is the `black` removed from the `pre-commit` configuration?
> 
> I had missed the cleanup on the pre-commit previously with the tox.ini rework. Black is stripped out completely across the system now along with isort and flake8 since ruff has full parity with those and runs much faster

alrigh. but in that case `ruff` can replace `isort` (_probably ruff not that good, but anyway_) and `flake8` as well.

PR #23, opened 2024-07-29
Expand ruff configurations to enforce comprehensive code quality standards

markurtz:
## Summary

Enhances our `ruff` configurations to enforce comprehensive standards across various aspects of code quality, including style, formatting, structure, complexity, security, bug prevention, and documentation.

### Changes Implemented:

1. **Code Style and Formatting:**
    - Enforced PEP 8 conventions, including spacing, indentation, and line length.
    - Included `isort` for consistent import sorting.
    - Added rules for trailing commas, commented-out code, import conventions, naming conventions, and quote usage.
2. **Code Structure and Complexity:**
    - Simplified complex expressions and improved readability.
    - Applied best practices for list, set, and dict comprehensions.
    - Checked for overly complex code using cyclomatic complexity.
3. **Code Security and Bug Prevention:**
    - Detected unused function arguments, boolean traps, and implicit string concatenations.
    - Prevented common programming mistakes and potential bugs.
    - Enforced correct exception handling and detected security issues.
4. **Code Documentation:**
    - Ensured adherence to documentation standards.
    - Detected unresolved FIXMEs and other temporary comments.

### Files Modified:

- `.pre-commit-config.yaml`: Updated configurations to include new ruff rules and standards.
- `pyproject.toml`: Extended ruff linting rules to cover additional checks for various code quality aspects.
- Source and test files: Applied necessary changes to comply with the new ruff configurations, including refactoring code, updating documentation, and adding comments.

### Test Plan:

- **Automated Testing:**
    - Automated tests updated to run the latest changes
- **Manual Testing:**
    - Manually reviewed changes to ensure they align with the updated linting standards.
    - Manually ran tests and checks


src/guidellm/backend/base.py
```diff
@@ -157,5 +155,14 @@ def model_tokenizer(self, model: str) -> Optional[str]:
...
+
+@functools.lru_cache(maxsize=1)
+def _cachable_default_model(backend: Backend) -> str:
+    if models := backend.available_models():
+        logger.debug(f"Default model: {models[0]}")
+        return models[0]
 
-        pass
+    logger.error("No models available.")
+    raise ValueError("No models available.")
```
parfeniukink, line 168: Does the `lru_cache` work fine with `type: Backend`? I am not sure bit it caches only hashable types since it is just a dictionary under the hood.

src/guidellm/backend/openai.py
```diff
@@ -1,15 +1,16 @@
...
 from openai import OpenAI, Stream
-from openai.types import Completion
 from transformers import AutoTokenizer
 
-from config import settings
 from guidellm.backend import Backend, BackendEngine, GenerativeResponse
 from guidellm.core import TextGenerationRequest
 
+if TYPE_CHECKING:
+    from openai.types import Completion
```
parfeniukink: Do we really need it?
I would suggest importing it as usual, so we can detect circular imports faster. And they point to architecture problems in the code structure. might be helpful i believe.

parfeniukink (approved):

src/guidellm/backend/openai.py
markurtz: Since these pathways are for third-party libraries, we shouldn't be able to hit any circular imports there. It was recommended to add a bit more performance for the pyc files, and based on the spec, seemed like a reasonable thing to do. I'm indifferent on fully enforcing it, but see the spec here for more background: https://docs.astral.sh/ruff/rules/typing-only-third-party-import/

parfeniukink: Well.

> Unused imports add a performance overhead at runtime

I believe that the impact of such operations is tiny. We don't use any 3-rd party imports that have any sort of computation. I mean calling it overhead is quite ridiculous 😄. Like we use Python, common.

> and risk creating import cycles

So this is the topic that I am talking about. If you have the cyclic import in your code - this is literally the architecture problem.  Like don't you think that the regular import is going to be way easier because of the caching than having such an `if` statement in most of the modules on a distance?

_P.S. I was using the `typing.TYPE_CHECKING` before but seems it gives nothing if you don't **_check types_** and import modules that execute the code on import which is usually wild for most of the packages._

src/guidellm/backend/base.py
markurtz: Double checked, yes, this will work with type: Backend. The default implementation for hashable on python classes is to return a unique identifier for the class

parfeniukink (approved):

parfeniukink (approved):

PR #24, opened 2024-08-01
Add Report Generation and Configuration Improvements for GuideLLM

markurtz:
## Summary

Introduces a report file generation feature for GuideLLM and fixes the configuration functionality. Once the frontend is live, we will add the report generation functionality to the CLI and add integration and end to end tests 

## Details

- **Report Generation Feature**:
    - Added logic to generate and inject reports from Pydantic models.
    - Created utility functions for loading HTML templates, injecting data, and saving reports.
    - Implemented support for both local file paths and URLs as sources for HTML templates.
    - Included handling for different environments (local, dev, staging, prod) with environment-specific report URLs.
- **Configuration Improvements**:
    - Refactored configuration structure into a more modular and maintainable format.
    - Updated configuration to use Pydantic's `BaseSettings` and `model_validator` for environment-based settings.
    - Improved logging configuration with customizable options for console and file logging levels.
    - Enhanced OpenAI settings for more robust API connectivity.
- **Testing**:
    - Added comprehensive unit and integration tests for the new report generation feature.
    - Updated existing tests to align with the new configuration structure.
    - Ensured all tests pass successfully and that code coverage is maintained.

## Test Plan

- **Automation Testing**:
    - Added and verified unit tests for the report generation and configuration functionalities pass without issues.
- **Manual Testing**:
    - Manually tested the report generation process using various Pydantic models and HTML templates.


src/guidellm/config/base.py
```diff
@@ -0,0 +1,113 @@
...
+
+class Environment(str, Enum):
+    """
+    Enum for the supported environments
+    """
+
+    LOCAL = "local"
+    DEV = "dev"
+    STAGING = "staging"
+    PROD = "prod"
```
parfeniukink, line 25: What about [this approach](https://github.com/neuralmagic/guidellm/blob/e1ec99ac85ac9f6a7ee4b497ccaaf227a119889d/src/guidellm/core/serializable.py#L13) here?

src/guidellm/config/base.py
```diff
@@ -0,0 +1,113 @@
...
+    ```
+
+    """
+
+    model_config = SettingsConfigDict(
+        env_prefix="GUIDELLM__",
+        env_nested_delimiter="__",
+        extra="ignore",
+        validate_default=True,
+    )
```
parfeniukink: I still think we can use .env btw. It does not break anything. Reading from the environment of the Shell session has the highest priority.

src/guidellm/utils/constants.py
```diff
@@ -17,3 +24,9 @@
 PREFERRED_DATA_SPLITS = ["test", "validation", "train"]
 
 STANDARD_SLEEP_INTERVAL = 0.1
+
+STANDARD_REQUEST_TIMEOUT = 30
```
parfeniukink: What about having this in settings as `settings.requests.timeout` along with `settings.requests.sleep_interval`?

tests/unit/config/test_base.py
```diff
@@ -0,0 +1,88 @@
...
+    )
+
+
+@pytest.mark.unit
+def test_settings_from_env_variables(monkeypatch):
+    monkeypatch.setenv("GUIDELLM__env", "dev")
+    monkeypatch.setenv("GUIDELLM__logging__disabled", "true")
+    monkeypatch.setenv("GUIDELLM__OPENAI__API_KEY", "test_key")
+    monkeypatch.setenv("GUIDELLM__OPENAI__BASE_URL", "http://test.url")
+    monkeypatch.setenv("GUIDELLM__REPORT_GENERATION__SOURCE", "http://custom.url")
```
parfeniukink: https://docs.python.org/3/library/unittest.mock.html#patch-dict

```suggestion
@pytest.mark.unit
def test_settings_from_env_variables(mocker):
    mocker.path.dict(os.environ, {"GUIDELLM__ENV": "dev", ...})
```

pyproject.toml
```diff
@@ -48,6 +48,7 @@ dev = [
     "pytest~=8.2.2",
     "pytest-cov~=5.0.0",
     "pytest-mock~=3.14.0",
+    "requests-mock~=1.12.1",
```
parfeniukink, line 51: Do we need it?

src/guidellm/config/base.py
```diff
@@ -0,0 +1,113 @@
...
+            "report_generation"
+        ].get("source"):
+            values["report_generation"]["source"] = ENV_REPORT_MAPPING.get(env)
+        elif (
+            isinstance(values["report_generation"], ReportGenerationSettings)
+            and not values["report_generation"].source
+        ):
+            values["report_generation"].source = ENV_REPORT_MAPPING[env]
+
+        return values
```
parfeniukink: The ReportGenerationSettings class has only a single configuration. Couldn't we just do next?

```py
@field_validator("report_generation", mode="after")  # mode is optional in that case I believe
def set_default_source(self, report_generation: ReportGenerationSettings):
    report_generation.source = ENV_REPORT_MAPPING[self.env]
    return report_generation
```

parfeniukink (approved):
Nice!

src/guidellm/config/base.py
markurtz: Sounds good, will add env file back in

src/guidellm/utils/constants.py
markurtz: That's fair, it would be nice to easily control that with env variables, will update these

tests/unit/config/test_base.py
markurtz: good catch, updated

src/guidellm/config/base.py
markurtz: I had it incorrect here and it should have had a classmethod decorator, so we can't access the current instance with self. But, this is fair in that we can validate after with model_validator so we have access to env and report_generation and check if the source is an empty string, if it is, then we should set it from the report mapping. Updating to that

pyproject.toml
markurtz: yes, added in tests for the injector code that relies on mocking the requests library

src/guidellm/config/base.py
markurtz: Need to add that approach back in, and I think it works well for that use case, but with Pydantic we don't need to deal with the constructors and it auto adds the choices to make sure they match rather than us adding some more configuration to a field

PR #29, opened 2024-08-22
✨Deepsparse Backend implementation

parfeniukink:
### Summary

The Deepsparse Backend interface is implemented.
- Optional dependency is specified in `pyproject.toml` file
- Some `ruff` errors are suppressed
- `DeepsparseBackend` gets configurations from CLI, environment or defaults
- `settings.py::DeepsparseSettings` includes all the settings
- `tests/unit/backend/deepsparse.py` includes unit tests
    - `TestTextGenerationPipeline` mocks the `deepsparse.pipeline.Pipeline`
    - tests are skipped if the Python version is not in a range between Python3.8 and Python3.11 (including).

### Usage

This is an example of a command you can use in your terminal:
* `--data=openai_humaneval`: determines the dataset
* `--model=/local/path/my_model`: determines the local path to the model object. If not specified - the env variable will be used.
```sh
python -m src.guidellm.main --data=openai_humaneval --max-requests=1 --max-seconds=20 --rate-type=constant --rate=1.0 --backend=deepsparse --model=/local-path
```
#### Environment configuration
The model could also be set with `GUIDELLM__LLM_MODEL`. If the **_CLI value_** or **_environment variable_** is not set, then the default will be used. Currently, the default model is: `mistralai/Mistral-7B-Instruct-v0.3`.


src/guidellm/backend/deepsparse/backend.py
```diff
@@ -15,11 +17,30 @@ class DeepsparseBackend(Backend):
...
+    def _get_model(self, model_from_cli: Optional[str] = None) -> str:
+        """Provides the model by the next priority list:
+        1. from function argument (comes from CLI)
+        1. from environment variable
+        2. `self.default_model` from `self.available_models`
+        """
+
+        if model_from_cli is not None:
+            return model_from_cli
+        elif settings.deepsprase.model is not None:
```
mgoin: deepsprase should be deepsparse in a few places

src/guidellm/backend/deepsparse/backend.py
```diff
@@ -1,10 +1,12 @@
-from typing import Any, AsyncGenerator, List, Optional
+import os
+from typing import Any, AsyncGenerator, Dict, List, Optional
 
-from deepsparse import Pipeline
+from deepsparse import Pipeline, TextGeneration
```
mgoin: Does this mean deepsparse is a dep of guidellm? We should keep it optional at most IMO, so could this be in a try catch with an informational message to install?

parfeniukink: @mgoin Could you please check [this file](https://github.com/neuralmagic/guidellm/pull/29/files#diff-86e6dcc39e1c774e18c8e3a99a1e7c7f63399a786842ca8a6b2cba48e2c270aa)? Is it kind of a validation you are talking about?

parfeniukink: So, since the `backend/deepsparse/__init__.py` exists we determine it as a module so before you reach the `deepsparse/backend.py` you are going to run such a validation.

Also, all the imports are like `from guidellm.backend import Backend, BackendEngine` but if you need the deepsparse just `from .guidellm.backend.deepsparse import DeepsparseBackend` which runs this validation.

Also the `deepsparse` is an optional dependency. At least [it is in optional section](https://github.com/neuralmagic/guidellm/pull/29/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711)

src/guidellm/backend/deepsparse/backend.py
parfeniukink: Oh, nice catch. Thanks!

PR #46, opened 2024-09-04
Add summary metrics to saved json file

anmarques:
At the end of the run GuideLLM will print summary metrics that are computed from the raw results, but these are not currently saved anywhere.

This PR adds these metrics as serializable properties of the `TextGenerationBenchmark` class. Most of the metrics are already declared as properties, but are not serializable. Adding the `@computed_field` decorator is enough in these cases. Other properties were added to complete the list.

### DIFF
--- a/src/guidellm/core/report.py
+++ b/src/guidellm/core/report.py
@@ -147,19 +147,15 @@ def _create_benchmark_report_data_tokens_summary(
     for benchmark in report.benchmarks_sorted:
         table.add_row(
             _benchmark_rate_id(benchmark),
-            f"{benchmark.prompt_token_distribution.mean:.2f}",
+            f"{benchmark.prompt_token:.2f}",
             ", ".join(
                 f"{percentile:.1f}"
-                for percentile in benchmark.prompt_token_distribution.percentiles(
-                    [1, 5, 50, 95, 99]
-                )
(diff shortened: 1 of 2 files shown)


EricAntidot:
This is a nice addition we are expecting this to be merged.

EricAntidot:
@anmarques In order to calm down the mypy check you should add `# type: ignore[misc]` to the `@computed_field` annotation line.
cf: https://github.com/python/mypy/issues/1362
or pydantic doc as well: https://docs.pydantic.dev/2.0/usage/computed_fields/

anmarques:
> @anmarques In order to calm down the mypy check you should add `# type: ignore[misc]` to the `@computed_field` annotation line. cf: [python/mypy#1362](https://github.com/python/mypy/issues/1362) or pydantic doc as well: https://docs.pydantic.dev/2.0/usage/computed_fields/

Thanks for the tip!

rgreenberg1 (approved):
LGTM, thanks Alex!

mgoin (approved):
Thanks Alex! Could you add a short example to the readme how to get the summary metrics from the json file, hopefully in csv format?

anmarques:
I don't have instructions to convert to csv readily available. Will look into it in a following PR. Will land this one to unblock anyone needing the information saved.

PR #48, opened 2024-09-06
[Dataset]: Iterate through benchmark dataset once

parfeniukink:
issue: https://github.com/neuralmagic/guidellm/issues/44


parfeniukink:
@philschmid Hey there. Could you install the version of the branch `parfeniukink/issues/44` and tell me if works for you?

tests/unit/test_main.py
```diff
@@ -252,3 +252,54 @@ def test_generate_benchmark_report_cli_smoke(
     assert "Benchmarks" in result.output
     assert "Generating report..." in result.output
     assert "Benchmark Report 1" in result.output
+
```
markurtz, line 293: We should add in a test case here for the main pathway to ensure transformers/file dataset inputs work with the "dataset" max_requests type and they execute the expected number of requests

src/guidellm/request/transformers.py
```diff
@@ -1,9 +1,9 @@
 from pathlib import Path
 from typing import Optional, Union
 
-from datasets import (  # type: ignore  # noqa: PGH003
+from datasets import (
     Dataset,
-    DatasetDict,
+    DatasetDict,  # type: ignore  # noqa: PGH003
```
markurtz: With the mypy overrides added in, the type: ignore should be able to be removed now, correct?

markurtz (changes requested):
Looks good, minor NIT on removing typing overrides and we should add in a test case for the main pathway for this.

Two additional points on testing:
- For automation, we should add in a quick test to ensure for the sweep case that we looped the dataset correctly and on start of a new benchmark it is the first item from the dataset -- can be merged in with the previous main tests
- For manual, please validate locally that calling with a hugging face dataset as well as a simple local csv file (can download the csv from the files for the linked dataset) will work properly
    - Running a constant rate will execute the exact number of requests as in the dataset along with the first and last prompts sent matching the first and last in the dataset
    - Running a sweep with a smaller (10 prompt) csv file will properly cycle and each benchmark run starts with the same beginning prompt and ends with the end prompt

src/guidellm/request/transformers.py
parfeniukink: Oh, yeah!

philschmid:
Hey Guys, any idea on when we could get this merged?

markurtz (approved):

PR #51, opened 2024-09-09
FIX TGI example

philschmid:


philschmid:
@eldarkurtic made some error when creating it.

Issue #52, opened 2024-09-09
[bug] Model without system prompt fail


philschmid:
Hello, 

When trying to benchmark [google/gemma-2-9b-it](https://huggingface.co/google/gemma-2-9b-it) i ran into an issue. It seems like `guidellm` uses `[system](https://github.com/neuralmagic/guidellm/blob/f7947110ae1446ce765d6019cc4ab0a31d335923/src/guidellm/backend/openai.py#L109)` message to sent request, which leads to 
```bash
Template error: syntax error: System role not supported (in <string>:1)
```

I tried to change to `user` but this leads to `pydantic` validation errors. Trying to debug this but any pointers or help is appreciated.

philschmid:
Nope I worked from the fork i had and it worked, will confirm from `main`

philschmid:
Tested on commit works as well.

PR #53, opened 2024-09-09
Change default "role" from "system" to "user"

philschmid:
This is a PR for #52. I am not sure if there are other implications but with that change it works for Gemma 2 9b IT.


markurtz (approved):
Looks good, thanks for the contribution @philschmid!

PR #60, opened 2024-09-12
38 guidance reportjson from default flow is very large

parfeniukink:


parfeniukink:
@markurtz Could you provide fields that we are going to expose in the output CSV file?

markurtz:
Closing this out as the core is now provided in #91. We'll include further fixes for the report generation types in a later PR / scope

PR #61, opened 2024-09-12
Docs Updates 

rgreenberg1:
Summary:

This pull request introduces the GuideLLM CLI guide, README enhancements, image uploads, and the supported backends documentation to highlight all the backends that can be used with GuideLLM.

Test Cases:
The GuideLLM CLI has been tested with various LLM models and backends.
Unit tests ensure core functionalities work as expected.

Documentation:
Created documentation detailing the GuideLLM CLI usage and output metrics.
Created documentation detailing the openai-compatible API/HTTP pathway for TGI, llama.cpp, and DeepSparse in supported_backends.md

Additional Information:
The pull request includes changes to the docs/guides directory for the CLI documentation.
Binary files containing performance summary visualizations are added to the docs/assets directory.

Please review and provide feedback.


docs/guides/cli.md
```diff
@@ -1 +1,176 @@
...
+  --model "neuralmagic/Meta-Llama-3.1-8B-Instruct-quantized.w4a16"
+```
+
+The above command will begin the evaluation and output progress updates similar to the following: <img src="https://github.com/neuralmagic/guidellm/blob/main/docs/assets/sample-benchmark.gif" />
+
+Notes:
+
+- The `--target` flag specifies the server hosting the model. In this case, it is a local vLLM server.
+- The `--model` flag specifies the model to evaluate. The model name should match the name of the model deployed on the server
+- By default, GuideLLM will run a `sweep` of performance evaluations across different request rates, each lasting 120 seconds. The results will be saved to a local directory.
```
parfeniukink, line 40: 1. I would rename `flag` to `parameter` since our CLI supports both: parameters and flags. If you specify a flag - there is no value next to it. If you specify parameter - the value is requied then.

2. In some cases we may get an error if the tokenizer is not specified. I would add another item here. Text is below:

- The `--tokenizer` parameter specifies the tokenizer to encount the number of tokens in the dataset. If you  faced any issues try using `--tokenizer neuralmagic/Meta-Llama-3.1-8B-quantized.w8a8`.

parfeniukink (approved):
Great job!

PR #66, opened 2024-11-04
Multimodal

anmarques:
This PR adds support for benchmarking multimodal models.

It mostly extends existing infrastructure to add support to requests containing images. For emulated requests it downloads images from an illustrated version from Pride and Prejudice and randomly selects from them.

The load_images logic is currently limited to download from url. It should be extended to HF datasets or local files in the future.


rgreenberg1:
This is image models only right? text-to-image only?

sjmonson:
Obsolete

PR #71, opened 2025-02-12
Concurrent load generation option is implemented

parfeniukink:
### Summary

- `--rate-type concurrent` is implemented as an additional Profile Generation mode.
    - if `--rate-type` is concurrent you must set the `--rate` which belongs to the number of concurrent workers that will be started to simulate multiple users working with the system.


### Detailed

- You can find the `consistent` type of Load Generation, which differs from the rest of the generation modes since the `consistent` does not use times to simulate delays. All the requests go one by one within a single asynchronous task.


### Example of usage

```sh
$ guidellm --target http://localhost:8080/v1 --model Phi-3-mini-4k-instruct-q4.gguf --data 'prompt_tokens=128,generated_tokens=128' --data-type emulated --tokenizer "hf-internal-testing/llama-tokenizer" --max-requests 2 --rate-type concurrent --rate 2

╭─ Benchmarks ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ [10:43:43]   100% consistent   (0.13 req/sec avg)                                                                                                                                                             │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
  Generating report... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ (1/1) [ 0:00:29 < 0:00:00 ]
╭─ GuideLLM Benchmarks Report (stdout) ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ ╭─ Benchmark Report 1 ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮ │
│ │ Backend(type=openai_server, target=http://localhost:8080/v1, model=Phi-3-mini-4k-instruct-q4.gguf)                                                                                                        │ │
│ │ Data(type=emulated, source=prompt_tokens=128,generated_tokens=128, tokenizer=hf-internal-testing/llama-tokenizer)                                                                                         │ │
│ │ Rate(type=concurrent, rate=(2.0,))                                                                                                                                                                        │ │
│ │ Limits(max_number=2 requests, max_duration=120 sec)                                                                                                                                                       │ │
│ │                                                                                                                                                                                                           │ │
│ │                                                                                                                                                                                                           │ │
│ │ Requests Data by Benchmark                                                                                                                                                                                │ │
│ │ ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━┓                                                                                                   │ │
│ │ ┃ Benchmark                 ┃ Requests Completed ┃ Request Failed ┃ Duration  ┃ Start Time ┃ End Time ┃                                                                                                   │ │
│ │ ┡━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━┩                                                                                                   │ │
│ │ │ asynchronous@2.00 req/sec │ 4/4                │ 0/4            │ 29.74 sec │ 10:43:43   │ 10:44:12 │                                                                                                   │ │
│ │ └───────────────────────────┴────────────────────┴────────────────┴───────────┴────────────┴──────────┘                                                                                                   │ │
│ │                                                                                                                                                                                                           │ │
│ │ Tokens Data by Benchmark                                                                                                                                                                                  │ │
│ │ ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓                                                                                   │ │
│ │ ┃ Benchmark                 ┃ Prompt ┃ Prompt (1%, 5%, 50%, 95%, 99%)    ┃ Output ┃ Output (1%, 5%, 50%, 95%, 99%)    ┃                                                                                   │ │
│ │ ┡━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩                                                                                   │ │
│ │ │ asynchronous@2.00 req/sec │ 128.00 │ 128.0, 128.0, 128.0, 128.0, 128.0 │ 128.00 │ 128.0, 128.0, 128.0, 128.0, 128.0 │                                                                                   │ │
│ │ └───────────────────────────┴────────┴───────────────────────────────────┴────────┴───────────────────────────────────┘                                                                                   │ │
│ │                                                                                                                                                                                                           │ │
│ │ Performance Stats by Benchmark                                                                                                                                                                            │ │
│ │ ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓ │ │
│ │ ┃                           ┃ Request Latency [1%, 5%, 10%, 50%, 90%, 95%, 99%]      ┃ Time to First Token [1%, 5%, 10%, 50%, 90%, 95%, 99%]   ┃ Inter Token Latency [1%, 5%, 10%, 50%, 90% 95%, 99%]   ┃ │ │
│ │ ┃ Benchmark                 ┃ (sec)                                                  ┃ (ms)                                                    ┃ (ms)                                                   ┃ │ │
│ │ ┡━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩ │ │
│ │ │ asynchronous@2.00 req/sec │ 7.91, 8.79, 9.89, 18.70, 27.52, 28.62, 29.51           │ 1146.2, 2053.7, 3188.0, 12159.8, 20962.7, 22060.9,      │ 48.2, 49.1, 49.7, 51.3, 54.8, 57.3, 66.9               │ │ │
│ │ │                           │                                                        │ 22939.4                                                 │                                                        │ │ │
│ │ └───────────────────────────┴────────────────────────────────────────────────────────┴─────────────────────────────────────────────────────────┴────────────────────────────────────────────────────────┘ │ │
│ │                                                                                                                                                                                                           │ │
│ │ Performance Summary by Benchmark                                                                                                                                                                          │ │
│ │ ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━┓                                                               │ │
│ │ ┃ Benchmark                 ┃ Requests per Second ┃ Request Latency ┃ Time to First Token ┃ Inter Token Latency ┃ Output Token Throughput ┃                                                               │ │
│ │ ┡━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━┩                                                               │ │
│ │ │ asynchronous@2.00 req/sec │ 0.13 req/sec        │ 18.71 sec       │ 12099.47 ms         │ 52.00 ms            │ 17.22 tokens/sec        │                                                               │ │
│ │ └───────────────────────────┴─────────────────────┴─────────────────┴─────────────────────┴─────────────────────┴─────────────────────────┘                                                               │ │
│ ╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

```

### DIFF
--- a/src/guidellm/executor/profile_generator.py
+++ b/src/guidellm/executor/profile_generator.py
@@ -17,7 +17,12 @@ __all__ = [
 ]
 
 ProfileGenerationMode = Literal[
-    "sweep", "synchronous", "throughput", "constant", "poisson"
+    "sweep",
+    "synchronous",
+    "throughput",
+    "constant",
+    "poisson",
+    "concurrent",
 ]
 
 
@@ -61,7 +66,7 @@ class ProfileGenerator:
             logger.error(err)
             raise err
 
-        self._mode = mode
+        self._mode: ProfileGenerationMode = mode
 
         if self._mode in ("sweep", "throughput", "synchronous"):
             if rate is not None:
@@ -135,7 +140,7 @@ class ProfileGenerator:
         return self._generated_count
 
     @property
-    def profile_generation_modes(self) -> Sequence[ProfileGenerationMode]:
+    def profile_generation_modes(self) -> List[ProfileGenerationMode]:
         """
         Return the list of profile modes to be run in the report.
 
@@ -147,7 +152,8 @@ class ProfileGenerator:
                 settings.num_sweep_profiles
             )
 
-        if self._mode in ["throughput", "synchronous"]:
+        # WIP: think about moving this concurrent above
+        if self._mode in ["throughput", "synchronous", "concurrent"]:
             return [self._mode]
 
         if self._rates is None:
@@ -155,6 +161,13 @@ class ProfileGenerator:
 
         if self._mode in ["constant", "poisson"]:
             return [self._mode] * len(self._rates)
+
+        # WIP
+        # if self._mode in ["concurrent"]:
+        #     if self._rates is None:
+        #         raise ValueError("rate ")
+
+        #     return [self._mode] * int(self._rates[0])
 
         raise ValueError(f"Invalid mode: {self._mode}")
 
@@ -173,13 +186,13 @@ class ProfileGenerator:
             current_report,
         )
 
-        if self.mode in ["constant", "poisson"]:
+        if self.mode in ("constant", "poisson", "concurrent"):
             if not self.rates:
                 err = ValueError(f"Rates are required for {self.mode} mode")
                 logger.error(err)
                 raise err
 
-            profile = self.create_fixed_rate_profile(
+            profile: Optional[Profile] = self.create_fixed_rate_profile(
                 self.generated_count,
                 self.mode,
                 self.rates,
@@ -229,9 +242,11 @@ class ProfileGenerator:
         :return: The generated profile or None if index is out of range.
         :rtype: Optional[Profile]
         """
+
         modes_map: Dict[str, LoadGenerationMode] = {
             "constant": "constant",
             "poisson": "poisson",
+            "concurrent": "consistent",
         }
 
         if mode not in modes_map:
--- a/src/guidellm/executor/profile_generator.py
+++ b/src/guidellm/executor/profile_generator.py
@@ -162,13 +162,6 @@ class ProfileGenerator:
         if self._mode in ["constant", "poisson"]:
             return [self._mode] * len(self._rates)
 
-        # WIP
-        # if self._mode in ["concurrent"]:
-        #     if self._rates is None:
-        #         raise ValueError("rate ")
-
-        #     return [self._mode] * int(self._rates[0])
-
         raise ValueError(f"Invalid mode: {self._mode}")
 
     def next(self, current_report: TextGenerationBenchmarkReport) -> Optional[Profile]:
--- a/src/guidellm/scheduler/base.py
+++ b/src/guidellm/scheduler/base.py
@@ -316,7 +316,8 @@ class Scheduler:
         if self.mode == "consistent":
             if self.rate is None:
                 raise ValueError(
-                    "The rate must be specified in order to provide concurrent execution"
+                    "The `rate` must be specified in order to provide "
+                    "the concurrent execution"
                 )
             for index, request in enumerate(self.generator):
                 while (index + 1 - completed) >= settings.max_concurrency:
--- a/src/guidellm/scheduler/base.py
+++ b/src/guidellm/scheduler/base.py
@@ -114,12 +114,12 @@ class Scheduler:
             logger.error(err)
             raise err
 
-        self._generator = generator
-        self._worker = worker
-        self._mode = mode
-        self._rate = rate
-        self._max_number = max_number
-        self._max_duration = max_duration
+        self._generator: RequestGenerator = generator
+        self._worker: Backend = worker
+        self._mode: LoadGenerationMode = mode
+        self._rate: Optional[float] = rate
+        self._max_number: Optional[int] = max_number
+        self._max_duration: Optional[float] = max_duration
 
         self._load_generator = LoadGenerator(mode, rate)
 
@@ -227,9 +227,7 @@ class Scheduler:
         count_total = (
             self.max_number
             if self.max_number
-            else round(self.max_duration)
-            if self.max_duration
-            else 0
+            else round(self.max_duration) if self.max_duration else 0
         )
 
         # yield initial result for progress tracking
@@ -246,9 +244,7 @@ class Scheduler:
             count_completed = (
                 min(run_count, self.max_number)
                 if self.max_number
-                else round(time.time() - start_time)
-                if self.max_duration
-                else 0
+                else round(time.time() - start_time) if self.max_duration else 0
             )
 
             yield SchedulerResult(
@@ -267,9 +263,7 @@ class Scheduler:
             count_completed=(
                 benchmark.request_count + benchmark.error_count
                 if self.max_number
-                else round(time.time() - start_time)
-                if self.max_duration
-                else 0
+                else round(time.time() - start_time) if self.max_duration else 0
             ),
             benchmark=benchmark,
         )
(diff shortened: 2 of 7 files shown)


sjmonson:
I'm having trouble with this PR, it seems to hang after the first request. Tried with:

```bash
guidellm --target http://localhost:8000/v1 \
         --model ibm-granite/granite-3.1-2b-instruct \
         --data-type emulated --data "prompt_tokens=128,generated_tokens=128" \
         --rate-type concurrent --rate 256 --max-seconds 20
```

src/guidellm/scheduler/base.py
```diff
@@ -298,42 +294,81 @@ async def _run_sync(
...
-
-            def _completed(_task: asyncio.Task) -> None:
-                nonlocal completed
-                completed += 1
-                _res = _task.result()
-
-                if _res:
-                    benchmark.request_completed(_res)
-                    logger.debug("Request completed: {}", _res)
+        if self.mode == "consistent":
```
markurtz, line 316: Given the longer if else here and specific logic for handling, I think it would make sense to break it out into its own function to handle the fixed concurrent request values rather than merging here if we go this route

src/guidellm/scheduler/base.py
```diff
@@ -298,42 +294,81 @@ async def _run_sync(
...
+            if self.rate is None:
+                raise ValueError(
+                    "The `rate` must be specified in order to provide "
+                    "the concurrent execution"
+                )
+            for index, request in enumerate(self.generator):
+                while (index + 1 - completed) >= settings.max_concurrency:
```
markurtz, line 323: Shouldn't we ignore the max_concurrency setting in this case / raise an error if the concurrent requests number is smaller than the rate type we need to run at? 

Additionally, our check should be on the rate here and we would only allow a new request if we're below our rate or have passed our time / request count restrictions, right?

src/guidellm/scheduler/base.py
```diff
@@ -298,42 +294,81 @@ async def _run_sync(
...
+                if index >= max_number or time.time() >= end_time:
+                    break
+
+                logger.debug(f"Running concurrently request={request}")
+
+                benchmark.request_started()
+
+                # Create multiple concurrent tasks
+                tasks: list[asyncio.Task] = []
+                for _ in range(int(self.rate)):
```
markurtz, line 335: I'm not following how this is going to keep the concurrent request count fixed. Due to the outer loop, this would always just max out around max concurrency. I say around, because this inner for loop enables it to go above the max concurrency by appending N=rate new requests before checking max concurrency again

src/guidellm/executor/profile_generator.py
```diff
@@ -229,9 +235,11 @@ def create_fixed_rate_profile(
         :return: The generated profile or None if index is out of range.
         :rtype: Optional[Profile]
         """
+
         modes_map: Dict[str, LoadGenerationMode] = {
             "constant": "constant",
             "poisson": "poisson",
+            "concurrent": "consistent",
```
markurtz, line 242: any rationale for remapping this to consistent rather than keeping it as concurrent through the entire codebase?

markurtz (changes requested):
@parfeniukink I'm not following how this will satisfy the goal of given a request or time limit, keep a consistent number of concurrent requests open at the targeted number. Currently, it looks like it will max out at max_concurrency from settings plus the rate passed in due to the inner and outer loop. 

I think a simple implementation in here would be a new function for _run_concurrent which keeps the number of requests open equal to the rate passed in within the scheduler and should raise an exception if that rate is greater than the settings max concurrency.

A more complicated implementation would be to figure out how to pass either max concurrency from the load generator or have load generator aware of how many requests have been completed and then enable a new iteration once ones have completed.

src/guidellm/executor/profile_generator.py
parfeniukink: ### TLDR:
Makes sense. Will change it to `concurrent`

### Detailed
Well, I would say that `consistent` is related to the `LoadGenerationMode`, which differs from the `ProfileGenerationMode`, which includes the `concurrent` mode.
This LoadGeneration will produce a new request immediately, so I called this `consistent`. Like we produce consistent load generation.

src/guidellm/scheduler/base.py
parfeniukink: got it

rgreenberg1:
#96  this refactor gives us concurrent support.

PR #72, opened 2025-02-17
`--workers` CLI parameter is added

parfeniukink:
* new CLI parameter allows to control concurrent workers that are generating requests

### DIFF
--- a/src/guidellm/executor/base.py
+++ b/src/guidellm/executor/base.py
@@ -73,13 +73,25 @@ class Executor:
         rate: Optional[Union[float, Sequence[float]]] = None,
         max_number: Optional[int] = None,
         max_duration: Optional[float] = None,
+        workers: int = 1,
     ):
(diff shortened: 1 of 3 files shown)


markurtz (changes requested):
@parfeniukink I'm not following the need to add in workers here. We have the configuration settings value for this which is passed through: https://github.com/neuralmagic/guidellm/blob/main/src/guidellm/config.py#L145. What do we gain in adding this?

parfeniukink:
> @parfeniukink I'm not following the need to add in workers here. We have the configuration settings value for this which is passed through: https://github.com/neuralmagic/guidellm/blob/main/src/guidellm/config.py#L145. What do we gain in adding this and how does it fix the problem listed in the max concurrency off by 1 issue?

@markurtz yes, you are correct. This is intermediate implementation that comes out from the discussion. Closing it

PR #78, opened 2025-02-26
Add backend_kwargs to generate_benchmark_report

jackcook:
This PR adds support for passing additional keyword arguments to the backend used in the `generate_benchmark_report` function. This allows us to use the already-implemented [`request_args` parameter](https://github.com/neuralmagic/guidellm/blob/ecf2984645202eea022cc7c9f245947e2dd5457d/src/guidellm/backend/openai.py#L37) on `OpenAIBackend`, which is needed in order to pass [custom headers/query params/etc](https://github.com/openai/openai-python/blob/3e69750d47df4f0759d4a28ddc68e4b38756d9ca/src/openai/resources/beta/chat/completions.py#L100) to OpenAI chat completion requests.

### DIFF
--- a/src/guidellm/main.py
+++ b/src/guidellm/main.py
@@ -1,5 +1,5 @@
 import asyncio
-from typing import Literal, Optional, Union, get_args
+from typing import Any, Literal, Mapping, Optional, Union, get_args
 
 import click
 from loguru import logger
@@ -197,6 +197,7 @@ def generate_benchmark_report(
     max_requests: Union[Literal["dataset"], int, None],
     output_path: str,
     cont_refresh_table: bool,
+    backend_kwargs: Mapping[str, Any] = {},
 ) -> GuidanceReport:
     """
     Generate a benchmark report for a specified backend and dataset.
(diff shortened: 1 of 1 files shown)


src/guidellm/main.py
```diff
@@ -197,6 +197,7 @@ def generate_benchmark_report(
     max_requests: Union[Literal["dataset"], int, None],
     output_path: str,
     cont_refresh_table: bool,
+    backend_kwargs: Mapping[str, Any] = {},
```
sjmonson: Can you redo this to avoid [mutable arguments](https://github.com/satwikkansal/wtfpython?tab=readme-ov-file#-beware-of-default-mutable-arguments)? Something like the following with a `None` check in the function.

```suggestion
    backend_kwargs: Optional[Mapping[str, Any]] = None,
```

sjmonson (changes requested):

sjmonson (approved):
LGTM

markurtz (approved):
This looks great, thanks for the contribution!

PR #80, opened 2025-02-27
Replace busy wait in async loop with a Semaphore

### LINKED ISSUE #70: GUIDELLM__MAX_CONCURRENCY is off by 1
The number of concurrent requests in throughput mode is always one less then `Settings.max_concurrency`.

**Example**

```sh
export GUIDELLM__MAX_CONCURRENCY="2"

guidellm --target http://localhost:8000/v1 \
         --model meta-llama/Llama-3.2-3B \
         --data-type emulated \
         --data prompt_tokens=512,generated_tokens=2048 \
         --rate-type throughput \
         --max-seconds 300
```

*Observe from server side that number of requests in queue is 1*

sjmonson:
Use a semaphore to concurrency limit the scheduling of new tasks rather than a busy wait. Additionally reduce async sleep at the end of the loop iteration as that is sufficient to cause preemption. As a side effect #70 is fixed by this PR since the offending code has been reworked.

Closes #70

### DIFF
--- a/src/guidellm/scheduler/base.py
+++ b/src/guidellm/scheduler/base.py
@@ -299,13 +299,13 @@ class Scheduler:
         self, benchmark: TextGenerationBenchmark, end_time: float, max_number: float
     ) -> AsyncGenerator[Union[TextGenerationResult, TextGenerationError], None]:
         tasks = []
-        completed = 0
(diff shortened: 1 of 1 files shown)


markurtz (approved):
This looks great!

PR #87, opened 2025-03-05
Set up data for ui - WIP

DaltheCow:
This is a roughly thrown together version of my understanding of how guidellm data can populate the UI. It isn't tied into the html injection functionality which will serve the UI (as I'm just generating json I manually drop into the UI for now). Nor is it set up to be the basis of the api that will serve the guidellm UI, which would be where this logic would likely fit in the future. But the calculations are meant to be 100% correct/accurate.

I'll add more tests when I finish the frontend work to help walk through what this is trying to achieve. For now attention is most needed on the calculations for benchmark metrics, prompt/output token metrics, requests over time, etc, which are used to make histograms and line charts in the [UI](https://www.figma.com/design/smh6PEYHdva7Fdwyn8BzW4/GuideLLM?node-id=4034-105736&t=x4SdnX4BhFLGFM6x-4).

I've attached the data generated from this code in files below, and the model I've used thus far (microsoft/DialoGPT-small) is not a great chat model afaik so it probably isn't ideal data, but the performance is a little better running on my mac due to its small size which I think will make for more realistic looking plots on the charts. Will have to get a realistic set up to output better data soon.

I don't have a goal of trying to get this merged in urgently, there is a bit more UI work to do before this is useful. But I'll be pestering people for more reviews soon.


src/guidellm/core/result.py
```diff
@@ -403,6 +403,22 @@ def output_token_throughput(self) -> float:
...
+        :rtype: Distribution
+        """
+        throughputs = []
+        for r in self.results:
+            duration = (r.end_time or 0) - (r.start_time or 0)
+            if duration > 0:
+                throughputs.append(r.output_token_count / duration)
+
+        return Distribution(data=throughputs)
+
```
DaltheCow, line 418: The UI relies on the output throughput distribution, and I didn't find any methods/properties that were in the token/(unit of time) shape the UI expects so I added this.

src/guidellm/main.py
```diff
@@ -285,6 +285,8 @@ def generate_benchmark_report(
     )
     report = asyncio.run(_run_executor_for_result(executor))
 
+    generate_ui_api_data(report)
+
```
DaltheCow: This is just so I can run this simply and look at the json generated

src/guidellm/utils/generate_ui_data.py
```diff
@@ -0,0 +1,154 @@
...
+                idx = n_buckets - 1
+            bucket_counts[idx] += 1
+
+        buckets = []
+        for i, count in enumerate(bucket_counts):
+            bucket_start = minv + i * bucket_width
+            buckets.append({
+                "value": bucket_start,
+                "count": count
+            })
```
DaltheCow, line 46: I am not sure the proper way to generate these buckets or if there is code somewhere else in guidellm that could manage this and I missed it.

But this code assumes we have a set number of buckets we want to generate and then determines the bucket width based off of that. It is a hard coded approach, and some data analysis first might result in a better number of buckets or bucket size. But generally I figured the UI would look good with there being a set number of buckets so our histograms conveniently look the same and take up a comfortable amount of space.

src/guidellm/utils/generate_ui_data.py
```diff
@@ -0,0 +1,154 @@
...
+    workload_details_json = generate_workload_details(report)
+    benchmarks_json = generate_benchmarks_json(report.benchmarks)
+    os.makedirs("ben_test", exist_ok=True)
+    # generate json files based off of api specs, https://codepen.io/dalthecow/pen/bNGVQbq, for consumption by UI
+    with open("ben_test/run_info.json", "w") as f:
+        json.dump(run_info_json, f, indent=2)
+    with open("ben_test/workload_details.json", "w") as f:
+        json.dump(workload_details_json, f, indent=2)
+    with open("ben_test/benchmarks.json", "w") as f:
+        json.dump(benchmarks_json, f, indent=2)
```
DaltheCow: This is just for testing purposes, to view the generated json.

DaltheCow:
What the data looks like currently.

[benchmarks.json](https://github.com/user-attachments/files/19080825/benchmarks.json)
[run_info.json](https://github.com/user-attachments/files/19080826/run_info.json)
[workload_details.json](https://github.com/user-attachments/files/19080827/workload_details.json)

Issue #81, opened 2025-02-27
Base enablement to enable a dataset configuration file to be passed in rather than standard args, allow alias's for the folder structure


rgreenberg1:
**Description:**
We will need to add support for GuideLLM to take in a dataset configuration profile as an argument (with the details of i/o tokens, for example) that can be used for benchmarking with GuideLLM. These dataset profiles and push them upstream to a folder under GuideLLM to then be pulled in and used as reference datasets for GuideLLM. 

**User Story:**
As a ml developer, I want to be able to provide standard dataset profiles, targeted at different use cases (RAG, Summarization, Chat...) and use that profile as a benchmarking target to better understand what my inference performance will look like for my model/use case when I move to production. 

**Additional Documentation:**

https://docs.google.com/document/d/1Ql6RI3_LbhQxqFCIu_n2A2CewK6OVVFZFYgu4KEjjtw/edit?usp=sharing 

**Acceptance criteria:**
Extend a new argument format for GuideLLM to accept a dataset profile as an argument. 
Dataset profile should have the following extensions and formats to be valid as arguments: 
- Text - .txt
- Audio - .wav
- image - .png or .jpg

markurtz:
@sjmonson you're taking this task on, correct?

Aside from that, general thoughts from my side:
1. We should set it up such that the API pathways support passing in configs / aliases for those configs into the APIs. Ex: guidellm --scenario <ALIAS | CONFIG_PATH>.
2. The aliases that are passed in will resolve based on the name of the config that would be stored within the code base such as src/guidellm/benchmark/scenarios
3. Within each config, it contains all required information for fully configuring the scenario including pointing at an HF dataset, locally stored file, a file accessible over HTTP, or synthetic data configuration. For the different dataset pathways, it should support the args for constructing the dataset; ex for synthetic data generation it should contain the source text, prompt tokens with distribution settings. It should specify how to decide on the number of output tokens / output token distribution to use including loading from the dataset, specifying the statistics for the output token distribution, or None. Finally, we should likely include a random seed for reproducibility pathways.

For point 3, a lot of that will become easier with the planned data pipelines rework with expansion to multi modality. Additionally, we'll need to consider what this is going to look like for expansion to multi modal as well.

PR #99, opened 2025-03-19
Add support for "benchmarking scenarios"

sjmonson:
This PR adds support for "scenarios" that allow specifying benchmark argument in a file / as a single Pydantic object. CLI argument defaults are loaded from the scenario object defaults to give benchmark-as-code users the same defaults as CLI. Argument values in the CLI follow the following precedence: `Scenario (class defaults) < Scenario (CLI provided Scenario) < CLI Arguments`.

Closes:
* #81


markurtz:
@sjmonson I like the direction of this. A few quick thoughts from my side:
- I think using Pydantic's built-in support for dotenv files along the lines of what we're doing for the general settings might work well here, and we can store the different scenarios then as just .env files and also enable potentially a default .env file: https://docs.pydantic.dev/latest/concepts/pydantic_settings/#dotenv-env-support
- I'd like to see if we could get a unified setup for validation of values, types, and auto conversion that click is handling currently standardized through this, so all entry point pathways go through the same validations and fail early
- It also would be great if we could get the Python API entrypoint and the CLI then pulling from the same defaults and able to support the same pathways

Not sure if there's something out there already to automatically handle the last two points, but if so, that would be a great inclusion

markurtz:
@sjmonson, for the last one, yes, something else. Specifically the entrypoint for benchmarking Python API here: https://github.com/neuralmagic/guidellm/blob/main/src/guidellm/benchmark/entrypoints.py#L22. Towards @anmarques's previous issues with needing to set all argument values when not using the CLI. Also would help towards suporting both CLI and Python API for the scenarios

sjmonson:
Ah I see. The workflow I imagined for @anmarques's use-case is to call the higher-level entrypoint [`benchmark_with_scenario`](https://github.com/sjmonson/guidellm/blob/ce4c73baea3de4da00ac65bdcf2e10c8744ba09b/src/guidellm/benchmark/entrypoints.py#L25-L33) since it handles the defaults on scenario instantiation so a call like the following would be the same as calling the CLI with only the required args:

```python
result = await benchmark_with_scenario(
    GenerativeTextScenario(
        target="http://localhost:8000",
        data={
            "prompt_tokens": 128,
            "output_tokens": 128,
        },
        rate_type="sweep",
    ),
    output_path="output.json",
)
```

markurtz (approved):

PR #117, opened 2025-04-16
Output support expansion, code hygiene, and tests

markurtz:
Expand outputs with CSV support and a more friendly table output for copying. Additionally, add code docs and tests for the output files and fix any bugs that popped up


src/guidellm/benchmark/profile.py
```diff
@@ -327,6 +327,14 @@ def from_standard_args(  # type: ignore[override]
         if "sweep_size" in kwargs:
             raise ValueError("Sweep size must not be provided, use rate instead.")
 
+        if isinstance(rate, Sequence):
+            if len(rate) != 1:
+                raise ValueError(
+                    "Rate must be a single value for sweep profile, received "
+                    f"{len(rate)} values."
+                )
+            rate = rate[0]
```
sjmonson, line 336: Note, the CLI parser always return a list so I think this check will not work properly.

https://github.com/neuralmagic/guidellm/blob/cb4bf2ed9116d9e34757d3f9f2832a2e25185e6d/src/guidellm/__main__.py#L26-L37

markurtz: List extends from Sequence which enables the __getitem__ needed for the indexing call. Fair to call out that we could hardcode that to list, but based on Python docs, they prefer checking on Sequence rather than list: https://docs.python.org/3/library/typing.html#typing.List

sjmonson: Oh my bad I just didn't read the code well enough, missed the `if len(rate) != 1` check.

sjmonson (approved):

PR #118, opened 2025-04-17
v0.2.0 Version Update and Docs Expansions

markurtz:
Update version to 0.2.0 and expand docs to account for the latest changes along with backfill of missing docs


pyproject.toml
```diff
@@ -17,7 +17,7 @@ include = ["*"]
 
 [project]
 name = "guidellm"
-version = "0.1.0"
+version = "0.2.0"
 description = "Guidance platform for deploying and managing large language models."
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.8.0,<4.0"
```
sjmonson: Before release we should bump this to `>=3.9.0,<4.0` to avoid PyPi building a 3.8 package.

sjmonson (approved):
LGTM

sjmonson (changes requested):
See above

pyproject.toml
markurtz: Good catch! Given the below, I'm going to say we follow up with a separate diff to fix this -- I think you had mentioned you hit this before as well

Screenshot 2025-04-17 at 3 56 42 PM

README.md
```diff
@@ -38,7 +38,13 @@ Before installing, ensure you have the following prerequisites:
 - OS: Linux or MacOS
 - Python: 3.9 – 3.13
```
jennyyangyi-magic, line 39: Should installing vllm or a link to that be part of the pre-requisite?

markurtz:
> See above

@sjmonson, any issues with pushing the update to 3.9 in the toml file off to a follow up diff so we can keep this one focused on docs?

README.md
```diff
@@ -58,93 +64,94 @@ vllm serve "neuralmagic/Meta-Llama-3.1-8B-Instruct-quantized.w4a16"
...
+  - `output_tokens`: Average number of tokens for outputs.
+  - `TYPE_stdev`, `TYPE_min`, `TYPE_max`: Standard deviation, minimum, and maximum values for the specified type (e.g., `prompt_tokens`, `output_tokens`). If not provided, will use the provided tokens value only.
+  - `samples`: Number of samples to generate, defaults to 1000.
+  - `source`: Source text data for generation, defaults to a local copy of Pride and Prejudice.
+
+- `--data-args`: A JSON string used to specify the columns to source data from (e.g., `prompt_column`, `output_tokens_count_column`) and additional arguments to pass into the HuggingFace datasets constructor.
+
+- `--data-sampler`: Enables applying `random` shuffling or sampling to the dataset. If not set, no sampling is applied.
+
+- `--rate-type`: Defines the type of benchmark to run. Supported types include:
```
jennyyangyi-magic: I'd just add here. "Default is sweep unless the following types are declared:"

docs/slos_and_slas.md
rgreenberg1, line 1: Can we kill the SLA talk and just focus on SLOs? SLA's get into contractual agreements - SLOs are more general imo

rgreenberg1, line 1: How did you come up w/ these being different numbers?

SLOs:
TTFT: ≤ 200ms for 99% of requests
ITL: ≤ 50ms for 99% of requests
SLAs:
Guarantee TTFT ≤ 250ms for 95% of requests

README.md
markurtz: I wouldn't put it as a prereq since GuideLLM itself doesn't require it and the server may be running on another system. Added some more in the getting started as well as the detailed backend pieces, though

rgreenberg1 (approved):

PR #120, opened 2025-04-17
Fix issue if async task count does not evenly divide accross process pool

sjmonson:
If `SchedulingStrategy.processing_requests_limit` does not evenly divide over `SchedulingStrategy.processes_limit` then we will either run too few requests or run "0" requests which will cause and unlimited number of request to be scheduled at a time.

### DIFF
--- a/src/guidellm/scheduler/worker.py
+++ b/src/guidellm/scheduler/worker.py
@@ -220,24 +220,27 @@ class RequestsWorker(ABC, Generic[RequestT, ResponseT]):
         self,
         requests_queue: multiprocessing.Queue,
         results_queue: multiprocessing.Queue,
-        max_concurrency: Optional[int],
+        max_concurrency: int,
         process_id: int,
     ):
         async def _process_runner():
-            pending = asyncio.Semaphore(max_concurrency) if max_concurrency else None
+            pending = asyncio.Semaphore(max_concurrency)
+
+            if pending.locked():
+                raise ValueError(
+                    "Async worker called with max_concurrency < 1"
+                )
 
             while (
                 process_request := await self.get_request(requests_queue)
             ) is not None:
                 dequeued_time = time.time()
 
-                if pending:
-                    await pending.acquire()
+                await pending.acquire()
 
                 def _task_done(_: asyncio.Task):
                     nonlocal pending
-                    if pending:
-                        pending.release()
+                    pending.release()
 
                 task = asyncio.create_task(
                     self.resolve_scheduler_request(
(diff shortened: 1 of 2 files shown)


src/guidellm/scheduler/strategy.py
```diff
@@ -94,6 +95,23 @@ def processing_requests_limit(self) -> int:
         """
         return settings.max_concurrency
 
+    @property
+    def process_requests_limits(self) -> List[int]:
```
markurtz: I think we're going to hit issues introducing another property here. For example, the ConcurrentStrategy isn't updated in this diff. For derivative formulas from the other properties, I think that this logic might be better in the _start_processes Scheduler function or as a function that works on the base class. For the latter, we'll end up exporting this data, which could be a good or a bad thing. Also, if we update it to something like the following, we won't have to have the dead else statement and instead can just loop normally and always launch a process:

```
num_processes = min(scheduling_strategy.processes_limit, scheduling_strategy.processing_requests_limit)
requests_limit_split = scheduling_strategy.processing_requests_limit // scheduling_strategy.processes_limit
requests_limit_remain = scheduling_strategy.processing_requests_limit % scheduling_strategy.processes_limit
process_ids = [id_ for id_ in range(num_processes)]
process_requests_limits = [requests_limit_split + 1 if ind < requests_limit_remain else requests_limit_split for find in range(num_processes)]

for id_, requests_limit in zip(process_ids, process_requests_limits):
...
```

sjmonson: Yeah I think that makes much more sense. `ConcurrentStrategy` does inherit this method from `SchedulingStrategy` but its kinda of unclear. Your suggestion also has the advantage of moving all of the logic into the scheduler rather then spiting it across various sections.

src/guidellm/scheduler/worker.py
```diff
@@ -224,7 +224,10 @@ def process_loop_asynchronous(
         process_id: int,
     ):
         async def _process_runner():
-            pending = asyncio.Semaphore(max_concurrency) if max_concurrency else None
+            if max_concurrency is not None:
+                pending = asyncio.Semaphore(max_concurrency)
+            else:
+                pending = None
```
sjmonson: @markurtz Is there a reason max_concurrency is optional, we don't support it in the current scheduler? Also I removed the bounds check here, though my concern is if `max_concurrency` is ever `< 1` in some future iteration of the code the failure will be hard to detect (will just hang on the lock acquisition).

markurtz: Good question, I believe this was a relic from a few different iterations over the top of it. Everything should be passing in valid values through the stack now, so I think it's safe to remove and code in a check at the start that max_concurrency is a counting number

markurtz (approved):

PR #123, opened 2025-04-17
Add Containerized Benchmarking Support for GuideLLM

### LINKED ISSUE #111: [Nice to Have] Container builds
It would be nice to have packages as container images to quickly get up to using guidellm and also reduces the volume of potential issues faced by users if we have a ready to go container available. 

Ref: [llm-load-test](https://github.com/orgs/openshift-psap/packages?repo_name=llm-load-test)

### LINKED ISSUE #119: Add Docker and Kubernetes Support for GuideLLM Benchmarking
We need a dockerfile to enable running GuideLLM benchmarks in a containerized environment and as a Kubernetes job. This would be helpful for users who want to run benchmarks in a more controlled and scalable environment.

The first change should include:

**Dockerfile**
- Based on Python 3.11 slim image
- Installs GuideLLM and its dependencies
- Entrypoint to run benchmarks with environment variables

wangchen615:
# Add Containerized Benchmarking Support for GuideLLM

## Description
This PR adds containerization support for GuideLLM benchmarking, making it easier to run benchmarks in containerized environments. The changes include:

1. A new `Dockerfile` that:
   - Uses Python 3.11 slim as the base image
   - Installs necessary dependencies, including git and curl
   - Sets up a non-root user for security
   - Includes healthcheck functionality
   - Properly configures the benchmark script

2. A new `run_benchmark.sh` script that:
   - Provides a flexible way to run benchmarks with configurable parameters
   - Supports environment variable overrides for customization
   - Handles different output formats (json, yaml)
   - Includes proper error handling and logging

## Testing
The container can be built and tested using:
```bash
docker build -t <container-image-repo>/guidellm-benchmark:latest .
docker run -e TARGET="http://localhost:8000" -e MODEL="neuralmagic/Meta-Llama-3.1-8B-Instruct-quantized.w4a16" <container-image-repo>/guidellm-benchmark:latest
```

## Configuration Options
The container supports the following environment variables:
- `TARGET`: The target endpoint for benchmarking (default: http://localhost:8000)
- `MODEL`: The model to benchmark (default: neuralmagic/Meta-Llama-3.1-8B-Instruct-quantized.w4a16)
- `RATE_TYPE`: The rate type for benchmarking (default: sweep)
- `DATA`: The data configuration (default: prompt_tokens=256,output_tokens=128)
- `MAX_REQUESTS`: Maximum number of requests (default: 100)
- `MAX_SECONDS`: Maximum duration in seconds
- `OUTPUT_PATH`: Path for output files (default: /results/guidellm_benchmark_results)
- `OUTPUT_FORMAT`: Output format (json, yaml, or yml)

## Security Considerations
- Uses a non-root user for running the benchmark
- Includes proper file permissions
- Implements healthcheck for container monitoring
- Follows container security best practices

## Related Issues
- Closes #119 
- Closes #111 

## Checklist
- [x] Added Dockerfile with proper security configurations
- [x] Added run_benchmark.sh script with environment variable support
- [x] Added documentation for configuration options
- [x] Tested container build and run
- [x] Followed security best practices


sjmonson:
Could you put these files under a `/build/` directory? Also why python 3.11 and not 3.13 since its the latest version we offer support for?

Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+# Set working directory
+WORKDIR /app
+
+# Install GuideLLM
+RUN pip install git+https://github.com/neuralmagic/guidellm.git
+
+# Copy and set up the benchmark script
+COPY run_benchmark.sh /app/
+RUN chmod +x /app/run_benchmark.sh
```
surajssd: I think this can be avoided if you do this on the code file itself, so you don't have to do it again inside the container.

Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+# Create non-root user
+RUN useradd -m -u 1000 guidellm
+
+# Set working directory
+WORKDIR /app
+
+# Install GuideLLM
+RUN pip install git+https://github.com/neuralmagic/guidellm.git
```
surajssd: Could you merge the multiple RUN commands into one to reduce the number of layers?

```suggestion
# Set working directory
WORKDIR /app

# Copy the benchmark script
COPY run_benchmark.sh /app/

# Create non-root user
RUN useradd -m -u 1000 guidellm && \
    # Install GuideLLM
    pip install git+https://github.com/neuralmagic/guidellm.git && \
    # Set ownership to non-root user
    RUN chown -R guidellm:guidellm /app
```

run_benchmark.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
surajssd:
```suggestion
#!/usr/bin/env bash

set -euo pipefail
```

This ensures that the script fails on any failing command.

Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+# Set ownership to non-root user
+RUN chown -R guidellm:guidellm /app
+
+# Switch to non-root user
+USER guidellm
+
+# Healthcheck
+HEALTHCHECK --interval=30s --timeout=30s --start-period=5s --retries=3 \
+    CMD curl -f http://localhost:8000/health || exit 1
+
```
sjmonson: I assume this was an accidental inclusion since GuideLLM does not provide a web server.

```suggestion
```

Dockerfile
sjmonson: Rather than installing main, install our current HEAD. Also `cp` from within the container to avoid having two COPY layers.

```suggestion
COPY . /source
RUN pip install /source && \
         cp /source/build/run_benchmark.sh /app/
```

run_benchmark.sh
sjmonson, line 1: If this script is going to be the ENTRYPOINT and not the CMD than it needs to be more flexible. Either add an extra arguments variable or append command line arguments to the command.

sjmonson (changes requested):

run_benchmark.sh
wangchen615: Yes, I allow it to add extra arguments.

Dockerfile
wangchen615: Yes, I remove this in the new build/Dockerfile.

build/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM python:3.12-slim
```
markurtz: I think it would be better to have this at one of the extremes for the supported versions rather than somewhere in the middle, especially since the max and min version are tested much more; ie 3.9 or 3.13

build/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+LABEL org.opencontainers.image.source="https://github.com/neuralmagic/guidellm"
+LABEL org.opencontainers.image.description="GuideLLM Benchmark Container"
```
markurtz: NIT: can we remove the benchmark from this so we keep it flexible towards future goals of evals and things like that?

build/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+# Install dependencies and set up environment in a single layer
+RUN apt-get update && apt-get install -y \
+    git \
+    curl \
+    && pip install git+https://github.com/neuralmagic/guidellm.git \
```
markurtz: I'm concerned that this is too restrictive in that it always builds from the latest main, so it will be tough to specify what should be built here other than from the latest main either with the build system for a specific release or for a user. 

Can we expand this with a build-arg for the package version to build? Which we can set as default to build from the latest main, but otherwise can utilize it to specify the release tag/branch to build from.

build/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+# Switch to non-root user
+USER guidellm
+
+# Healthcheck
+HEALTHCHECK --interval=30s --timeout=30s --start-period=5s --retries=3 \
+    CMD curl -f http://localhost:8000/health || exit 1
+
+# Set the entrypoint
+ENTRYPOINT ["/app/run_benchmark.sh"]
```
markurtz: Can we change this out so that we are setting the entrypoint to "guidellm" or "guidellm benchmark"? We can then set the CMD after to include any specific args we need passed in / surface from the settings or set it to --help for users to get started with. This way we can enable and passthrough and remove duplication of arguments we need to pass through and another script to maintain and add functionality to.

build/run_benchmark.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
```
markurtz: See note on the entrypoint for the Dockerfile, I'd like to see if we can remove this script fully or make it optional to package in

build/Dockerfile
markurtz: Also, what are our plans for supporting the range of python versions? Anything we should do here to create specific Dockerfiles for each supported python version?

markurtz (changes requested):
Let some comments throughout. Additionally, I know we had a request to move it to build/, but this is a commonly named folder that python and potential future builds will use. 

I've generally seen the parent folder be docker/ at the root and then different python versions could be either nested folders under that with a Dockerfile or multiple Dockerfiles a the root with extensions of the python version such as Dockerfile.py39

build/Dockerfile
sjmonson: Normally I would recommend handling this either though build args or postfixes on the Containerfile name. E.g.

```suggestion
ARG python_version=3.13

FROM python:${python_version}-slim
```

Or name this file something like `Containerfile.python-3.13`

build/run_benchmark.sh
sjmonson: In latest version, the entrypoint will fall back to executing unwrapped guidellm if anything is set in CMD. The entrypoint is now mostly a stopgap until scenarios with env support land.

https://github.com/wangchen615/guidellm/blob/51c6508a1a117417a897d7b0907bad7d325cac0f/build/entrypoint.sh#L8-L11

build/Dockerfile
sjmonson: I think benchmark is a neutral term for both performance and evaluations.

markurtz (approved):

PR #137, opened 2025-04-25
Add docs for data/datasets and how to configure them in GuideLLM

### LINKED ISSUE #133: Documentation regarding the DS format to be fed to GuideLLM
# There is a need in writing down examples for all DSs formats that can be fed as an input to the GuideLLM benchmark.

Tried using the example in the current README - 
`guidellm benchmark --target "http://${IP}:${PORT}" --rate-type sweep --max-seconds 30 --data "prompt_tokens=256,output_tokens=128"` - 
Got an error for autoTokenizer not known. After debugging a little thought that it might be that tokenizers for Mistarl and Llama3 aren't supported yet (used VLLM instances with those loaded), and the error occurs since processor is required for synthetic data bencmarking.  

Then tried using a HF model ID with - 
`guidellm benchmark --target "http://${IP}:${PORT}" --rate-type synchronous --max-seconds 30 --data ${HF_DS_ID} --data-args '{"prompt_column": "prompt"}'` and the benchmark ran but got an error in openAIHTTPBackend for every request, which made me think that the data format I'm sending isn't correct.

then tried to used a local JSON file as a DS and ran with - 
`guidellm benchmark --target "http://${IP}:${PORT}" --rate-type synchronous --max-seconds 30 --data "prompts.json"`
ad got a `TypeError: 'PosixPath' object is not iterable` error

Would appreciate help in running GuideLLM benchmarks with different DSs configurations.

markurtz:
Fixes #133

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,7 @@ Our comprehensive documentation offers detailed guides and resources to help you
 
 - [**Installation Guide**](https://github.com/neuralmagic/guidellm/tree/main/docs/install.md) - This guide provides step-by-step instructions for installing GuideLLM, including prerequisites and setup tips.
 - [**Backends Guide**](https://github.com/neuralmagic/guidellm/tree/main/docs/backends.md) - A comprehensive overview of supported backends and how to set them up for use with GuideLLM.
+- [**Data/Datasets Guide**](https://github.com/neuralmagic/guidellm/tree/main/docs/datasets.md) - Information on supported datasets, including how to use them for benchmarking.
 - [**Metrics Guide**](https://github.com/neuralmagic/guidellm/tree/main/docs/metrics.md) - Detailed explanations of the metrics used in GuideLLM, including definitions and how to interpret them.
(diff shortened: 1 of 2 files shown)


SharonGil:
@markurtz Thank you very much for this, this simplifies and makes things much clearer.

2 small comments from my end:

1. In lines 27 and 142 - you put suffix .ext - just wanted to make sure this is intentionally as a place holder for some external file or a typo instead of .txt.
2. As of the data formats themselves, its written "Ensure the file format matches the expected structure for the dataset" - for HF model maybe to add a short elaboration regarding the 'dataset' library format expected, in order to understand which of HF models are supported and can be used and which aren't, rather than just trial and error...
Also for the file based DS format - there are few examples there for .txt, .csv and .json files - do these represent the only formats supported for these files? Meaning - for .txt files the data should be plain newline delimited strings that represent the prompts without and keys added and without output, but for .csv for example you have to give a prompt and an output columns?

SharonGil:
@markurtz That's perfect. Thanks a lot for the clatifications!

PR #113, opened 2025-04-12
Fix units for Req Latency in output to seconds

smalleni:
`request_latency()` returns time in seconds not ms.

PR #144, opened 2025-05-01
Fix Units of Req Latency

sjmonson:
This was previously fixed in #113 but was overwritten at some point.


sjmonson:
Duplicate of #142

PR #150, opened 2025-05-07
Add the ability to pass in extra body params for the OpenAI pathway, similar to the extra query params that were added

markurtz:


MeganEFlynn (approved):
Works for setting temperature and max tokens!

PR #157, opened 2025-05-09
set up next.js app and static build

DaltheCow:
This is a starter next.js app set up for static builds (static site generation comes from output: 'export' in the next.config.ts). It uses the 'use client' directive to tell Next.js to treat everything as client side code in the page.tsx file meaning that most of the app won't be prerendered. This mirrors our set up since we use MUI theming and CSS in JS which require most of the app to be client code.


DaltheCow:
My main open question for this PR is should we already include base github workflows that ensure the basics work. I'll probably just add those in the next PR

sjmonson (commented):
Looks good but needs linting fixes

.gitignore
```diff
@@ -178,3 +178,6 @@ cython_debug/
 # Project specific files
 *.json
 *.yaml
+
+# Turn off UI related ignores, defer to ui/.gitignore
+!ui/**/*.json
```
markurtz: I think it would be better and easier to track if we include all ignores here rather than having nested .gitignore files. Is there any strong reason for including a separate one rather than adding a section in this root gitignore?

ui/.eslintrc.json
```diff
@@ -0,0 +1,21 @@
...
+  },
+  "extends": [
+    "eslint:recommended",
+    "next",
+    "next/core-web-vitals",
+    "plugin:@typescript-eslint/recommended"
+  ],
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "ecmaVersion": 2021,
```
markurtz: Could we bump this to a more recent version such as 2024?

ui/.eslintrc.json
```diff
@@ -0,0 +1,21 @@
+{
+  "env": {
+    "browser": true,
+    "es2021": true,
```
markurtz: Could we bump this to a more recent version matching the bump in ecmaVersion?

ui/.eslintrc.json
```diff
@@ -0,0 +1,21 @@
...
+    "node": true
+  },
+  "extends": [
```
markurtz: Can we add prettier in here as well as explicitly adding react with plugin:react/recommended? For prettier, it disables formatting rules in ESLint so prettier will handle them

ui/.eslintrc.json
```diff
@@ -0,0 +1,21 @@
...
+    "plugin:@typescript-eslint/recommended"
+  ],
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "ecmaVersion": 2021,
+    "sourceType": "module"
+  },
+  "plugins": ["@typescript-eslint"],
+  "root": true,
+  "rules": {}
```
markurtz: Can we add in "@typescript-eslint/no-unused-vars": ["warn"] to rules and double check if there's anything else we want here?

ui/.gitignore
```diff
@@ -0,0 +1,41 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
markurtz: See note in root gitignore to merge this in there

ui/README.md
```diff
@@ -0,0 +1,36 @@
+This is a [Next.js](https://nextjs.org) project bootstrapped with [`create-next-app`](https://nextjs.org/docs/app/api-reference/cli/create-next-app).
```
markurtz: I'd rather add this base info to the root md files so we can localize and standardize expectations there such as adding a section for UI to the developer file

ui/package.json
```diff
@@ -0,0 +1,25 @@
+{
```
markurtz: Can we double check if we can add package.json and package-lock.json to the root pathway so everything is immediately usable after checking out the repo including running UI commands rather than having to cd into a nested directory?

ui/.eslintrc.json
```diff
@@ -0,0 +1,21 @@
+{
```
markurtz: Can we double check if we can add this to the root pathway so everything is immediately usable after checking out the repo including running UI commands rather than having to cd into a nested directory?

ui/tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
```
markurtz: Can we double check if we can add this to the root pathway so everything is immediately usable after checking out the repo including running UI commands rather than having to cd into a nested directory?

ui/tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+        "name": "next"
+      }
+    ],
+    "paths": {
+      "@/*": ["./src/*"]
+    },
+    "resolveJsonModule": true,
+    "skipLibCheck": true,
+    "strict": true,
+    "target": "ES2017"
```
markurtz: This is a bit dated and browsers have moved up significantly since this for compatibility. Can we update this to ES2022? Details for ES2022 features support: https://caniuse.com/?feats=mdn-javascript_builtins_array_at,mdn-javascript_builtins_regexp_hasindices,mdn-javascript_builtins_object_hasown,mdn-javascript_builtins_error_cause,mdn-javascript_operators_await_top_level,mdn-javascript_classes_private_class_fields,mdn-javascript_classes_private_class_methods,mdn-javascript_classes_static_class_fields,mdn-javascript_classes_static_initialization_blocks

markurtz (changes requested):
Overall looking good. Left a few comments. Additionally, can we move the UI folder under src to keep everything for code in a central place? Specifically:
src/ui/...
src/guidellm/...

.gitignore
DaltheCow: Hmm that's a good question. I guess the main reason is that here I can blanket turn off all of the .gitignore pieces and apply only relevant things in the other one, so it keeps things focused.

Whereas if we keep them in one .gitignore then without going through the whole root level .gitignore, there are rules like ignore *.json where I'll now have to ensure that doesn't affect anything in src/ui. I'll just do it though.

ui/.eslintrc.json
DaltheCow: I was using this PR as a default state of a next.js app and just made sure the static build would function. All of the linting concerns were included in the next PR so if it's cool I'll check there and see if this is covered there

ui/.eslintrc.json
DaltheCow: Same as above comment, going to take this comment into the next PR #159 and see if it is needed. There are a lot of extensions and things added and I expect this is covered by those.

I think we had a pretty good set up in the old guidellm-ui repo, and the only thing I removed there were the airbnb linting extensions since they are not supported by certain versions and so I was in a depedency version hell, just opted to nuke them, but I think we are still in a good state. Will take this concern into that PR though

.gitignore
DaltheCow: Considering you've requested other things like package.json be moved to the root of the directory, I take it back, merging them is definitely the way forward, no reason not to.

DaltheCow:
Just a note here, but I'm looking into moving all of these things around. Namely, package.json, tsconfig.json and .eslintrc.json. Thinking about next.config.ts as well. Mostly just trying to figure out big picture which things would be better off living in the root and which things are better of staying nested in src/ui.

Due to Next.js being a framework and having certain expectations, it appears that we can't be too freeform, but it looks very possible to move package.json/package-lock.json to the root and then tsconfig.json and .eslintrc.json and other future configs we will add are optional and I don't know if there is an explicit benefit to moving them out of src/ui. But also I think it is fine to and expresses the intentions of the repo a bit better.

Next.js looks for things in certain places and I haven't yet figured out if next.config.ts should be moved to the root or not. You didn't request it, but I'm just trying to think all of this through for what makes sense. Regardless this will all work and is a little more complicated than I expected for a simple thing.

ui/.eslintrc.json
markurtz: Sounds good

ui/tsconfig.json
DaltheCow: @markurtz I'm not sure if there is any practical benefit to moving anything other than package.json/package-lock.json to the top level at this moment. That alone allows users to run things without having to navigate directories. But I did move .eslintrc.json and tsconfig.json to the root and Next.js was picky and required a tsconfig.json to be present in src/ui where next.config.ts lives so it was regenerating one. So for now I've decided to move tsconfig.json back down to src/ui. I think it is probably better to move .eslintrc.json back down as well just so that we keep things mostly together.

I think if we want to keep most of the configs at the root then it might make sense to revisit that in later PRs when more of the configs are added in and then see if there is a solution. Although it seems like Next.js is the main blocker here and will not make things very easy on us with moving folders/files around.

It is hard to know exactly what vision you are imagining, but I can see some nicities in keeping more configs at the root so it is obvious the UI is a key part of this and it isn't hidden away. But I think keeping the README.md and DEVELOPING.md docs detailed about running the UI will be enough. And of course the package.json at the root makes it easy to run things.

Let me know your thoughts about this.

markurtz (approved):

PR #159, opened 2025-05-12
Set up core framework pieces such as configs for eslint, typescript, prettier, husky precommit hook, basic GHA

DaltheCow:


sjmonson (dismissed):

package.json
```diff
@@ -5,20 +5,41 @@
...
+    "tslint-config-prettier": "^1.18.0",
+    "typescript": "^5"
+  },
+  "lint-staged": {
+    "*.js": "eslint --cache --fix",
+    "*.ts": "eslint --cache --fix",
+    "*.jsx": "eslint --cache --fix",
+    "*.tsx": "eslint --cache --fix"
   }
 }
```
markurtz, line 47: Can we add engines in here as well to standardize the node version someone is using? IE:
```
"engines": {
  "node": ">=X.X"
}
```

For the version, 22 looks like it would be a good one since it's flagged as LTS and will be maintained for awhile: https://nodejs.org/en/about/previous-releases

package.json
```diff
@@ -5,20 +5,41 @@
...
+    "eslint-config-prettier": "^8.5.0",
+    "eslint-plugin-import": "^2.29.1",
+    "eslint-plugin-jsx-a11y": "^6.6.1",
+    "eslint-plugin-no-secrets": "^1.0.2",
+    "eslint-plugin-prettier": "^5.4.0",
+    "eslint-plugin-react": "^7.31.10",
+    "eslint-plugin-react-hooks": "^5.2.0",
+    "husky": "^9.1.7",
+    "prettier": "^3.5.3",
+    "tslint-config-prettier": "^1.18.0",
```
markurtz: I think this is deprecated and wouldn't be used, good thing to double check

package.json
markurtz, line 47: Is jest or some other testing framework included in one of the later diffs? If not, let's add that in here

src/ui/.eslintignore
```diff
@@ -0,0 +1,2 @@
+node_modules
```
markurtz: Just double checking that this will resolve correctly for node_modules at the root of the project even though this is nested under UI

src/ui/.prettierignore
```diff
@@ -0,0 +1,8 @@
+/node_modules
```
markurtz: Double checking how these are supposed to resolve with the / and whether those are the root of the project, root of the users namespace, or root relative to prettierignore. I believe node_modules will be at the root of the project, but public is next to prettierignore

src/ui/.prettierrc
```diff
@@ -0,0 +1,13 @@
+{
+  "arrowParens": "always",
+  "bracketSameLine": false,
+  "bracketSpacing": true,
+  "jsxBracketSameLine": false,
```
markurtz:
```suggestion
  "bracketSameLine": false,
```

jsxBracketSameLine was deprecated I believe

src/ui/.prettierrc
```diff
@@ -0,0 +1,13 @@
...
+  "jsxSingleQuote": false,
+  "printWidth": 100,
```
markurtz:
```suggestion
  "printWidth": 88,
```

Let's set this to the same as the python code base is

src/ui/tsconfig.json
```diff
@@ -1,15 +1,18 @@
 {
   "compilerOptions": {
     "allowJs": true,
```
markurtz: Do we plan on needing any javascript files? Ideally I'd like to fix it to typescript only

src/ui/tsconfig.json
```diff
@@ -1,15 +1,18 @@
...
+    "allowSyntheticDefaultImports": true,
+    "baseUrl": "./src/",
```
markurtz: is this correct now with the changes to move app directory up one level?

markurtz (changes requested):

package.json
DaltheCow: Yes will look into this, our node types are set up with `"@types/node": "^20"` above so I have to see if there is a reason vs 22 would be an issue, hopefully not. I know some stuff I've downgraded due to conflicts but I don't remember downgrading node intentionally.

DaltheCow: That is the next PR!

src/ui/.eslintignore
DaltheCow: I'll double check, definitely haven't had any issues with node_modules being linted yet but that is a good point.

src/ui/.prettierignore
DaltheCow: Yes I'll double check!

src/ui/tsconfig.json
DaltheCow: I don't think there will ever be any need for javascript files, probably don't need, will look into this

src/ui/.eslintignore
DaltheCow: Realized eslint won't even attempt to access node_modules since it is a two directories up, so I deleted .eslintignore since it isn't needed

src/ui/.prettierignore
DaltheCow: Removed since it is 2 directories up so this isn't needed

src/ui/tsconfig.json
DaltheCow: Changed

markurtz (approved):

PR #160, opened 2025-05-13
Testing tooling

DaltheCow:
Set up unit/integration/e2e testing with test coverage for unit/integration. Test coverage is not used, but can be pulled into a GH workflow when/if desired. Currently as there are very few tests written there is no threshold set in this repo.


jaredoconnell (commented):
I am not familiar with the project, so I won't approve anything, but it's good to see progress towards unit testing.

src/ui/cypress/support/e2e.ts
```diff
@@ -0,0 +1,20 @@
+// ***********************************************************
```
markurtz: Can we clean up the boilerplate comments in here and focus it just on the pathways we are going to use and support?

src/ui/cypress/support/commands.ts
```diff
@@ -0,0 +1,37 @@
+/// <reference types="cypress" />
```
markurtz: Do we need this file for now? Not a fan of adding in a bunch of commented-out code/examples.

src/ui/jest.config.js
```diff
@@ -0,0 +1,20 @@
+const nextJest = require('next/jest');
+
+const createJestConfig = nextJest({
+  dir: __dirname,
+});
+
+const customJestConfig = {
+  collectCoverage: false,
+  collectCoverageFrom: ['tests'],
```
markurtz: This isn't going to resolve converage correctly, right? Shouldn't it be pointed at the source files?

src/ui/jest.config.js
```diff
@@ -0,0 +1,20 @@
+const nextJest = require('next/jest');
+
+const createJestConfig = nextJest({
+  dir: __dirname,
+});
+
+const customJestConfig = {
```
markurtz: Can we explicilty set `testMatch` here for where tests are located and found?

src/ui/jest.config.js
```diff
@@ -0,0 +1,20 @@
...
+  collectCoverage: false,
+  collectCoverageFrom: ['tests'],
+  coverageDirectory: './.meta',
+  coverageReporters: ['json-summary'],
```
markurtz: Can we also export a human readable format here?

markurtz (changes requested):
@DaltheCow a few minor things, but can we also see if we can nest the test pieces under guidellm/tests/ui rather than keeping them in the src folder? That way we encapsulate everything for easier navigation as well as clear standards

src/ui/jest.config.js
DaltheCow: Yeah in regards to this and the comment below I haven't done the work to set up test coverage yet. I had a set up in the control plane project I was happy enough with that I was going to look into implementing when I got around to it.

DaltheCow:
> @DaltheCow a few minor things, but can we also see if we can nest the test pieces under guidellm/tests/ui rather than keeping them in the src folder? That way we encapsulate everything for easier navigation as well as clear standards

I think this is doable, am mostly hoping I don't have to move too many configs around since that always breaks a lot of things

DaltheCow:
@markurtz since we are moving the tests from src/ui to tests/ui I'm going to go ahead and move all of the relevant configs such as .eslintrc.json, jest.config.js, cypress.config.ts, .prettierrc and the tsconfigs (tsconfig.base.json tsconfig.cypress.json tsconfig.json tsconfig.test.json) to the root. Originally I avoided this since it was going to require me to make some extension files that were just going to be there to pull in the files from the root for reference inside of src/ui, which seemed overly complicated given that we could just leave package.json at the root and avoid most of the complexity.

But it is wrong to leave configs needed for testing in src/ui when tests live elsewhere now.

I'm hoping I have a decent understanding of what this entails already but expect some more surprises along the way. Hopefully it ends up feeling better.

src/ui/jest.config.js
DaltheCow: I'm going to implement something for test coverage now and we can see if it seems right to you

DaltheCow: Okay I have added lcov which has a nice UI, and I've added test coverage badges. I have gone with the strategy of merging the test coverage of our unit and integration tests so that we can have just one set of badges. I don't know if that is something we would ever add to the readme like some repos do, we had done that before in the [sparsezoo-ui](https://github.com/neuralmagic/sparsezoo-ui). Anyways, for now test coverage would be incredibly low so probably not worth displaying like a badge of pride.

markurtz (approved):

PR #162, opened 2025-05-15
Fead/add preprocess dataset [Issue-106]

TomerG711:
Added a new command - preprocess. This command allows the users to preprocess dataset (from HF, file etc.), and limit the prompts the specific token sizes distribution. The generated dataset is saved to a local file and optionally to HF, and can later be used by GuideLLM benchmark.

### DIFF
--- a/src/guidellm/__main__.py
+++ b/src/guidellm/__main__.py
@@ -41,7 +41,6 @@ def parse_number_str(ctx, param, value):  # noqa: ARG001
 
 @click.group()
 def cli():
-    """GuideLLM CLI tools."""
     pass
 
 
@@ -58,8 +57,8 @@ def cli():
     "--backend-type",
     type=click.Choice(list(get_args(BackendType))),
     help=(
-            "The type of backend to use to run requests against. Defaults to 'openai_http'."
-            f" Supported types: {', '.join(get_args(BackendType))}"
+        "The type of backend to use to run requests against. Defaults to 'openai_http'."
+        f" Supported types: {', '.join(get_args(BackendType))}"
     ),
     default="openai_http",
 )
@@ -68,8 +67,8 @@ def cli():
     callback=parse_json,
     default=None,
     help=(
-            "A JSON string containing any arguments to pass to the backend as a "
-            "dict with **kwargs."
+        "A JSON string containing any arguments to pass to the backend as a "
+        "dict with **kwargs."
     ),
 )
 @click.option(
@@ -77,8 +76,8 @@ def cli():
     default=None,
     type=str,
     help=(
-            "The ID of the model to benchmark within the backend. "
-            "If None provided (default), then it will use the first model available."
+        "The ID of the model to benchmark within the backend. "
+        "If None provided (default), then it will use the first model available."
     ),
 )
 @click.option(
@@ -86,9 +85,9 @@ def cli():
     default=None,
     type=str,
     help=(
-            "The processor or tokenizer to use to calculate token counts for statistics "
-            "and synthetic data generation. If None provided (default), will load "
-            "using the model arg, if needed."
+        "The processor or tokenizer to use to calculate token counts for statistics "
+        "and synthetic data generation. If None provided (default), will load "
+        "using the model arg, if needed."
     ),
 )
 @click.option(
@@ -96,8 +95,8 @@ def cli():
     default=None,
     callback=parse_json,
     help=(
-            "A JSON string containing any arguments to pass to the processor constructor "
-            "as a dict with **kwargs."
+        "A JSON string containing any arguments to pass to the processor constructor "
+        "as a dict with **kwargs."
     ),
 )
 @click.option(
@@ -105,17 +104,17 @@ def cli():
     required=True,
     type=str,
     help=(
-            "The HuggingFace dataset ID, a path to a HuggingFace dataset, "
-            "a path to a data file csv, json, jsonl, or txt, "
-            "or a synthetic data config as a json or key=value string."
+        "The HuggingFace dataset ID, a path to a HuggingFace dataset, "
+        "a path to a data file csv, json, jsonl, or txt, "
+        "or a synthetic data config as a json or key=value string."
     ),
 )
 @click.option(
     "--data-args",
     callback=parse_json,
     help=(
-            "A JSON string containing any arguments to pass to the dataset creation "
-            "as a dict with **kwargs."
+        "A JSON string containing any arguments to pass to the dataset creation "
+        "as a dict with **kwargs."
     ),
 )
 @click.option(
@@ -123,8 +122,8 @@ def cli():
     default=None,
     type=click.Choice(["random"]),
     help=(
-            "The data sampler type to use. 'random' will add a random shuffle on the data. "
-            "Defaults to None"
+        "The data sampler type to use. 'random' will add a random shuffle on the data. "
+        "Defaults to None"
     ),
 )
 @click.option(
@@ -132,8 +131,8 @@ def cli():
     required=True,
     type=click.Choice(STRATEGY_PROFILE_CHOICES),
     help=(
-            "The type of benchmark to run. "
-            f"Supported types {', '.join(STRATEGY_PROFILE_CHOICES)}. "
+        "The type of benchmark to run. "
+        f"Supported types {', '.join(STRATEGY_PROFILE_CHOICES)}. "
     ),
 )
 @click.option(
@@ -141,28 +140,28 @@ def cli():
     default=None,
     callback=parse_number_str,
     help=(
-            "The rates to run the benchmark at. "
-            "Can be a single number or a comma-separated list of numbers. "
-            "For rate-type=sweep, this is the number of benchmarks it runs in the sweep. "
-            "For rate-type=concurrent, this is the number of concurrent requests. "
-            "For rate-type=async,constant,poisson, this is the rate requests per second. "
-            "For rate-type=synchronous,throughput, this must not be set."
+        "The rates to run the benchmark at. "
+        "Can be a single number or a comma-separated list of numbers. "
+        "For rate-type=sweep, this is the number of benchmarks it runs in the sweep. "
+        "For rate-type=concurrent, this is the number of concurrent requests. "
+        "For rate-type=async,constant,poisson, this is the rate requests per second. "
+        "For rate-type=synchronous,throughput, this must not be set."
     ),
 )
 @click.option(
     "--max-seconds",
     type=float,
     help=(
-            "The maximum number of seconds each benchmark can run for. "
-            "If None, will run until max_requests or the data is exhausted."
+        "The maximum number of seconds each benchmark can run for. "
+        "If None, will run until max_requests or the data is exhausted."
     ),
 )
 @click.option(
     "--max-requests",
     type=int,
     help=(
-            "The maximum number of requests each benchmark can run for. "
-            "If None, will run until max_seconds or the data is exhausted."
+        "The maximum number of requests each benchmark can run for. "
+        "If None, will run until max_seconds or the data is exhausted."
     ),
 )
 @click.option(
@@ -170,18 +169,18 @@ def cli():
     type=float,
     default=None,
     help=(
-            "The percent of the benchmark (based on max-seconds, max-requets, "
-            "or lenth of dataset) to run as a warmup and not include in the final results. "
-            "Defaults to None."
+        "The percent of the benchmark (based on max-seconds, max-requets, "
+        "or lenth of dataset) to run as a warmup and not include in the final results. "
+        "Defaults to None."
     ),
 )
 @click.option(
     "--cooldown-percent",
     type=float,
     help=(
-            "The percent of the benchmark (based on max-seconds, max-requets, or lenth "
-            "of dataset) to run as a cooldown and not include in the final results. "
-            "Defaults to None."
+        "The percent of the benchmark (based on max-seconds, max-requets, or lenth "
+        "of dataset) to run as a cooldown and not include in the final results. "
+        "Defaults to None."
     ),
 )
 @click.option(
@@ -204,10 +203,10 @@ def cli():
     type=click.Path(),
     default=Path.cwd() / "benchmarks.json",
     help=(
-            "The path to save the output to. If it is a directory, "
-            "it will save benchmarks.json under it. "
-            "Otherwise, json, yaml, or csv files are supported for output types "
-            "which will be read from the extension for the file path."
+        "The path to save the output to. If it is a directory, "
+        "it will save benchmarks.json under it. "
+        "Otherwise, json, yaml, or csv files are supported for output types "
+        "which will be read from the extension for the file path."
     ),
 )
 @click.option(
@@ -219,8 +218,8 @@ def cli():
     "--output-sampling",
     type=int,
     help=(
-            "The number of samples to save in the output file. "
-            "If None (default), will save all samples."
+        "The number of samples to save in the output file. "
+        "If None (default), will save all samples."
     ),
     default=None,
 )
@@ -231,28 +230,28 @@ def cli():
     help="The random seed to use for benchmarking to ensure reproducibility.",
 )
 def benchmark(
-        target,
-        backend_type,
-        backend_args,
-        model,
-        processor,
-        processor_args,
-        data,
-        data_args,
-        data_sampler,
-        rate_type,
-        rate,
-        max_seconds,
-        max_requests,
-        warmup_percent,
-        cooldown_percent,
-        disable_progress,
-        display_scheduler_stats,
-        disable_console_outputs,
-        output_path,
-        output_extras,
-        output_sampling,
-        random_seed,
+    target,
+    backend_type,
+    backend_args,
+    model,
+    processor,
+    processor_args,
+    data,
+    data_args,
+    data_sampler,
+    rate_type,
+    rate,
+    max_seconds,
+    max_requests,
+    warmup_percent,
+    cooldown_percent,
+    disable_progress,
+    display_scheduler_stats,
+    disable_console_outputs,
+    output_path,
+    output_extras,
+    output_sampling,
+    random_seed,
 ):
     asyncio.run(
         benchmark_generative_text(
@@ -284,8 +283,8 @@ def benchmark(
 
 @cli.command(
     help=(
-            "Print out the available configuration settings that can be set "
-            "through environment variables."
+        "Print out the available configuration settings that can be set "
+        "through environment variables."
     )
 )
 def config():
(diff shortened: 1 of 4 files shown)


src/guidellm/__main__.py
```diff
@@ -290,5 +291,187 @@ def config():
     print_config()
 
 
+@cli.group(help="Preprocessing utilities for datasets.")
```
markurtz: NIT: with the preprocess global namespace for the commands, can we make the help statement here a bit more generic so it isn't specific to datasets?

src/guidellm/__main__.py
```diff
@@ -290,5 +291,187 @@ def config():
...
+def preprocess():
+    pass
+
+
+@preprocess.command(
+    help="Convert a dataset to have specific prompt and output token sizes.\n\n"
```
markurtz: Nit, add parentheses around the string statement here to better match code styling in the repo

src/guidellm/__main__.py
```diff
@@ -290,5 +291,187 @@ def config():
...
+    "INPUT_DATA: Path to the input dataset or dataset ID.\n"
+    "OUTPUT_PATH: Directory to save the converted dataset. "
+    "The dataset will be saved as an Arrow dataset (.arrow) inside the directory."
+)
+@click.argument(
+    "input_data",
+    type=str,
+    metavar="INPUT_DATA",
```
markurtz: I believe click's default is to uppercase the argument param, so by default it should evaluate to the same value here

src/guidellm/__main__.py
```diff
@@ -290,5 +291,187 @@ def config():
...
+    required=True,
+)
+@click.argument(
+    "output_path",
+    type=click.Path(file_okay=True, dir_okay=False, writable=True, resolve_path=True),
+    metavar="OUTPUT_PATH",
```
markurtz: I believe click's default is to uppercase the argument param, so by default it should evaluate to the same value here

src/guidellm/__main__.py
```diff
@@ -290,5 +291,187 @@ def config():
...
+
+
+@preprocess.command(
+    help="Convert a dataset to have specific prompt and output token sizes.\n\n"
+    "INPUT_DATA: Path to the input dataset or dataset ID.\n"
+    "OUTPUT_PATH: Directory to save the converted dataset. "
+    "The dataset will be saved as an Arrow dataset (.arrow) inside the directory."
+)
+@click.argument(
+    "input_data",
```
markurtz: NIT: could we call this `data` or `input` to simplify naming especially since we're already under `guidellm preprocess data`. Along those lines, it might make sense to have this as a position arg then since this is the core argument for the functionality

src/guidellm/__main__.py
```diff
@@ -290,5 +291,187 @@ def config():
...
+)
+@click.option(
+    "--short-prompt-strategy",
+    type=click.Choice([s.value for s in ShortPromptStrategy]),
+    default=ShortPromptStrategy.IGNORE.value,
+    show_default=True,
+    help="Strategy to handle prompts shorter than the target length. ",
+)
+@click.option(
+    "--pad-token",
```
markurtz: NIT: pad-token here could be taken to mean the token_id. Could we change this out for character or str?

src/guidellm/__main__.py
```diff
@@ -290,5 +291,187 @@ def config():
...
+    help="Minimum number of prompt tokens.",
+)
+@click.option(
+    "--prompt-tokens-max",
+    type=int,
+    default=None,
+    help="Maximum number of prompt tokens.",
+)
+@click.option(
+    "--prompt-random-seed",
```
markurtz: NIT: I think we generally should only ask the user to set one random seed rather than a specific one for prompt and output. If we need those to generate differently, then auto incrementing the seed they give us should be fine. Enables easier reproducibility and fewer arguments

src/guidellm/__main__.py
```diff
@@ -290,5 +291,187 @@ def config():
...
+    show_default=True,
+    help="Strategy to handle prompts shorter than the target length. ",
+)
+@click.option(
+    "--pad-token",
+    type=str,
+    default=None,
+    help="The token to pad short prompts with when using the 'pad' strategy.",
+)
+@click.option(
```
markurtz, line 371: NIT: these are a lot of args and duplicates across prompt and output. Does it make sense to support a JSON or CSV schema that we parse into a pydantic class that is shared by the prompt and output token settings here to simplify? Also enables the user to type less with something like: --prompt-tokens "mean=10,stdev=5,min=3,max=20"

src/guidellm/preprocess/dataset.py
```diff
@@ -0,0 +1,224 @@
...
+    tokenizer: PreTrainedTokenizerBase,
+    pad_token: str,
+    **_kwargs,
+) -> str:
+    while len(tokenizer.encode(current_prompt)) < min_prompt_tokens:
+        current_prompt += pad_token
+    return current_prompt
+
+
+STRATEGY_HANDLERS: dict[ShortPromptStrategy, Callable] = {
```
markurtz, line 140: We should support an error here as well as an option

src/guidellm/preprocess/dataset.py
```diff
@@ -0,0 +1,224 @@
...
+    ShortPromptStrategy.IGNORE: handle_ignore_strategy,
+    ShortPromptStrategy.CONCATENATE: handle_concatenate_strategy,
+    ShortPromptStrategy.PAD: handle_pad_strategy,
+}
+
+
+def save_dataset_to_file(dataset: Dataset, output_path: Union[str, Path]) -> None:
```
markurtz: NIT: for these can we ensure we're doing a lower() as well for the suffixes?

src/guidellm/preprocess/dataset.py
```diff
@@ -0,0 +1,224 @@
...
+    output_path = Path(output_path)
+    output_path.parent.mkdir(parents=True, exist_ok=True)
+
+    if output_path.suffix == ".csv":
+        dataset.to_csv(str(output_path))
+    elif output_path.suffix == ".json":
```
markurtz:
```suggestion
    elif output_path.suffix == ".json" or output_path.suffix == ".jsonl":
```

And need to add to supported types dict as well

src/guidellm/preprocess/dataset.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def handle_pad_strategy(
+    current_prompt: str,
+    min_prompt_tokens: int,
+    tokenizer: PreTrainedTokenizerBase,
+    pad_token: str,
+    **_kwargs,
+) -> str:
+    while len(tokenizer.encode(current_prompt)) < min_prompt_tokens:
```
markurtz: NIT: this could be a bit slow especially if there is a lot of padding to add. Given the number of tokens added for our pad_token should be at least 1, could we encode once, padd to the target, and then return decode with tokens truncated to the desired length?

src/guidellm/preprocess/dataset.py
```diff
@@ -0,0 +1,224 @@
...
+    tokenizer: PreTrainedTokenizerBase,
+    **_kwargs,
+) -> Optional[str]:
+    if len(tokenizer.encode(current_prompt)) < min_prompt_tokens:
+        logger.warning("Prompt too short, ignoring")
+        return None
+    return current_prompt
+
+
+def handle_concatenate_strategy(
```
markurtz, line 51: The logic here might return tokens that are above the max the user set with the configs. Can we generalize the statistics and pass those into the handlers so the handlers can decide how best to fit the prompt to the desired range?

src/guidellm/preprocess/dataset.py
```diff
@@ -0,0 +1,224 @@
...
+    tokens_len = len(tokenizer.encode(current_prompt))
+    while tokens_len < min_prompt_tokens:
+        try:
+            next_row = next(dataset_iterator)
+        except StopIteration:
+            logger.warning(
+                "Could not concatenate enough prompts to reach minimum length, ignoring"
+            )
+            return None
+        current_prompt += next_row[prompt_column]
```
markurtz: We likely should add a space or new line with the concatenation so we aren't ending up with errored text / potentially enable the user how to specify the delimiter for combination here

src/guidellm/preprocess/dataset.py
```diff
@@ -0,0 +1,224 @@
...
+        tokens_len = len(tokenizer.encode(current_prompt))
+    return current_prompt
+
+
+def handle_pad_strategy(
```
markurtz, line 86: Similar issue to handle_concatenate_strategy where the logic here might create prompts that are outside of the max the user set since we are treating the sampled prompt tokens as the min rather than the target

src/guidellm/preprocess/dataset.py
```diff
@@ -0,0 +1,224 @@
...
+        )
+    )
+
+    dataset_iterator = iter(dataset)
+    processed_prompts = []
+    handler = STRATEGY_HANDLERS[short_prompt_strategy]
+
+    for prompt_row in dataset_iterator:
+        prompt_text = prompt_row[prompt_column]
+        target_prompt_len = next(prompt_token_sampler)
```
markurtz, line 316: Mentioned this above in the handlers, but I think we end up with issues here for sampling to create the target_prompt_len and treating it as the min since we could have lengths that exceed our maximum settings and additionally we might throw away or combine data that technically would have worked within the target ranges the user set

markurtz: This gets tricky, though, towards accepting a value within the stats range since that could shift our stats distribution

markurtz (changes requested):
Overall looks good. Added some comments towards potential issues that we should address as well as some NITs. 

For more general changes, can we add in doc statements for any public functions containing a quick writeup on the functionality and adding in :param, :return, and :raises? For anything that is an entrypoint, we should also add in a simple example of usage. And finally, for the tests, can you mark each test with a decorator of @pytest.mark.smoke, @pytest.mark.sanity, or @pytest.mark.regression based on the frequency they should run with?

src/guidellm/__main__.py
TomerG711: I udpated to help to be more general.

src/guidellm/__main__.py
TomerG711: Added parentheses to match code styling.

src/guidellm/__main__.py
TomerG711: I removed the `metavar`.

src/guidellm/__main__.py
TomerG711: I removed the `metavar`.

src/guidellm/__main__.py
TomerG711: `input` as already built-in name so I changed to `data`. I'm not quite sure what you mean, isn't it already a position arg?

src/guidellm/__main__.py
TomerG711: I modified to `--pad-char` to improve clarity.

src/guidellm/preprocess/dataset.py
TomerG711: Added a new `error_strategy_handler` + relevant UT.

src/guidellm/preprocess/dataset.py
TomerG711: Added new suffix type `jsonl`.

src/guidellm/preprocess/dataset.py
TomerG711: I might be mistaken here, but I think it does not work that way.
For example, let's say I have a prompt of 7 tokens and I want it to be 10 tokens. If I set my `pad_char` to `\n`, it does not mean we should add 3 times `\n`.  We might actually need more than that, depending of the tokenizer.
To validate this, I tried the following:
tokenizer: `Qwen2TokenizerFast` (tokenizer of `Qwen/Qwen2.5-0.5B`)
prompt: `Give three tips for staying healthy.` (initially equals to 7 tokens)
pad_char: `\n`
min_prompt_tokens: 10
Adding 3 times `\n` adds up to a single token using this tokenizer.
It actually needed to add `\n` 31 times to get total of 10 tokens (combined with the existing prompt).

Thus, we cannot assume how many padding tokens are needed beforehand, and we have to keep adding one by one and check each time.

src/guidellm/preprocess/dataset.py
TomerG711: Added lower case (so capitalized suffixes are supported now).

src/guidellm/__main__.py
TomerG711: I removed the duplicated seeds and changed to a single `random_seed` that is used for both prompt and output tokens.

src/guidellm/preprocess/dataset.py
TomerG711: It indeed might return tokens above the desired tokens count for the prompt. Note that for each of these strategies: `ignore`, `concatenate` and `pad`, the handlers only make sure we get to at least the desired token length. After the handler is done, we check if the prompt is too long now (back in `process_dataset`) and truncate it if needed. This happens after any of the handlers.

We could change it so that the handler itself will truncate if needed but it'll just happen the same in every handler. Do you think it's better?

src/guidellm/preprocess/dataset.py
TomerG711: Same comment as the for the concat handler.

src/guidellm/preprocess/dataset.py
TomerG711: I think that's the same as the pad/concat handler comments, right? It is being handled after the handler is done to have the prompt at the exact size we get from the sampler.

src/guidellm/preprocess/dataset.py
TomerG711: Introduced a new CLI argument: `concat_delimiter`

src/guidellm/__main__.py
TomerG711: I created a new Pydantic model: `TokensConfig` for this purpose, and both `prompt_tokens` and `output_tokens` utilize it now.

src/guidellm/__main__.py
markurtz: Ah yes, correct, I misread!

src/guidellm/preprocess/dataset.py
markurtz: Just noting here what we discussed in the weekly that we can increment faster than 1 to converge by doing an incrementing multiplier and then truncating. Another complimentary option is building an estimate of the characters per token, but more complicated

src/guidellm/preprocess/dataset.py
markurtz: Yep, correct

src/guidellm/preprocess/dataset.py
```diff
@@ -0,0 +1,398 @@
...
+
+STRATEGY_HANDLERS: dict[ShortPromptStrategy, Callable] = {
+    ShortPromptStrategy.IGNORE: handle_ignore_strategy,
+    ShortPromptStrategy.CONCATENATE: handle_concatenate_strategy,
+    ShortPromptStrategy.PAD: handle_pad_strategy,
+    ShortPromptStrategy.ERROR: handle_error_strategy,
+}
+
+
+class TokensConfig(BaseModel):
```
markurtz, line 148: NIT: Since we're using the functionality for parse_* in two classes now, I think it would make sense to add a general base object they extend from rather than duplicating code. That can be a follow up cleanup, though, since it's not critical

src/guidellm/preprocess/dataset.py
```diff
@@ -0,0 +1,398 @@
...
+
+    @staticmethod
+    def parse_config_file(data: Union[str, Path]) -> "TokensConfig":
+        with Path(data).open("r") as file:
+            config_dict = yaml.safe_load(file)
+
+        return TokensConfig(**config_dict)
+
+
+def save_dataset_to_file(dataset: Dataset, output_path: Union[str, Path]) -> None:
```
markurtz: NIT: this seems like a nice utility function to add to the HF utilities, but not a major thing

src/guidellm/preprocess/dataset.py
```diff
@@ -0,0 +1,398 @@
...
+            dataset_iterator=dataset_iterator,
+            prompt_column=prompt_column,
+            tokenizer=tokenizer,
+            pad_char=pad_char,
+            concat_delimiter=concat_delimiter,
+        )
+        if prompt_text is None:
+            continue
+
+        if len(tokenizer.encode(prompt_text)) > target_prompt_len:
```
markurtz:
```suggestion
        if (tokens := len(tokenizer.encode(prompt_text))) > target_prompt_len:
```

src/guidellm/preprocess/dataset.py
```diff
@@ -0,0 +1,398 @@
...
+            tokens = tokenizer.encode(prompt_text)
```
markurtz:
```suggestion
```

markurtz (dismissed):
@TomerG711 changes are looking good. I left two minor NITs that aren't blockers for me to land and added a code suggestion for simplification to avoid running tokenizer.encode twice. Also added a comment towards what we had talked about to get the tokenizer with padding to converge faster. That can also be a follow up fix if you'd like to land this.

src/guidellm/preprocess/dataset.py
TomerG711: I moved it to a new file: utils/hf_datasets.py. I updated the UT accordingly.

src/guidellm/preprocess/dataset.py
TomerG711: It does not work that way, as `tokens` is the list of tokens and not the length. However, I modified to get `tokens` before the `if` to prevent double encoding.

src/guidellm/preprocess/dataset.py
TomerG711: Removed it as discussed above

src/guidellm/preprocess/dataset.py
TomerG711: While I agree, I prefer to not do it in this PR, since the other parts of GuideLLM that use this model doesn't have UTs, and I will not have time to test that I didn't break anything.

src/guidellm/preprocess/dataset.py
TomerG711: I added a padding_multiplier. Again, remember that we always truncate after the handlers, so it's OK that it'll return longer prompt at this point.

src/guidellm/preprocess/dataset.py
markurtz: Ah sorry @TomerG711, I thought I had wrapped that in a len() with it. What you have in there works as well

PR #167, opened 2025-05-19
Implmement guidellm UI

DaltheCow:
This PR actually implements the application logic including the UI/charts and the api setup that retrieves data from the window object.

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -1,1 +1,37 @@
 # GuideLLM UI
+
+The GuideLLM UI is a companion to GuideLLM that allows you to visualize the peformance of your model in regards to a specific benchmark run.
+
+## Getting Started
+
+The two pathways to running the UI for your benchmark report are:
+
+1. Rely on the hosted build.
+
+If you choose this option, then you can just run your benchmarks and a report.html will be generated automatically. Assets will pull from the latest compatible version of GuideLLM UI. By default the report will be located
+
+2. Build locally.
+
+For this option:
+
+First, install dependencies:
+
+```bash
+npm install
+```
+
+Then
+
+```bash
+npm run build
+```
+
+This builds the app for production to the `out` folder.\
+
+Following this step you can serve the build however you like, for example:
+
+```bash
+npx serve out
+```
+
+Tell GuideLLM you are running things locally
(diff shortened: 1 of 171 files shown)


ui/README.md
```diff
@@ -1 +1,52 @@
-# GuideLLM UI
+### GuideLLM UI
+
+GuideLLM UI is a companion frontend for visualizing the results of a GuideLLM benchmark run.
+
+### 🛠 Running the UI
+
+1. Use the Hosted Build (Recommended for Most Users)
+
+After running a benchmark with GuideLLM, a report.html file will be generated (by default at guidellm_report/report.html). This file references the latest stable version of the UI hosted at:
```
DaltheCow, line 9: This is based off of a WIP effort i have. I don't know if this is where report.html will end up but that is what I'm working with for now.

ui/README.md
```diff
@@ -1 +1,52 @@
...
+
+- You want full control over how the report is displayed
+
+```bash
+npm install
+npm run build
+npx serve out
+```
+
+This will start a local server (e.g., at http://localhost:3000). Then, in your GuideLLM config or CLI flags, point to this local server as the asset base for report generation.
```
DaltheCow, line 32: This is not something I've set up yet. Currently in [this file](https://github.com/neuralmagic/guidellm/blob/set-up-data-for-ui/src/guidellm/config.py#L36) the asset bases are configured via environment. One for prod, staging, dev, and local. But what I have in the readme implies you can see it to something other than the preconfigured options. Maybe it'd be better to have local hardcoded to localhost:3000 and wait on adding an option for the user to configure. Not sure.

ui/README.md
```diff
@@ -1 +1,52 @@
...
+
+```
+src/lib/store/[runInfo/workloadDetails/benchmarks]WindowData.ts
+```
+
+In the future this will be replaced by a configurable untracked file for dev use.
+
+### 🚧 Future Possibilities
+
+We're evaluating options for hosting dev/staging/prod builds on GitHub Pages. For now, production builds will be published at:
```
DaltheCow, line 46: I think we should go with GitHub pages. I just left the other ideas in there as food for thought since gh pages is being used for docs right now, but I think these things can coexist.

PR #171, opened 2025-05-21
Feat/max error rate

markVaykhansky:
Addressing this issue:
https://github.com/neuralmagic/guidellm/issues/105

`--max-error-rate` flag was added. 
It is only applicable in either or the the following scenarios:
1) Fixed length dataset
2) `max_requests` is set
2) `rate_type` is `constant` and `max_duration` is set

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ The `guidellm benchmark` command is used to run benchmarks against a generative 
 
 - `--max-requests`: Sets the maximum number of requests for each benchmark run. If not provided, the benchmark will run until `--max-seconds` is reached or the dataset is exhausted.
 
-- `--max-error-rate`: The maximum error rate after which a benchmark will stop. Applicable only for finite deterministic scenarios i.e `rate_type` is `constant` and `--max-seconds` exists OR `--max-requests` exists OR the dataset is finite. If `--max-error-rate` is `None`, benchmarks will continue regardless of error rate.
+- `--max-error-rate`: The maximum error rate after which a benchmark will stop. Applicable only for finite deterministic scenarios i.e `rate_type` is `constant` and `--max-seconds` exists OR `--max-requests` exists OR the dataset is finite. If `--max-error-rate` is `None` or not applicable, benchmarks will continue regardless of error rate.
 
 - `--warmup-percent`: Specifies the percentage of the benchmark to treat as a warmup phase. Requests during this phase are excluded from the final results.
 
--- a/README.md
+++ b/README.md
@@ -147,6 +147,8 @@ The `guidellm benchmark` command is used to run benchmarks against a generative 
 
 - `--max-requests`: Sets the maximum number of requests for each benchmark run. If not provided, the benchmark will run until `--max-seconds` is reached or the dataset is exhausted.
 
+- `--max-error-rate`: The maximum error rate after which a benchmark will stop. Applicable only for finite deterministic scenarios i.e `rate_type` is `constant` and `--max-seconds` exists OR `--max-requests` exists OR the dataset is finite. If `--max-error-rate` is `None`, benchmarks will continue regardless of error rate.
+
 - `--warmup-percent`: Specifies the percentage of the benchmark to treat as a warmup phase. Requests during this phase are excluded from the final results.
 
 - `--cooldown-percent`: Specifies the percentage of the benchmark to treat as a cooldown phase. Requests during this phase are excluded from the final results.
--- a/src/guidellm/benchmark/aggregator.py
+++ b/src/guidellm/benchmark/aggregator.py
@@ -599,8 +599,8 @@ class GenerativeBenchmarkAggregator(
         and return the compiled object.
         """
         successful, incomplete, errored = self._compile_results()
-        error_rate = self.requests_stats.totals.errored.total / \
-            (self.requests_stats.totals.successful.total + self.requests_stats.totals.errored.total)
+
+        error_rate = self._calculate_error_rate()
 
         return GenerativeBenchmark.from_stats(
             run_id=self.run_id,
@@ -633,6 +633,12 @@ class GenerativeBenchmarkAggregator(
             requests_loader=self.request_loader_description,
             extras=self.extras,
         )
+
+    def _calculate_error_rate(self) -> float:
+        total_successful = self.requests_stats.totals.successful.total
+        total_errored = self.requests_stats.totals.errored.total
+        total_sent = total_errored + total_successful
+        return total_errored / total_sent
 
     def _compile_results(
         self,
--- a/src/guidellm/benchmark/aggregator.py
+++ b/src/guidellm/benchmark/aggregator.py
@@ -600,7 +600,7 @@ class GenerativeBenchmarkAggregator(
         """
         successful, incomplete, errored = self._compile_results()
         error_rate = self.requests_stats.totals.errored.total / \
-            (self.requests_stats.totals.successful + self.requests_stats.totals.errored.total)
+            (self.requests_stats.totals.successful.total + self.requests_stats.totals.errored.total)
 
         return GenerativeBenchmark.from_stats(
             run_id=self.run_id,
--- a/src/guidellm/benchmark/aggregator.py
+++ b/src/guidellm/benchmark/aggregator.py
@@ -599,6 +599,8 @@ class GenerativeBenchmarkAggregator(
         and return the compiled object.
         """
         successful, incomplete, errored = self._compile_results()
+        error_rate = self.requests_stats.totals.errored.total / \
+            (self.requests_stats.totals.successful + self.requests_stats.totals.errored.total)
 
         return GenerativeBenchmark.from_stats(
             run_id=self.run_id,
@@ -625,6 +627,7 @@ class GenerativeBenchmarkAggregator(
                 request_start_time_targeted_delay_avg=self.requests_stats.request_start_time_targeted_delay.mean,
                 request_time_delay_avg=self.requests_stats.request_time_delay.mean,
                 request_time_avg=self.requests_stats.request_time.mean,
+                error_rate=error_rate,
             ),
             worker=self.worker_description,
             requests_loader=self.request_loader_description,
--- a/src/guidellm/benchmark/benchmark.py
+++ b/src/guidellm/benchmark/benchmark.py
@@ -90,6 +90,9 @@ class BenchmarkArgs(StandardBaseModel):
     max_duration: Optional[float] = Field(
         description="The maximum duration in seconds to run this benchmark, if any."
     )
+    max_error_rate: Optional[float] = Field(
+        description="Maximum error rate after which a benchmark will stop."
+    )
     warmup_number: Optional[int] = Field(
         description=(
             "The number of requests to run for the warmup phase of this benchmark, "
@@ -211,6 +214,12 @@ class BenchmarkRunStats(StandardBaseModel):
             "The average time spent processing all requests in the benchmark run. "
             "This is the time from when the actual request was started to when "
             "it was completed."
+        )
+    )
+    error_rate: float = Field(
+        description=(
+            "The number of errored requests divided by the number of errored requests. This can be higher "
+            "than max_error_rate (if applicable) cause it does not take into account incomplete requests."
         )
     )
 
--- a/src/guidellm/benchmark/benchmark.py
+++ b/src/guidellm/benchmark/benchmark.py
@@ -218,8 +218,10 @@ class BenchmarkRunStats(StandardBaseModel):
     )
     error_rate: float = Field(
         description=(
-            "The number of errored requests divided by the number of errored requests. This can be higher "
-            "than max_error_rate (if applicable) cause it does not take into account incomplete requests."
+            "The number of errored requests divided by the number "
+            "of errored requests. This can be higher than max_error_rate "
+            "(if applicable) cause it does not take into "
+            "account incomplete requests."
         )
     )
 
--- a/src/guidellm/benchmark/benchmark.py
+++ b/src/guidellm/benchmark/benchmark.py
@@ -710,7 +710,7 @@ class GenerativeBenchmark(Benchmark):
             *["incomplete"] * len(incomplete),  # type: ignore[list-item]
             *["error"] * len(errored),  # type: ignore[list-item]
         ]
-        start_time = min(req.start_time for req in total) # ToDo: Fix if total is empty
+        start_time = min(req.start_time for req in total)
         end_time = max(req.end_time for req in total)
 
         total_with_prompt, total_types_with_prompt = (
--- a/src/guidellm/benchmark/benchmark.py
+++ b/src/guidellm/benchmark/benchmark.py
@@ -701,7 +701,7 @@ class GenerativeBenchmark(Benchmark):
             *["incomplete"] * len(incomplete),  # type: ignore[list-item]
             *["error"] * len(errored),  # type: ignore[list-item]
         ]
-        start_time = min(req.start_time for req in total)
+        start_time = min(req.start_time for req in total) # ToDo: Fix if total is empty
         end_time = max(req.end_time for req in total)
 
         total_with_prompt, total_types_with_prompt = (
--- a/src/guidellm/request/loader.py
+++ b/src/guidellm/request/loader.py
@@ -21,7 +21,12 @@ __all__ = [
     "GenerativeRequestLoaderDescription",
     "RequestLoader",
     "RequestLoaderDescription",
+    "InfiniteDatasetError"
 ]
+
+
+class InfiniteDatasetError(Exception):
+    pass
 
 
 class RequestLoaderDescription(StandardBaseModel):
@@ -120,7 +125,8 @@ class GenerativeRequestLoader(RequestLoader):
         if self.iter_type == "finite":
             return self.num_unique_items()
 
-        raise ValueError(f"Unable to determine length of dataset: {self.data}")
+        assert self.iter_type == "infinite"
+        raise InfiniteDatasetError(f"Dataset {self.data} is infinite and thus unable to determine length")
 
     @property
     def description(self) -> GenerativeRequestLoaderDescription:
--- a/src/guidellm/request/loader.py
+++ b/src/guidellm/request/loader.py
@@ -19,13 +19,13 @@ from guidellm.request.request import GenerationRequest
 __all__ = [
     "GenerativeRequestLoader",
     "GenerativeRequestLoaderDescription",
+    "GetInfiniteDatasetLengthError",
     "RequestLoader",
     "RequestLoaderDescription",
-    "InfiniteDatasetError"
 ]
 
 
-class InfiniteDatasetError(Exception):
+class GetInfiniteDatasetLengthError(Exception):
     pass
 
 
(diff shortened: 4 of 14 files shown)


src/guidellm/benchmark/aggregator.py
```diff
@@ -625,12 +627,19 @@ def compile(self) -> GenerativeBenchmark:
                 request_start_time_targeted_delay_avg=self.requests_stats.request_start_time_targeted_delay.mean,
                 request_time_delay_avg=self.requests_stats.request_time_delay.mean,
                 request_time_avg=self.requests_stats.request_time.mean,
+                error_rate=error_rate,
```
markurtz, line 630: It seems that implementing this as a calculated field under the StatusBreakdown would make sense, but that would require some work in that class to start generating computed values based on the types (numerics, collections, and generics). Putting this note here in case you're interested in generalizing the solution a bit more, but not a blocker from my side

src/guidellm/benchmark/benchmark.py
```diff
@@ -213,6 +216,14 @@ class BenchmarkRunStats(StandardBaseModel):
             "it was completed."
         )
     )
+    error_rate: float = Field(
+        description=(
+            "The number of errored requests divided by the number "
+            "of errored requests. This can be higher than max_error_rate "
```
markurtz:
```suggestion
            "of successful and errored requests. This can be higher than max_error_rate "
```

src/guidellm/benchmark/benchmark.py
```diff
@@ -213,6 +216,14 @@ class BenchmarkRunStats(StandardBaseModel):
...
+            "(if applicable) cause it does not take into "
+            "account incomplete requests."
```
markurtz, line 225: I'm not following this point, the error rate we calculate and compare to should never include incomplete, right?

src/guidellm/scheduler/scheduler.py
```diff
@@ -159,6 +166,17 @@ async def run(
                         run_info,
                     )
                     if iter_result is not None:
+                        if iter_result.request_info.errored \
```
markurtz: NIT: should use paranetheses rather than line escapes for if statements across multiple lines, tox -e style should fix

markurtz (changes requested):
I left a few minor comments for the less intrusive changes. Also, please be sure to run `tox -e style` and `tox -e types` locally and check those are passing -- there are a bunch of styling pieces, such as where parentheses are placed, how new lines are escaped, etc that are inconsistent with the ruff styling setup. Running tox -e style locally applied several fixes and raised some errors that couldn't be resolved without either modifying the code or ignoring them where appropriate.

For implementation, I think defaulting always to use the shutdown event is a good pattern rather than relying on two signals: either `None` being placed on the queue or the shutdown event being set if the maximum error rate is reached. I'd like to incorporate that into this to reduce complexity. 

Additionally, I think we should generally combine these into a single parameter, max_errors, which can be set as either max_error_rate or an absolute number of errors that can occur, allowing us to support all benchmarking pathways.

Additionally, with the above, as well as some combination logic in asyncio, we should be able to simplify the changes, so we don't need to make significant changes to the resolve requests pathways and can cancel for any type. For example, we should be able to replace the root process runner to call into async loops and handle the shutdown event globally in a central place with something like the following:
```python
    def run_process(
        self,
        type_: Literal["synchronous", "asynchronous"],
        requests_queue: multiprocessing.Queue,
        results_queue: multiprocessing.Queue,
        shutdown_event: multiprocessing.Event,
        shutdown_poll_interval: float,
        process_id: int,
        max_concurrency: int,
    ):
        async def _process_runner():
            if type_ == "synchronous":
                loop_task = asyncio.create_task(self.synchronous_loop(...))
            elif type_ == "asynchronous":
                loop_task = asyncio.create_task(self.asyncronous_loop(...))
            else:
                raise ValueError(f"Invalid process type: {type_}")
            
            shutdown_task = asyncio.create_task(
                self.wait_for_shutdown(shutdown_event, shutdown_poll_interval)
            )
            
            done, pending = await asyncio.wait(
                [
                    loop_task,
                    shutdown_task,
                ],
                return_when=asyncio.FIRST_EXCEPTION,
            )

            for task in pending:
                task.cancel()
                try:
                    await task
                except asyncio.CancelledError:
                    pass

            for task in done:
                if task.exception():
                    raise task.exception()  # need to check and ignore cancelations
        
        try:
            asyncio.run(_process_runner())
        except Exception as exc:  # noqa: BLE001
            logger.error(
                f"Error in worker process {process_id}: {exc}",
                exc_info=True,
                stack_info=True,
            )
        finally:
            shutdown_event.set()  # ensure shutdown event is set to stop other processes

    async def wait_for_shutdown(
        self, shutdown_event: multiprocessing.Event, shutdown_poll_interval: float
    ):
        while not shutdown_event.is_set():
            await asyncio.sleep(shutdown_poll_interval)

        raise asyncio.CancelledError("Shutdown event set, cancelling process loop.")
    
    async def synchronous_loop(self, ...):
        while True:
            process_request = await self.get_request(requests_queue)
            ...

    async def asyncronous_loop(self, ...):
        while True:
            process_request = await self.get_request(requests_queue)
            ...
```

src/guidellm/benchmark/benchmark.py
markVaykhansky: Correct

src/guidellm/scheduler/scheduler.py
markVaykhansky: Done

markVaykhansky:
I've changed the logic as suggested to have a single termination method which is a `shutdown event`. 
However, this did not go a smoothly as expected:

After implementing the code like you suggested above I ran into an issue with `get_request`. After a shutdown was initiated and the `_wait_for_shutdown` task raised an exception the code moved to cancel the pending task i.e the request sending task. However, it was stuck indefinitely cause the logic there's `asyncio.to_thread(requests_queue.get)` which is blocking. To resolve the issue I've re-introduced the `sleep_intermittently` logic which check the `shutdown_event` periodically inside the `get_request`. Then I ran into another issue, after sending the shutdown event there's a race condition between the `asyncio.wait` and checking for errors in `get_request`, to resolve the issue I've created a separate `internal_shutdown_event` which is set after the `task.cancel()`. 

Another super weird thing that happened is the raising specifically `asyncio.CancellationError` in `_wait _for_shutdown` did not trigger the return of `asyncio.wait` with `FIRST_EXCEPTION` and it was stuck forever, I'm not sure why couldn't find a reasonable explanation. Ended up throwing a custom error instead. 

Also, I added the context of last X request as suggested but I feel that for the scenario where we do know the exact amount, i.e const rate + fixed duration, it is best to look at the entire run.

.pre-commit-config.yaml
```diff
@@ -4,40 +4,40 @@ repos:
   hooks:
   - id: trailing-whitespace
   - id: end-of-file-fixer
-- repo: https://github.com/astral-sh/ruff-pre-commit
```
markurtz, line 7: We'll need these reenabled. If you ever need to push and skip recommit, you can pass in the --no-verify flag with your git commit command

README.md
```diff
@@ -147,6 +147,8 @@ The `guidellm benchmark` command is used to run benchmarks against a generative
 
 - `--max-requests`: Sets the maximum number of requests for each benchmark run. If not provided, the benchmark will run until `--max-seconds` is reached or the dataset is exhausted.
 
+- `--max-error-rate`: The maximum error rate after which a benchmark will stop. Applicable only for finite deterministic scenarios i.e `rate_type` is `constant` and `--max-seconds` exists OR `--max-requests` exists OR the dataset is finite. If `--max-error-rate` is `None` or not applicable, benchmarks will continue regardless of error rate.
```
markurtz, line 150:
```suggestion
- `--max-error`: The maximum error rate after which a benchmark will stop. Can either be a rate i.e 0 < rate < 1 or constant number. If rate is given and rate_type is 'constant' and 'max_seconds' exists then the rate will be calculated as part of the total expected requests counts i.e rate * duration. If rate is given and number of requests is not pre-determined than a context window of the last requests will be looked at. Context window size is configurable under GUIDELLM__ERROR_CHECK_WINDOW_SIZE. If a number above 1 is given than we just count the total number of error and check if it's above the threshold.
```

src/guidellm/scheduler/result.py
```diff
@@ -46,12 +50,15 @@ class SchedulerRunInfo(StandardBaseModel):
     end_number: float
     processes: int
     strategy: SchedulingStrategy
+    last_requests_statuses: deque[RequestStatus]
```
markurtz, line 53: I think we can simplify the logic here so we don't need to add this in especially since this list can be come very long and it is not json serializable so would break any pydantic serialization in the event someone wants to save that state. See note later down where that logic is located

src/guidellm/request/loader.py
```diff
@@ -19,11 +19,16 @@
 __all__ = [
     "GenerativeRequestLoader",
     "GenerativeRequestLoaderDescription",
+    "GetInfiniteDatasetLengthError",
     "RequestLoader",
     "RequestLoaderDescription",
 ]
 
 
+class GetInfiniteDatasetLengthError(Exception):
```
markurtz, line 28: I think this exception and logic is no longer needed since if we can get the length then we go through a specific logic route and if we can't, then we fallback on the window

src/guidellm/scheduler/scheduler.py
```diff
@@ -98,20 +105,17 @@ async def run(
...
-        ):
-            raise ValueError(f"Invalid scheduling strategy: {scheduling_strategy}")
-
-        if max_number is not None and max_number < 1:
-            raise ValueError(f"Invalid max_number: {max_number}")
-
-        if max_duration is not None and max_duration < 0:
-            raise ValueError(f"Invalid max_duration: {max_duration}")
+        self._validate_scheduler_params(
+            scheduling_strategy, max_duration, max_error, max_number
```
markurtz, line 117: NIT: could we keep the arg order the same as the surrounding function? Makes it a bit easier to track especially if additions com in later

src/guidellm/scheduler/scheduler.py
```diff
@@ -159,6 +170,18 @@ async def run(
                         run_info,
                     )
                     if iter_result is not None:
+                        if (
+                            iter_result.request_info.errored
```
markurtz, line 174: NIT: could we nest these checks into the is_max_error_reached function? That way the main function is a little less populated and the max error function is responsible for determining how a result should be counted

src/guidellm/scheduler/scheduler.py
```diff
@@ -120,19 +124,26 @@ async def run(
...
             )
             yield SchedulerResult(
                 type_="run_start",
                 run_info=run_info,
             )
 
             try:
-                while True:
+                max_error_reached = False
+                while not max_error_reached:
```
markurtz, line 146: Since we're replacing this with stopping loop logic, can we add in the checks for if all requests have been run here as well so we centralize where the exit from the loop is located?

src/guidellm/scheduler/scheduler.py
```diff
@@ -171,7 +194,62 @@ async def run(
...
+            )
+            return max_error < run_info.errored_requests
+        elif run_info.strategy.type_ == "constant" and run_info.end_number != math.inf:
+            current_error_ratio = run_info.errored_requests / run_info.end_number
+            logger.debug(
+                f"Current error rate {current_error_ratio} "
+                f"i.e total_finished [success / error] / max total possible"
+            )
+            return max_error < current_error_ratio
+        elif settings.error_check_window_size <= run_info.completed_requests:
```
markurtz, line 237: I think we could simplify this even further by combining it with the previous if statement, treating it as a growing window size from the start of requests. Therefore, we can have both pathways continue until they reach an error rate that exceeds a specific number, and thus achieve a certain precision. Or, we can do this just for the non constant requests -- I think it might be nice to unify logic rather than having a branching pathway for constant, but not something I'm strongly opposed to.

The precision would be determined from the rate the user passed in. For example, if they passed in a decimal to the tenths place (0.1, 0.4, etc), then we need at least 10 requests to come in to be able to check that precision. For a decimal to the hundredths place (0.01, 0.25, etc), then we need at least 100 requests to come in to be able to check that precision, and so on.

The other benefit other than the simplified logic is that we don't need to keep state around to check within the window (sorry I didn't think of this before).

src/guidellm/scheduler/scheduler.py
```diff
@@ -276,10 +344,40 @@ def _run_setup(
...
+        max_duration: Optional[float],
+        max_number: Optional[int],
+    ) -> Union[int, float]:
+        end_number = max_number or math.inf
+        try:
+            # update end_number if the request_loader is finite and less than max_number
+            iter_length = len(self.request_loader)  # type: ignore[arg-type]
+            if 0 < iter_length < end_number:
+                end_number = iter_length
+        except GetInfiniteDatasetLengthError:
```
markurtz, line 367: I think this should just be the base case where we catch the Exception and set it to the indefinite pathway so we can simplify the logic a bit more

src/guidellm/scheduler/scheduler.py
```diff
@@ -374,9 +481,9 @@ def _check_result_ready(
     async def _stop_processes(
         self,
         futures: list[asyncio.Future],
-        requests_queue: multiprocessing.Queue,
+        shutdown_event: MultiprocessingEvent,
     ):
-        for _ in futures:
-            requests_queue.put(None)
-
+        if not shutdown_event.is_set():
```
markurtz, line 486: Is there a case where this would get set before we reach this stop_processes logic? Mainly checking if this is an error state for the logic and we raise an exception rather than silently ignoring it

.pre-commit-config.yaml
markurtz: Also @sjmonson just landed a diff that might clear up some of the issues you were seeing

src/guidellm/benchmark/benchmark.py
markurtz: Isn't max_error also not looking at incomplete requests, though? How would it be higher for error_rate vs max_error since those should be based off the same calculations, right?

src/guidellm/scheduler/scheduler.py
```diff
@@ -241,28 +314,23 @@ async def _start_processes(
...
         processes: list[asyncio.Future],
         scheduling_strategy: SchedulingStrategy,
         max_number: Optional[int],
         max_duration: Optional[float],
+        max_error: Optional[float],
     ) -> tuple[SchedulerRunInfo, Iterator[Any], Iterator[float]]:
         requests_iter = iter(self.request_loader)
         start_time = time.time()
         times_iter = iter(scheduling_strategy.request_times())
         end_time = time.time() + (max_duration or math.inf)
```
markurtz, line 330: NIT: could we move end_time into the dedicated function now for determining the end stopping criteria / targets alongside end_number? Alternatively, if we've simplified the logic sufficiently through the new assumptions based on end_number and error checking within a growing window, then we could remove these changes entirely.

src/guidellm/scheduler/scheduler.py
```diff
@@ -362,6 +460,15 @@ def _check_result_ready(
             run_info.processing_requests -= 1
             run_info.completed_requests += 1
 
+            is_errored = process_response.info.errored
+            if is_errored:
```
markurtz, line 464: NIT: check the property directly here rather than creating another local namespace variable

src/guidellm/scheduler/worker.py
```diff
@@ -121,9 +129,27 @@ async def resolve(
...
+        # We need to check shutdown_event intermittently cause
+        # if we simply use asyncio.to_thread(requests_queue.get)
+        # the cancellation task doesn't propagate because the
+        # asyncio.to_thread is blocking
+        def _get_queue_intermittently():
+            while True:
+                try:
+                    return requests_queue.get(timeout=shutdown_poll_interval_seconds)
+                except queue.Empty as e:
+                    logger.info("Checking shutdown even is set in get_request")
```
markurtz, line 147: This is going to log out a ton to info, we should remove or make debug / trace

src/guidellm/scheduler/worker.py
```diff
@@ -121,9 +129,27 @@ async def resolve(
...
+                    if shutdown_event.is_set():
```
markurtz, line 148: I think it would be cleaner here if we used a local namespace event that this function handles rather than relying on the global signal. That event would be set by a canceled error being raised in the local context so this function is properly cancellable at any point rather than the global shutdown event for future use / proper compatibility with asyncio

src/guidellm/scheduler/worker.py
```diff
@@ -149,25 +175,29 @@ async def resolve_scheduler_request(
...
+        request_scheduled_result: WorkerProcessResult[RequestT, ResponseT] = (
+            WorkerProcessResult(
+                type_="request_scheduled",
+                request=request,
+                response=None,
+                info=info,
+            )
         )
-        asyncio.create_task(self.send_result(results_queue, result))
+        asyncio.create_task(self.send_result(results_queue, request_scheduled_result))
```
markurtz, line 186: good catch on potential threading issues with these. Rather than polluting the local namespace with a bunch of vars that aren't used later, though, could we add the WorkerProcessResult to pass directly to this call function?

src/guidellm/scheduler/worker.py
```diff
@@ -185,27 +215,87 @@ async def resolve_scheduler_request(
...
+        shutdown_poll_interval_seconds: float,
         process_id: int,
+        max_concurrency: Optional[int] = None,
     ):
         async def _process_runner():
-            while (
-                process_request := await self.get_request(requests_queue)
-            ) is not None:
-                dequeued_time = time.time()
+            # We are using a separate internal event
```
markurtz, line 229: See my note on fixes for get_request so we don't hit this race condition ideally and can rely on the shutdown event here

src/guidellm/scheduler/worker.py
```diff
@@ -185,27 +215,87 @@ async def resolve_scheduler_request(
...
-    def process_loop_synchronous(
+    def run_process(
         self,
+        type_: Literal["sync", "async"],
         requests_queue: multiprocessing.Queue,
         results_queue: multiprocessing.Queue,
+        shutdown_event: MultiprocessingEvent,
+        shutdown_poll_interval_seconds: float,
         process_id: int,
+        max_concurrency: Optional[int] = None,
```
markurtz, line 226: Can we change this out so it isn't optional? I think we had a change either pushed in or queued for that and that way we don't need to validate it for async. I believe the changes are in so that for sync, it still returns a max_concurrency there as well

src/guidellm/scheduler/worker.py
```diff
@@ -185,27 +215,87 @@ async def resolve_scheduler_request(
...
+
+            for task in pending:
+                logger.debug(
+                    f"Cancelling task {task.get_name()}|| Process {process_id}"
                 )
+                task.cancel()
+                internal_shutdown_event.set()
+                try:  # noqa: SIM105
+                    await task
+                except asyncio.CancelledError:
```
markurtz, line 292: Since we are raising a different error in the shutdown task, there could be a race condition here where an error occurs and shutdown is set before we call await causing an uncaught exception here

src/guidellm/scheduler/worker.py
```diff
@@ -215,53 +305,107 @@ async def _process_runner():
...
+    async def _wait_for_shutdown(
+        self,
+        shutdown_event: MultiprocessingEvent,
+        shutdown_poll_interval: float,
+        process_id: int,
+    ):
+        while not shutdown_event.is_set():  # noqa: ASYNC110
+            await asyncio.sleep(shutdown_poll_interval)
+
+        # Raising asyncio.CancelledError instead would
```
markurtz, line 320: I think this is the correct way where we raise a specific error signaling this which we can track and catch. Can we remove the comments here since this is pretty self explanatory based on the name of the exception?

PR #172, opened 2025-05-21
fix links check fail, remove speculators from docs

DaltheCow:
We are getting link-check fails on the github rich display links. Not sure what the desired solution is but this gets the check passing by ignoring links like that.

Also we have what look to be some pasted doc remnants from the speculators repo so I changed everything saying speculators to guidellm

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -37,22 +37,6 @@ jobs:
       - name: Run quality checks
         run: tox -e types
 
-  link-checks:
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python: ["3.9"]
-    steps:
-      - uses: actions/checkout@v4
-      - name: Set up Python
-        uses: actions/setup-python@v5
-        with:
-          python-version: ${{ matrix.python }}
-      - name: Install dependencies
-        run: pip install tox
-      - name: Run link checks
-        run: tox -e links
-
   precommit-checks:
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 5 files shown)


sjmonson (approved):

markurtz (changes requested):
Thanks @DaltheCow! The big issue that is happening behind the scenes is that we're getting rate-limited by GitHub on checks, and the link checking software doesn't have any ability for us to apply a rate limit to it (in combination with GitHub not returning proper wait until headers). So I was going to migrate to a new tool that we could control, which would also pass in GitHub auth credentials to increase the potential rate at which we can hit GitHub. For now, though, the updates to the links are good catches, and if we can update it to just wipe out the link checks until we can get the new tool in, that would be perfect. I'd rather wipe out the tool than hardcode skipping of all GitHub links, so we don't have a false sense of security reported, since most of the links will be GitHub ones. I can update a bit later, but if you can get to it, that would be perfect and I can rereview

DaltheCow:
> Thanks @DaltheCow! The big issue that is happening behind the scenes is that we're getting rate-limited by GitHub on checks, and the link checking software doesn't have any ability for us to apply a rate limit to it (in combination with GitHub not returning proper wait until headers). So I was going to migrate to a new tool that we could control, which would also pass in GitHub auth credentials to increase the potential rate at which we can hit GitHub. For now, though, the updates to the links are good catches, and if we can update it to just wipe out the link checks until we can get the new tool in, that would be perfect. I'd rather wipe out the tool than hardcode skipping of all GitHub links, so we don't have a false sense of security reported, since most of the links will be GitHub ones. I can update a bit later, but if you can get to it, that would be perfect and I can rereview

Sounds good! Will do that

sjmonson (approved):

PR #174, opened 2025-05-29
Implement UI build pipeline

DaltheCow:
Builds would live at:
pr builds - neuralmagic.github.io/guidellm/ui/pr/12
dev - neuralmagic.github.io/guidellm/ui/dev
staging/release-candidate - neuralmagic.github.io/guidellm/ui/release/v1.0.0 or .../ui/release/latest
production/release - neuralmagic.github.io/guidellm/ui/v1.0.0 or .../ui/latest

PR builds will include mocked data, view example in comments below. All other builds currently will be blank slates, waiting to be pulled down and used to generate an html report.


DaltheCow:
> > 🎉 **Live Preview:** [Click here to view the live version](https://neuralmagic.github.io/guidellm/ui/pr/174)
> 
> Question: Should these live previews be limited to UI related PRs? We can just check if any UI related files were changed and only run in that case. There is no reason in my mind to run these on non-UI PRs since currently there is no plan for the UI PR builds to interact directly with the python side. Mock data is used to populate the app. I think think we will want these comments in every PR. I'm feeling pretty strongly about this, I can't see a reason to leave them in for non-UI PRs, so I'm going to make the change, but happy to revert if anyone has a reason for that.

The solution I have for this is not the cleanest, it lists out all of the files in the root that are related to the UI that could trigger a UI build, as well as checking src/ui and tests/ui.

Not sure if I should move forward with this as is. I can also create a file that the workflow reads from to find where to look for changes so that it doesn't live directly in the workflow.

I don't think it is great to have every commit build the UI especially when most of the commits in PRs in this repo will not be UI related. But regardless, I've updated the commenting logic so that when the actions posts a comment in the PR displaying the link to the UI preview, it will just keep updating that same comment instead of adding new ones. So it won't pollute things as much regardless.

DaltheCow:
I should add the UI jobs to the nightly workflow as well, I left that one out as an afterthought

DaltheCow:
I have confirmed that this precommit-check is bugging out, something staled out or is wrong, but looking at the branch itself instead of this PR, the workflow passes. I have spent a long time trying to fix this, I think the only way to fix this would be to recreate this PR (and in turn the next PR) from scratch off of the base branch. Hoping I don't have to do that but I think this branch is in a bad situation even though the check should be passing.

DaltheCow:
> I have confirmed that this precommit-check is bugging out, something staled out or is wrong, but looking at the branch itself instead of this PR, the workflow passes. I have spent a long time trying to fix this, I think the only way to fix this would be to recreate this PR (and in turn the next PR) from scratch off of the base branch. Hoping I don't have to do that but I think this branch is in a bad situation even though the check should be passing.

Oh thank god I figured it out, absurd hidden weird git issue. I updated .env.local in the base branch and failed to add the trailing line at the end, but I guess when it ran the checks it didn't fail in that PR. And then in this branch I fixed the file and added the trailing line but the check was failing. But fixing the issue in the base branch and merging in fixed things in this branch. Maybe just a weirdness with the pull_request_target trigger vs using pull_request or maybe just not basing a branch off of main. I'm not sure why it might choose to look at values from the old branch over the one that we are looking to merge in. The last PR should have been failing. Anyways. Everything is ready for review.

PR #185, opened 2025-06-13
Option to re-display a benchmark file

### LINKED ISSUE #175: Re-load and re-display a saved benchmark.json report
**Description:**
Enable the ability to load up and display a previous benchmark.json report in the terminal. 

**Acceptance Criteria:**
- Enable a `--view-report benchmark.json` command to surface an existing benchmark.json file result in the CLI (and in the future the UI)

jaredoconnell:
closes #175 

This adds a command to re-display a prior benchmarks file in the CLI.

Before marking this as ready for review, we need to decide what command format we want to use. During the call with Mark we discussed this being an option within the benchmark command.
Also, let me know if the stripped down results file is a good one to use. I manually removed data from a large results file.

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -224,3 +224,28 @@ jobs:
               They will be retained for **up to 30 days**.
               `
             })
+
+  build-and-push-container:
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Buildah build
+        id: build-image
+        uses: redhat-actions/buildah-build@v2
+        with:
+          image: ${{ github.event.repository.name }}
+          tags: "pr-${{ github.event.number }}"
+          containerfiles: |
+            ./deploy/Containerfile
+      - name: Push To ghcr.io
+        id: push-to-ghcr
+        uses: redhat-actions/push-to-registry@v2
+        with:
+          image: ${{ steps.build-image.outputs.image }}
+          tags: ${{ steps.build-image.outputs.tags }}
+          username: ${{ github.actor }}
+          password: ${{ github.token }}
+          registry: ghcr.io/${{ github.repository_owner }}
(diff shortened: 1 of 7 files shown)


src/guidellm/__main__.py
```diff
@@ -282,6 +286,16 @@ def benchmark(
     )
 
 
+@cli.command(help="Redisplay a saved benchmark report.")
+@click.argument(
+    "path",
+    type=click.Path(),
```
sjmonson:
```suggestion
    type=click.Path(file_okay=True, dir_okay=False, exists=True),
```

src/guidellm/benchmark/entrypoints.py
```diff
@@ -139,3 +134,12 @@ async def benchmark_generative_text(
     console.print_line("\nBenchmarking complete.")
 
     return report, saved_path
+
+def display_benchmarks_report(file: Path):
+    console = GenerativeBenchmarksConsole(enabled=True)
+    if not file.exists():
+        console.print_line(f"File {file} not found.")
+        return
```
sjmonson: Drop this, you can use click arg suggestion to handle input checking.

src/guidellm/__main__.py
```diff
@@ -282,6 +286,16 @@ def benchmark(
...
+    default=Path.cwd() / "benchmarks.json",
+)
+def display(path):
```
sjmonson: `display` seems fine to me, but @markurtz may have a better keyword in mind.

sjmonson (changes requested):

markurtz (changes requested):
Thanks, @jaredoconnell; overall, the code looks good. I want to expand the functionality to encompass not only displaying the results but also reexporting them to a new file if the user desires.

Given that, I'd recommend something along the lines of the following for the CLI:
`guidellm benchmark report PATH --output OUTPUT` where output is optional. If an output file is supplied, we will resave the report to that file path using the extension as the file type. It could also potentially be named `export`, `convert`, or anything along those lines.

For the benchmark CLI pathways, it would then look like the following:
`guidellm benchmark run ...`
`guidellm benchmark report ...`

And if that ACTION after the benchmark is not supplied, we will default to filling it in as run. This way, we namespace all of the commands under benchmark and add flexibility towards the future.

jaredoconnell:
The CI errors appears to be from using an older commit's code. That's very odd.

Issue #147, opened 2025-05-05
Event loop is closed error.


phalani-paladugu:
I’m encountering the following error for some requests when running the GuideLLM benchmark command using version v0.2.1 with Python 3.12.9.

The command I’m using is:

```
guidellm benchmark \
  --target "<valid target>" \
  --model "<model>" \
  --rate-type constant \
  --rate 0.31 \
  --data "prompt_tokens=4096,output_tokens=4096" \
  --processor "<valid tokenizer>" \
  --max-seconds 300 \
  --max-requests 100 \
  --output-path test.csv
```

And the error I’m seeing intermittently is:

` text_completions | ERROR - OpenAIHTTPBackend request with headers: {'Content-Type': 'application/json', 'Authorization': 'Bearer <valid token>'} and payload: {'prompt': '; <valid prompt>', 'model': '<valide model>', 'stream': True, 'stream_options': {'include_usage': True}, 'max_tokens': 128, 'max_completion_tokens': 128, 'stop': None, 'ignore_eos': True} failed: Event loop is closed`

Has anyone else encountered this issue? If so, was there a resolution or workaround that worked for you?

brianlkurtz:
I saw the same error with the configuration below on multiple endpoints, on being a L40s with vllm 0.8.2 that has been reliable for us, so I suspect its on the testing side.
guidellm benchmark \
     --target "" \
     --model "" \
     --processor"" \
       --rate-type sweep \
        --max-seconds 30 \
        --max-requests 128 \
        --data "prompt_tokens=512,output_tokens=512"

PR #188, opened 2025-06-13
Fix "Event loop Closed" error

sjmonson:
By default httpx will keep the last few connections open in a pool for re-use. Since we create the httpx connection inside an asyncio task, it is bound to the lifetime of the task rather than the httpx client. Thus when the client attempts to reuse the connection in a new task it fails. There may be performance implications for this change, but since the default is only to re-use the last 5 connections, its likely minor. ([ref](https://github.com/encode/httpx/discussions/2959))

Fixes: #147

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -1,7 +1,7 @@
 name: Development
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, synchronize, reopened]
 
 jobs:


ashishkamra:
Some other suggestions including Mark's:

1) Keep reuse, close pools on shutdown

Call await self._async_client.aclose() inside a __del__ or explicit shutdown() hook 
(httpx docs recommend this)
[python-httpx.org](https://www.python-httpx.org/async/)
[github.com](https://github.com/encode/httpx/discussions/1552)

2) Reuse a singleton client per event-loop

Instantiate the client in the backend constructor or in a FastAPI lifespan hook and share it across tasks; close it when the loop ends
[medium.com](https://medium.com/%40benshearlaw/how-to-use-httpx-request-client-with-fastapi-16255a9984a4)
[community.temporal.io](https://community.temporal.io/t/reusing-httpx-asyncclient-in-workwflow/14220)

3) Make keep-alive configurable

Expose --httpx-max-keepalive CLI flag or GUIDELLM_HTTPX_KEEPALIVE env-var; default to 20, let users drop to 0 if they see the bug.

4) Hybrid limits

httpx.Limits(max_keepalive_connections=1, keepalive_expiry=1) keeps a single hot connection per host but is immune to cross-loop reuse in most cases.

markurtz (approved):

PR #191, opened 2025-06-18
Save as html

DaltheCow:


src/guidellm/objects/statistics.py
```diff
@@ -37,6 +37,9 @@ class Percentiles(StandardBaseModel):
     p25: float = Field(
         description="The 25th percentile of the distribution.",
     )
+    p50: float = Field(
+        description="The 50th percentile of the distribution.",
+    )
```
DaltheCow, line 42: I think this was excluded because it represents the median so it is redundant, simplifies generating the UI side to add it in though so I added it back

src/ui/serve.json
```diff
@@ -0,0 +1,3 @@
+{
+  "cleanUrls": false
+}
```
DaltheCow, line 3: Impacts local serving redirects. Enables injector to be able to target localhost:3000/index.html without being redirected away

PR #200, opened 2025-06-25
Lock click to older version

sjmonson:
The latest version of click has dropped support for python 3.9 and has additionally introduced breaking interface changes. Until we drop python 3.9, lock to the last supported click version. Additionally types-click has been deprecated since click 8.0.0 so remove it.

### DIFF
--- a/.github/workflows/development-cleanup.yml
+++ b/.github/workflows/development-cleanup.yml
@@ -1,0 +1,87 @@
+name: Cleanup Dev
+
+on:
+  pull_request:
+    types: [closed]
+
+jobs:
+  cleanup-ui-pr-preview:
+    permissions:
+      contents: write
+      id-token: 'write'
+      issues: write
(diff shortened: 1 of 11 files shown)


jaredoconnell (dismissed):
Looks good, but we need to get the CI to pass. Is the error present in main, or is the version change causing a new error, is is it exposing one that wasn't there before?

markurtz (changes requested):
@sjmonson can you take a look at these again with the build jobs failing for pre-commit and type checks? I believe it's due to removing the types-click from pre commit config and types-click from the toml. Can we pin those to the latest version before 8.0 rather than removing?

sjmonson:
@markurtz Fixed linting failures; `types-click` was deprecated in click 8.0 (last update Nov 2021) so it was masking some incompatibilities.

markurtz (approved):

PR #226, opened 2025-07-10
Fix logging documentation

jaredoconnell:
The logging documentation is buried in a code file, and it's wrong, so this PR corrects that.


sjmonson (approved):
LGTM

PR #230, opened 2025-07-15
Add CLI options for backend args (like headers and verify)

kdelee:
This PR adds the ability to configure custom request headers and control SSL certificate verification when running benchmarks.

   * The OpenAIHTTPBackend now supports passing custom headers and a verify flag to disable SSL verification.
   * Headers are now merged with the following precedence: CLI arguments (--backend-args), scenario file arguments, environment variables, and then default values.
   * Headers can be removed by setting their value to null in the --backend-args JSON string.
   * The --backend-args help text has been updated with an example of how to use these new features.
   * New documentation has been added for the CLI, configuration options, and supported data formats.
   * Unit tests have been added to verify the new header and SSL verification logic, as well as the CLI argument parsing.

  This provides a way to benchmark targets that require custom authentication, other headers, or use self-signed SSL certificates.

### DIFF
--- a/src/guidellm/backend/openai.py
+++ b/src/guidellm/backend/openai.py
@@ -95,6 +95,8 @@ class OpenAIHTTPBackend(Backend):
         extra_query: Optional[dict] = None,
         extra_body: Optional[dict] = None,
         remove_from_body: Optional[list[str]] = None,
+        headers: Optional[dict] = None,
+        verify: Optional[bool] = None,
     ):
         super().__init__(type_="openai_http")
         self._target = target or settings.openai.base_url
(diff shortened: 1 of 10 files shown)


kdelee:
@sjmonson pushed some updates, looks like workflow needs approval

src/guidellm/backend/openai.py
```diff
@@ -94,6 +94,7 @@ def __init__(
         extra_query: Optional[dict] = None,
         extra_body: Optional[dict] = None,
         remove_from_body: Optional[list[str]] = None,
+        **kwargs,
```
sjmonson: Follow same pattern as other optionals

```suggestion
        headers: Optional[dict] = None,
        verify: Optional[bool] = None,
```

src/guidellm/backend/openai.py
```diff
@@ -110,20 +111,36 @@ def __init__(
...
+            default_headers["OpenAI-Organization"] = self.organization
+
         self.project = project or settings.openai.project
+        if self.project:
+            default_headers["OpenAI-Project"] = self.project
+
+        # User-provided headers from kwargs or settings override defaults
+        user_headers = kwargs.pop("headers", settings.openai.headers or {})
+        default_headers.update(user_headers)
+        self.headers = default_headers
```
sjmonson: It would be nice to layer this per-key and support removing headers. Something like this:

```suggestion
        default_headers.update(settings.openai.headers or {})
        default_headers.update(headers)
        self.headers = {k: v for k, v in default_headers.items() if v is not None}
```

This would also allow you to simplify the `default_headers` construction to something like:

```python
default_headers = {
  "Authorization": f"Bearer {api_key}" if api_key else settings.openai.bearer_token,
  "OpenAI-Organization": self.organization,
  "OpenAI-Project": project or settings.openai.project,
}
```

src/guidellm/backend/openai.py
```diff
@@ -110,20 +111,36 @@ def __init__(
...
+
         self.timeout = timeout if timeout is not None else settings.request_timeout
         self.http2 = http2 if http2 is not None else settings.request_http2
         self.follow_redirects = (
             follow_redirects
             if follow_redirects is not None
             else settings.request_follow_redirects
         )
+        self.verify = kwargs.pop("verify", settings.openai.verify)
```
sjmonson: Due to above suggestion:

```suggestion
        self.verify = verify or settings.openai.verify
```

sjmonson (changes requested):
Overall looks pretty good, just a few design nits. To fix markdown linting run `python -m mdformat <files>`

sjmonson (approved):

Issue #222, opened 2025-07-08
Different prompts generated for different concurrency levels despite same seed/config


Harshith-umesh:
## Summary
When running the concurrent benchmark across multiple concurrency levels (1, 2, 4, 8, 16, 32, 64), guidellm generates different prompts for each concurrency level, even when using the same seed/ synthetic data generation configuration. This makes it impossible to do fair "apples-to-apples" comparisons across different concurrency levels.

## Expected Behavior
When using the same seed and data configuration, all concurrency levels should generate the same set of prompts to ensure fair comparison across different concurrency levels.

## Evidence from Benchmark Results

Here's the analysis from a benchmark run with concurrency levels [1, 2, 4, 8, 16, 32, 64]:

### Configuration (identical for all benchmarks):
```json
{
  "request_loader": {
    "type_": "generative_request_loader",
    "data": "{\"prompt_tokens\":1000,\"prompt_tokens_min\":1,\"prompt_tokens_max\":4096,\"output_tokens\":1000,\"output_tokens_min\":1,\"output_tokens_max\":4096}",
    "processor": "meta-llama/Llama-3.1-8B-Instruct",
    "data_args": null,
    "processor_args": null
  }
}
```

### Different First Prompts by Concurrency Level:

- **Concurrency 1**: "such a sacrifice to her advantage as years of gratitude cannot enough acknowledge. By this time she is actually with them! If such goodness does not m..."

- **Concurrency 2**: ". How unlucky that you should have a reasonable answer to give, and that I should be so reasonable as to admit it! But I wonder how long you _would_ h..."

- **Concurrency 4**: "rejection. He who, she had been persuaded, would avoid her as his greatest enemy, seemed, on this accidental meeting, most eager to preserve the acqua..."

- **Concurrency 8**: "aunt now began; and each of them pronounced him to be infinitely superior to anything they had expected. He is perfectly well behaved,..."

- **Concurrency 16**: "you speak ill of a human being in my life. I would wish not to be hasty in censuring anyone; but I always speak what I think. I know you do and it is ..."

- **Concurrency 32**: "herself the reason, she had very little to say in reply. Mr. Gardiner left them soon after breakfast. The fishing scheme had been renewed the day befo..."

- **Concurrency 64**: ". Bingley did not take the house so much for the convenience of the neighbourhood as for his own, and we must expect him to keep or quit it on the sam..."

### Environment:
- Platform: OpenShift
[...]

PR #241, opened 2025-07-25
Add consistent synthetic data flag

Harshith-umesh:
This PR Introduces a new CLI flag `--consistent-synthetic-data` that ensures synthetic datasets generate the same prompts across different concurrency levels, enabling fair performance comparisons in benchmarks.

When benchmarking with synthetic data across multiple concurrency rates, different prompts were generated for each rate, making it difficult to compare performance fairly since the workload varied between tests.
Related Issue: https://github.com/vllm-project/guidellm/issues/222 

### Solution

Added a new boolean flag that, when enabled, resets the synthetic data iterator for each concurrency level to ensure consistent prompt generation.

###  Implementation

- Flag defaults to `False` to maintain backward compatibility
- Only affects synthetic datasets; regular datasets are unaffected
- Only resets iterator for infinite iter_type with synthetic data

### Testing
Comprehensive tests to verify that consistent prompts are generated across iterations when enabled and ensure existing functionality remains unchanged.

## Usage

```bash
# Enable consistent synthetic data generation
guidellm run --target http://localhost:8000 \
  --data '{"prompt_tokens": 50, "output_tokens": 25}' \
  --rate-type concurrent 
  --rate 1,2,4,8 \
  --consistent-synthetic-data
```


sjmonson:
I gave some advice offline regarding nesting args under `--data` but after reviewing this pull request I realized that there are other issues. Leaving the CLI aside for now the core of #222 is that we don't reset the dataset iterator between sub-benchmarks. Its a problem regardless of synthetic or real data. What we really need is a way of resetting our `RequestLoader` at the beginning of each loop in `Benchmarker.run`. Probably the best way to do this is add a new method `RequestLoader.reset()` and call it in `Benchmarker.run`. For `GenerativeRequestLoader` it will look something like:

```python
def reset(self):
    self._preserved_iter = None
```

sjmonson (changes requested):
See previous comment

PR #246, opened 2025-07-28
add custom dict camelize logic

DaltheCow:
This removes reliance on the library pyhumps converting dicts keys from snake_case to camelCase for use in the UI


src/guidellm/utils/dict.py
```diff
@@ -0,0 +1,23 @@
...
+
+  updated_key_pairs = []
+  for key, _ in d.items():
+    updated_key = key_update_func(key)
+    if key != updated_key:
+      updated_key_pairs.append((key, updated_key ))
+
+  for key_pair in updated_key_pairs:
+    old_key, updated_key = key_pair
+    d[updated_key] = d[old_key]
```
sjmonson: Small nit. Might want to throw an exception here in the case that `updated_key` already exists. Could save a bunch of debugging time later.

sjmonson (changes requested):

src/guidellm/utils/dict.py
DaltheCow: Isn't this handled by line 13 where we prevent any addition to updated_key_pairs if the old key and updated key are the same? So it is impossible for updated_key_pairs to contain anything but updated keys I would say.

DaltheCow: Also sort of a sidenote but it will be pretty common for keys to already exist in the case of a camelizing function being the update function, since for example `benchmarks` would not require any updating and that is an expected behavior that we don't need to update every key if it is already the same as the updated version

sjmonson (dismissed):

src/guidellm/utils/dict.py
sjmonson: Ack

sjmonson (approved):

PR #250, opened 2025-07-30
Drop Entrypoint and Improve Containerfile

sjmonson:
* Drop the container entrypoint script as GuideLLM has had native support for its features since #99 
* Make containerfile more rebuild friendly based on #213 
* Drop the ENV default scenario as it is confusing to users setting CLI args

Closes: #213


Harshith-umesh (approved):
LGTM!

Issue #196, opened 2025-06-24
Long running test spawn too many threads


sjmonson:
The `SyncManager` process spawns a huge number of threads over the lifetime of the benchmark. On pid constrained systems (such as OpenShift) we can run out of threads during longer tests or with shorter requests.

sjmonson:
This issue still occurs after replacing the `SyncManager` with shared queues.

sjmonson:
After a bunch of testing I've found that the issue only occurs with Python 3.12.

markurtz:
Need to double check this after #251 lands

PR #254, opened 2025-07-31
Various fixes to container image and CI 

sjmonson:
## Summary

Various fixes to containers images and CI. Containers images should now reuse most layers on build, need less hacks when running in OCP, and properly detect the guidellm version. Release and RC git tags/branches tag the container image with the version. The `latest` and `stable` tags now smartly update based on highest version tag (prev most recent build).

## Details

- Moves Containerfile to top-level
- Adds a `.containerignore` that mirrors `.gitignore`
- Fixes version detection in image builds
- Correctly sets release type and tags rc/release with version when building images
- Use PDM to lock venv in image builds
- Adds weekly job for tagging latest and stable based on existing version tagged images
- Adds weekly job to prune dev containers that are more than 2 weeks old
    - Note: This job is currently set to dry-run and will be enabled in a future PR
- Improves container compatibility with K8s unpriv users
- Reorders image layers to improve build caching

### 1. Workflows 

- Workflows were partially tested with [act](https://github.com/nektos/act) but need to be run on GitHub to verify

### 2. Container Changes 

1. Run `podman run --rm ghcr.io/vllm-project/guidellm:pr-254 --version` to verify version is set
2. Run `podman run --rm --entrypoint /opt/app-root/guidellm/bin/pip ghcr.io/vllm-project/guidellm:pr-254 freeze` and verify that versions match the `pylock.toml`.
3. Run a pod with the GuideLLM container and verify the user has write permissions to `/home/guidellm`. (For tokenizer/dataset caching)

## Related Issues

- Resolves #

---

## Use of AI

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -16,7 +16,7 @@ USER root
 
 # Install build tooling
 RUN dnf install -y git \
-    && python3 -m venv /tmp/pdm \
+    && /usr/bin/python3 -m venv /tmp/pdm \
     && /tmp/pdm/bin/pip install --no-cache-dir -U pdm \
     && ln -s /tmp/pdm/bin/pdm /usr/local/bin/pdm
 
@@ -27,11 +27,11 @@ ENV PDM_CHECK_UPDATE=false \
 
 # Copy repository files
 # Do this as late as possible to leverage layer caching
-COPY / /opt/app-root/src
+COPY / /src
 
 # Install guidellm and locked dependencies
-RUN pdm use -p /opt/app-root/src -f /opt/app-root \
-    && pdm install -p /opt/app-root/src --check --prod --no-editable
+RUN pdm use -p /src -f /opt/app-root \
+    && pdm install -p /src --check --prod --no-editable
 
 # Prod image
 FROM $BASE_IMAGE


.github/workflows/container-maintenance.yml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+      - name: Delete PR and untagged images older than 2 weeks
+        uses: snok/container-retention-policy@v3.0.0
+        with:
+          account: ${{ github.actor }}
+          token: ${{ github.token }}
+          image-names: ${{ github.event.repository.name }}
+          image-tags: "pr-*"
+          cut-off: 2w
+          dry-run: true
```
sjmonson, line 24: Set this to `dry-run` for now until we can confirm it deletes the correct set of images.

Containerfile
```diff
@@ -0,0 +1,65 @@
...
+
+# Use a multi-stage build to create a lightweight production image
+FROM $BASE_IMAGE as builder
+
+# Ensure files are installed as root
+USER root
+
+# Install build tooling
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends git \
```
sjmonson: I had hoped to keep this image distro-agnostic to enable downstream RHEL builds but we need `git` to have the correct package version tag after install. There are currently two alternatives I am considering:

1. Write a helper script that implements `apt` and `dnf` based installs based on `/etc/os-release`.
2. Change base_image to a community `dnf` based image.

I prefer (2) but currently there are not any minimal 3.13 images based on Fedora/CentOS/RHEL. The closest is [`quay.io/fedora/python-312-minimal`](https://quay.io/repository/fedora/python-312-minimal). I've filled https://github.com/sclorg/s2i-python-container/pull/753 with the project to hopeful enable `python-313-minimal` builds.

jaredoconnell: Is it possible to use a 3.12 base image instead?

sjmonson: Various reasons:

* In the past we have run into issues with 3.12 (ref: #196)
* I do not want to downgrade from our existing image
* 3.13 is what the CI runs tests on

sjmonson: Resolved this by self-publishing the `fedora-313-minimal` container. We can switch to the official image when its out (See https://github.com/sclorg/s2i-python-container/pull/756).

jaredoconnell (dismissed):
I don't love the burden of maintaining our own version of the python container. Is it set up to automatically update when the base container updates? If not there are security concerns.
The bug you referenced regarding Python 3.12 appears to be resolved, according to Mark.
I don't see any value to using the state of the art python version when the project is backwards compatible to 3.9, but I won't block you on it.

markurtz (approved):

PR #255, opened 2025-07-31
Run full tests on PRs & main

sjmonson:
Run the full unit test suite on all PRs and merges. Currently nightly is failing due to failing tests that were not caught.


jaredoconnell (approved):
This seems like the right move to me.

PR #266, opened 2025-08-08
Fix a Couple Errors With Metric Accumulation

sjmonson:
## Details

- [ ] 

## Test Plan

-

## Related Issues

- Resolves #

---

## Use of AI

### DIFF
--- a/src/guidellm/benchmark/benchmark.py
+++ b/src/guidellm/benchmark/benchmark.py
@@ -817,7 +817,7 @@ class GenerativeBenchmark(Benchmark):
                     ],
                     iter_counts=[req.output_tokens for req in total_with_output_first],
                     first_iter_counts=[
-                        req.prompt_tokens for req in total_with_output_first
+                        req.prompt_tokens + 1 for req in total_with_output_first
                     ],
                 ),
             ),
--- a/src/guidellm/objects/statistics.py
+++ b/src/guidellm/objects/statistics.py
@@ -219,7 +219,7 @@ class DistributionSummary(StandardBaseModel):
         )
 
     @staticmethod
-    def from_request_times(
+    def from_request_times(  # noqa: C901
         requests: list[tuple[float, float]],
         distribution_type: Literal["concurrency", "rate"],
         include_cdf: bool = False,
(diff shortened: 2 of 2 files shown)


src/guidellm/objects/statistics.py
```diff
@@ -248,13 +248,7 @@ def from_request_times(
...
                 time_deltas[end] -= 1
 
-            # convert to the events over time measuring concurrency changes
-            events = []
-            active = 0
-
-            for time, delta in sorted(time_deltas.items()):
-                active += delta
-                events.append((time, active))
+            events = list(time_deltas.items())
```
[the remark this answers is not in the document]
sjmonson: Already sorted below

src/guidellm/benchmark/benchmark.py
```diff
@@ -817,7 +817,7 @@ def from_stats(
                     ],
                     iter_counts=[req.output_tokens for req in total_with_output_first],
                     first_iter_counts=[
-                        req.prompt_tokens for req in total_with_output_first
+                        req.prompt_tokens + 1 for req in total_with_output_first
```
jaredoconnell: Can you add a comment explaining why the plus one is required?

src/guidellm/objects/statistics.py
```diff
@@ -281,6 +275,16 @@ def from_request_times(
             else:
                 flattened_events.append((time, val))
 
+        if distribution_type == "concurrency":
+            # convert to the events over time measuring concurrency changes
+            events_over_time: list[tuple[float, float]] = []
+            active = 0
+            for time, delta in flattened_events:
+                active += delta  # type: ignore [assignment]
```
jaredoconnell, line 280: How did moving the code here help with the fix? Does the code above not get negatively affected by the code using delta instead of active time?

sjmonson: This is the fix; the flattened events section assumes that each event is independent. Specifically line 274, `flattened_events[-1] = (last_time, last_val + val)`, `last_val + val` will cause double counting whenever an event is flattened.

src/guidellm/benchmark/benchmark.py
```diff
@@ -817,7 +817,9 @@ def from_stats(
                     ],
                     iter_counts=[req.output_tokens for req in total_with_output_first],
                     first_iter_counts=[
-                        req.prompt_tokens for req in total_with_output_first
+                        # prompt tokens + first token
```
markurtz, line 820: Yeah, good catch, found this in the refactor as well. We'll need to subtract one from the output_tokens as well since we are counting the first token here

sjmonson: Actually I believe this is already handled in the `from_iterable_request_times` function:

https://github.com/vllm-project/guidellm/blob/086d2b5b79d9b5e904b0f3a1d098b80c86638e73/src/guidellm/objects/statistics.py#L370-L373

PR #285, opened 2025-08-19
add dataset and model size to run info

DaltheCow:
## Summary

The run info will now grab the dataset and if the model is pulled from HF it will grab the model size

## Details

- Dataset added to run info
- Model size added to run info

## Test Plan

- Function which grabs dataset is tested


DaltheCow:
Closing, now refer to #391

Issue #274, opened 2025-08-12
UI erroneously labels ITL as TPOT


sjmonson:
**Describe the bug**
Graphs in the benchmark UI show TPOT in places which are actually showing ITL.

**Expected behavior**
All graphs that are labeled with "TPOT" should instead be labeled with "ITL".

DaltheCow:
Fix here: https://github.com/vllm-project/guidellm/pull/288
Opted to switch the data from ITL to TPOT since it is the simplest fix and then we don't have to use ITL when we've always had TPOT in the UI. If ITL is preferred though I can make that change

DaltheCow:
New PR up for the fix: #386

PR #288, opened 2025-08-20
switch metric to pull from tpot instead of itl

DaltheCow:
## Summary

This is a fix for this: https://github.com/vllm-project/guidellm/issues/274
Instead of switching the labels from TPOT to ITL I switched the data to use TPOT data as it should have been in the first place. But if we prefer ITL over TPOT then I can make that switch, I just wanted to avoid making the change since it wasn't necessary


markurtz (changes requested):
Per @sjmonson, change to ITL and not TPOT

PR #289, opened 2025-08-20
Scheduler refactor [utils]: auto_importer, pydantic_utils, registry, singleton

markurtz:
## Summary

This PR introduces four new utility modules that provide foundational capabilities for auto-discovery, registry patterns, singleton management, and enhanced Pydantic utilities. It additionally consolidates the pydantic utility classes from guidellm.objects into the dedicated guidellm.utils module containing the new additions. This change improves code organization and establishes the foundation for upcoming scheduler refactoring work.

## Details

- **Moved core utilities** from `guidellm.objects.pydantic` to `guidellm.utils.pydantic_utils`
- **Added `AutoImporterMixin`** for dynamic module discovery within packages
- **Added `RegistryMixin`** for object registration with optional auto-discovery
- **Added singleton mixins** (`SingletonMixin`, `ThreadSafeSingletonMixin`) for instance management
- **Enhanced Pydantic utilities** with `ReloadableBaseModel`, `StandardBaseDict`, and `PydanticClassRegistryMixin` for polymorphic model serialization
- **Updated all imports** across the codebase to reference new module locations
- **Added comprehensive test coverage** for all new utility modules
- **Removed obsolete** `guidellm.objects.pydantic` module and associated tests

## Test Plan

- Run existing test suite to ensure no regressions from import changes

## Related Issues

---

## Use of AI

### DIFF
--- a/src/guidellm/utils/pydantic_utils.py
+++ b/src/guidellm/utils/pydantic_utils.py
@@ -28,7 +28,6 @@ __all__ = [
 
 
 BaseModelT = TypeVar("BaseModelT", bound=BaseModel)
-T = TypeVar("T", bound=BaseModel)
 SuccessfulT = TypeVar("SuccessfulT")
 ErroredT = TypeVar("ErroredT")
 IncompleteT = TypeVar("IncompleteT")
@@ -90,7 +89,7 @@ class StandardBaseModel(BaseModel):
     )
 
     @classmethod
-    def get_default(cls: type[T], field: str) -> Any:
+    def get_default(cls: type[BaseModelT], field: str) -> Any:
         """
         Get default value for a model field.
 
--- a/src/guidellm/utils/singleton.py
+++ b/src/guidellm/utils/singleton.py
@@ -94,9 +94,6 @@ class ThreadSafeSingletonMixin(SingletonMixin):
         lock_attr_name = f"_singleton_lock_{cls.__name__}"
         instance_attr_name = f"_singleton_instance_{cls.__name__}"
 
-        if not hasattr(cls, lock_attr_name):
-            setattr(cls, lock_attr_name, threading.Lock())
-
         with getattr(cls, lock_attr_name):
             instance_exists = (
                 hasattr(cls, instance_attr_name)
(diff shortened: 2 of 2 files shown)


src/guidellm/utils/pydantic_utils.py
```diff
@@ -0,0 +1,303 @@
...
+    "PydanticClassRegistryMixin",
+    "ReloadableBaseModel",
+    "StandardBaseDict",
+    "StandardBaseModel",
+    "StatusBreakdown",
+]
+
+
+BaseModelT = TypeVar("BaseModelT", bound=BaseModel)
+T = TypeVar("T", bound=BaseModel)
```
sjmonson: Drop the redundant `T` and just use `BaseModelT` in its place.

src/guidellm/utils/singleton.py
```diff
@@ -0,0 +1,128 @@
...
+        :return: The singleton instance of the class
+        """
+        # Use class-specific lock and instance names to avoid inheritance issues
+        lock_attr_name = f"_singleton_lock_{cls.__name__}"
+        instance_attr_name = f"_singleton_instance_{cls.__name__}"
+
+        if not hasattr(cls, lock_attr_name):
+            setattr(cls, lock_attr_name, threading.Lock())
+
+        with getattr(cls, lock_attr_name):
```
sjmonson: Potential thread safety issue if free-threading. If two threads eval `not hasattr(cls, lock_attr_name)` to true and one of those threads makes it through line 100 we will have a split lock. AFAIK this is pretty much impossible to solve with Python's existing atomic primitives; we either need a way to manually lock the GIL or an atomic compare-and-swap. So for now maybe just add a note, its a very unlikely occurrence.

sjmonson: Actually found a way to make this work [\_\_init_subclass__](https://docs.python.org/3/reference/datamodel.html#object.__init_subclass__) is called once on the type instantiation of each subclass so the following should work:

```python

class ThreadSafeSingletonMixin(SingletonMixin):
    def __init_subclass__(cls, /, **kwargs):
        super().__init_subclass__(**kwargs)
        lock_attr_name = f"_singleton_lock_{cls.__name__}"
        setattr(cls, lock_attr_name, threading.Lock())

    def __new__(cls, *args, **kwargs): # noqa: ARG004
        # Everything above minus lock setup
        ...
```

sjmonson (changes requested):

src/guidellm/utils/singleton.py
markurtz: Nice, good find, added in

sjmonson (approved):

PR #290, opened 2025-08-21
Scheduler refactor [utils]: functions, mixins, statistics, text

markurtz:
## Summary

This PR adds new utility modules for safe math function operations, object info extraction, and reorganizes to consolidate into the utils package.

## Details

- **Move statistics classes from `objects` to `utils`**: Relocates `DistributionSummary`, `Percentiles`, `RunningStats`, `StatusDistributionSummary`, and `TimeRunningStats` from `guidellm.objects.statistics` to `guidellm.utils.statistics`
- **Add new `functions.py` module**: Implements defensive programming utilities including `safe_getattr`, `safe_divide`, `safe_multiply`, `safe_add`, `safe_format_timestamp`, and `all_defined` for handling None values and edge cases
- **Add new `mixins.py` module**: Provides `InfoMixin` class for standardized metadata extraction and object introspection across different class hierarchies
- **Enhance `text.py` module**: Adds comprehensive documentation, `format_value_display` function for consistent metric formatting, and improved text processing utilities
- **Update import statements**: Modifies all affected modules (`benchmark`, `presentation`) to import statistics classes from their new location in `utils`
- **Remove deprecated `objects` package**: Deletes the now-empty `objects` directory and associated test files
- **Add comprehensive test coverage**: Includes new test suites for `functions.py`, `mixins.py`, and `statistics.py`
- **Update `__init__.py` exports**: Adds new utility functions and classes to the main utils package exports for easy access

## Test Plan

- Run the existing test suite to ensure no regressions from the statistics class relocation
- Execute new test files:
    - test_functions.py - Tests for safe operation utilities
    - test_mixins.py - Tests for InfoMixin functionality
    - test_statistics.py - Comprehensive tests for statistical analysis utilities
    - test_text.py - Tests for enhanced text processing functions

## Related Issues

- Resolves #

---

- [X] "I certify that all code in this PR is my own, except as noted below."

## Use of AI

### DIFF
--- a/src/guidellm/utils/mixins.py
+++ b/src/guidellm/utils/mixins.py
@@ -12,6 +12,10 @@ from typing import Any
 __all__ = ["InfoMixin"]
 
 
+PYTHON_PRIMITIVES = (str, int, float, bool, list, tuple, dict)
+"""Type alias for serialized object representations"""
+
+
 class InfoMixin:
     """
     Mixin class providing standardized metadata extraction for introspection.
@@ -58,9 +62,7 @@ class InfoMixin:
             "module": obj.__class__.__module__ if hasattr(obj, "__class__") else None,
             "attributes": (
                 {
-                    key: val
-                    if isinstance(val, (str, int, float, bool, list, dict))
-                    else repr(val)
+                    key: val if isinstance(val, PYTHON_PRIMITIVES) else repr(val)
                     for key, val in obj.__dict__.items()
                     if not key.startswith("_")
                 }


src/guidellm/utils/functions.py
```diff
@@ -0,0 +1,133 @@
...
+    :param default: Value to return if timestamp is invalid or None
+    :return: Formatted timestamp string or default value
+    """
+    if timestamp is None or timestamp < 0 or timestamp > 2**31:
+        return default
+
+    try:
+        return datetime.fromtimestamp(timestamp).strftime(format_)
+    except (ValueError, OverflowError, OSError):
+        return default
```
sjmonson: The actual timestamp min/max is platform dependent. On my system its `[-2*35, 2**37]`. Maybe just drop the first check all together and add `TypeError` (for None) to the try/except.

```suggestion
    try:
        return datetime.fromtimestamp(timestamp).strftime(format_)
    except (ValueError, TypeError, OverflowError, OSError):
        return default
```

src/guidellm/utils/mixins.py
```diff
@@ -0,0 +1,113 @@
...
+        return {
+            "str": str(obj),
+            "type": type(obj).__name__,
+            "class": obj.__class__.__name__ if hasattr(obj, "__class__") else None,
+            "module": obj.__class__.__module__ if hasattr(obj, "__class__") else None,
+            "attributes": (
+                {
+                    key: val
+                    if isinstance(val, (str, int, float, bool, list, dict))
+                    else str(val)
```
sjmonson:
```suggestion
                    else repr(val)
```

src/guidellm/utils/mixins.py
```diff
@@ -0,0 +1,113 @@
...
```
sjmonson:
```suggestion
                    else repr(val)
```

src/guidellm/utils/mixins.py
```diff
@@ -0,0 +1,113 @@
...
+
+        return {
+            "str": str(obj),
+            "type": type(obj).__name__,
+            "class": obj.__class__.__name__ if hasattr(obj, "__class__") else None,
+            "module": obj.__class__.__module__ if hasattr(obj, "__class__") else None,
+            "attributes": (
+                {
+                    key: val
+                    if isinstance(val, (str, int, float, bool, list, dict))
```
sjmonson: Also `tuple`? Also maybe make this a TypeDef to ensure its stays constant between both methods.

sjmonson (changes requested):

src/guidellm/utils/text.py
```diff
@@ -1,45 +1,108 @@
...
+        if len(result) < total_characters:
+            result = result.rjust(total_characters)
+
+    return result
 
 
 def split_text_list_by_length(
     text_list: list[Any],
-    max_characters: Union[int, list[int]],
+    max_characters: int | list[int],
```
jaredoconnell, line 88: If you do this then we will need to bump the minimum Python version to 3.10. Are we ready for that commitment?

Also applies to several other instances in this file.

I'd say keeping 3.9 would be easier since it's the default RHEL Python version.

jaredoconnell (approved):
Once my concerns and Samuel's concerns are addressed, I think this is ready for merging.
I want to confirm that we're okay with switching to Python 3.10. Otherwise these changes would need to be revised. I don't love the idea of moving to 3.10.

src/guidellm/utils/text.py
markurtz: Unfortunately we can't give up 3.9 until October with EOL. The from future import \__annotations__ allows us to use the | where Unions/Optionals would be for Python 3.9 as well as some other nice features such as all types automatically being string types preventing some more issues

jaredoconnell: Thank you for clarifying. Good that it's possible to opt into that feature for 3.9.

src/guidellm/utils/mixins.py
markurtz: Makes sense, fixed -- need to create it as a tuple var, though, due to lack of support on Python 3.9 with isinstance and type aliases

sjmonson (approved):

PR #293, opened 2025-08-26
Scheduler refactor [utils]: multiprocessing encoding and messaging

markurtz:
## Summary

This PR introduces four new utility modules that provide improved capabilities for transferring information across processes to maximize performance. This is done through encoding.py to best serialize the data into a transferrable format and minimize the size of the data as well as through messaging.py which supports high performance multiprocessing queues and pipes for transfer that works with the encoding.

## Details

- Added encoding.py with the core `MessageEncoding.py` 
- Added messaging.py with `InterProcessMessaging`, `InterProcessMessagingManagerQueue`, `InterProcessMessagingPipe`, `InterProcessMessagingQueue`
- Added unit tests for all of the above classes and common data transfer use cases.

## Test Plan

## Related Issues

---

- [ X] "I certify that all code in this PR is my own, except as noted below."

## Use of AI

- [ X] Includes code generated by an AI application

### DIFF
--- a/src/guidellm/utils/encoding.py
+++ b/src/guidellm/utils/encoding.py
@@ -12,7 +12,7 @@ from __future__ import annotations
 
 import json
 from collections.abc import Mapping
-from typing import Annotated, Any, ClassVar, Generic, Literal, Optional, TypeVar, Union
+from typing import Annotated, Any, ClassVar, Generic, Literal, Optional, TypeVar
 
 try:
     import msgpack
--- a/src/guidellm/utils/encoding.py
+++ b/src/guidellm/utils/encoding.py
@@ -1,0 +1,785 @@
+"""
+Message encoding utilities for multiprocess communication with Pydantic model support.
+
+Provides binary serialization and deserialization of Python objects using various
+serialization formats and encoding packages to enable performance configurations
+for distributed scheduler operations. Supports configurable two-stage processing
+pipeline: object serialization (to dict/sequence) followed by binary encoding
+(msgpack/msgspec) with specialized Pydantic model handling for type preservation.
+"""
+
+from __future__ import annotations
+
+import json
+from collections.abc import Mapping
+from typing import Annotated, Any, ClassVar, Generic, Literal, Optional, TypeVar, Union
+
+try:
+    import msgpack
+    from msgpack import Packer, Unpacker
+
+    HAS_MSGPACK = True
+except ImportError:
+    msgpack = Packer = Unpacker = None
+    HAS_MSGPACK = False
+
+try:
+    from msgspec.msgpack import Decoder as MsgspecDecoder
+    from msgspec.msgpack import Encoder as MsgspecEncoder
+
+    HAS_MSGSPEC = True
+except ImportError:
+    MsgspecDecoder = MsgspecEncoder = None
+    HAS_MSGSPEC = False
+
+try:
+    import orjson
+
+    HAS_ORJSON = True
+except ImportError:
+    orjson = None
+    HAS_ORJSON = False
+
+from pydantic import BaseModel
+from typing_extensions import TypeAlias
+
+__all__ = [
+    "Encoder",
+    "EncodingTypesAlias",
+    "MessageEncoding",
+    "SerializationTypesAlias",
+    "Serializer",
+]
+
+ObjT = TypeVar("ObjT")
+MsgT = TypeVar("MsgT")
+
+SerializationTypesAlias: TypeAlias = Annotated[
+    Optional[Literal["dict", "sequence"]],
+    "Type alias for available serialization strategies",
+]
+EncodingTypesAlias: TypeAlias = Annotated[
+    Optional[Literal["msgpack", "msgspec"]],
+    "Type alias for available binary encoding formats",
+]
+
+
+class MessageEncoding(Generic[ObjT, MsgT]):
+    """
+    High-performance message encoding and decoding for multiprocessing communication.
+
+    Supports configurable object serialization and binary encoding with specialized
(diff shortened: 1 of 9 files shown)


src/guidellm/utils/encoding.py
```diff
@@ -0,0 +1,785 @@
...
+            for model in pydantic_models:
+                self.register_pydantic(model)
+
+    def register_pydantic(self, model: type[BaseModel]) -> None:
+        """
+        Register Pydantic model for specialized serialization handling.
+
+        :param model: Pydantic model class to register for type preservation
+        """
+        key = f"{model.__module__}:{model.__name__}"
```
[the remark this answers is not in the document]
sjmonson: Tuple is probably more efficient than string interpolation in this case. It does look like that was your intention in \_\_init__?

src/guidellm/utils/encoding.py
```diff
@@ -0,0 +1,785 @@
...
+            isinstance(value, BaseModel) for value in obj.values()
+        ):
+            return {
+                key: self.to_dict_pydantic(value)
+                if isinstance(value, BaseModel)
+                else value
+                for key, value in obj.items()
+            }
+
+        return obj
```
sjmonson, line 421: ~I'm thinking this might be a bit faster with recursion + a cache. Plus handles nested better if we ever need that.~ Actually I don't want to argue to hard in favor of this without data to back it up. Maybe ignore this one for now and I'll try to do a study after release to see which method tailors better to our usecase.

```suggestion
    def to_dict(self, obj: Any, _cache: Optional[dict[int, Any]] = None) -> Any:
        """
        Convert object to dictionary with Pydantic model type preservation.
        :param obj: Object to convert (BaseModel, collections, or primitive)
        :return: Dictionary representation with type metadata for Pydantic models
        """
        cache = _cache if _cache is not None else {}
        def cached_to_dict(obj: Any):
            if id(obj) in cache:
                return cache[id(obj)]
            cache[id(obj)] = None # Avoid infinite recursion
            d = self.to_dict(obj, cache)
            cache[id(obj)] = d
            return d
        
        if isinstance(obj, BaseModel):
            return self.to_dict_pydantic(obj)
        
        if isinstance(obj, (list, tuple)):
            return [ cached_to_dict(item) for item in obj ]

        if isinstance(obj, dict):
            return { key: cached_to_dict(value) for key, value in obj.items() }

        return obj
```

src/guidellm/utils/encoding.py
```diff
@@ -0,0 +1,785 @@
...
+        Handles deserialization of objects encoded with to_sequence, properly
+        restoring Pydantic models and collection structures.
+
+        :param data: Serialized sequence data to reconstruct
+        :return: Reconstructed object with proper types
+        :raises ValueError: If sequence format is invalid or contains multiple
+            packed sequences
+        """
+        type_, payload, remaining = self.unpack_next_sequence(data)
+        if remaining:
```
sjmonson:
```suggestion
        if remaining is not None:
```

src/guidellm/utils/encoding.py
```diff
@@ -0,0 +1,785 @@
...
+                        if is_pydantic
+                        else self.to_sequence_python(item)
+                    ),
+                    current=payload,
+                )
+        else:
+            payload_type = "python"
+            payload = self.to_sequence_python(obj)
+
+        return self.pack_next_sequence(payload_type, payload, None)
```
sjmonson, line 528: Same points as above regarding recursion.

src/guidellm/utils/encoding.py
sjmonson, line 1: Overall does look good but since this is a very latency critical section of code some mirco-benchmark data to back it up would be nice. Since we are running a bit short on time we can take that as a post release step.

sjmonson (changes requested):

src/guidellm/utils/encoding.py
markurtz: Going to resolve for now and then we can talk through more later. There isn't any active recursion happening here -- it's only supporting Pydantic models within the first level of a collection and therefore only one call level since that's where we need to intercept the serialization. Anything nested is assumed to be either handled by pydantic or fully a python primitive for serialization. I'll add a note in towards this

src/guidellm/utils/encoding.py
markurtz: See the note above, it's not doing any active recursion currently and just supports a pydantic model at the first level of a collection, anything lower would not work due to the performance implications for recursing through that structure in Python

src/guidellm/utils/encoding.py
markurtz: Definitely agree, I have some initial work towards this at least for e2e, but would be good to get some time breakdowns and see if there's anything to improve

src/guidellm/utils/encoding.py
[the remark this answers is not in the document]
markurtz: That's fair, can revert back to the tuple

src/guidellm/utils/messaging.py
```diff
@@ -0,0 +1,1056 @@
...
+    async def get(self, timeout: float | None = None) -> ReceiveMessageT:
+        """
+        Retrieve message from receive buffer with optional timeout.
+
+        :param timeout: Maximum time to wait for a message
+        :return: Decoded message from the receive buffer
+        """
+        return await asyncio.wait_for(
+            self.buffer_receive_queue.async_get(), timeout=timeout
+        )
```
jaredoconnell, line 251: Does this function return a `TimeoutError` when it times out? If so I think it would be beneficial to document that on all of these functions that have a timeout.

jaredoconnell (approved):
Looks good. Just one minor comment.

PR #294, opened 2025-08-26
Scheduler refactor [scheduler]: constraints, objects, and strategies

markurtz:
## Summary

This PR introduces the new concept of constraints and request timings for the scheduler refactor while reworking the base objects and strategies needed to work with the refactor. The implementations are setup now so that Strategy classes return request timings instances which are then used with in the worker processes; the core reason for this is to ensure scheduled request times are calculated per request rather than precalculated which disables any ability to dynamically adjust rates. Additionally, the constraints are added in which are now used to control when the scheduler stops running rather than the hardcoded implementations for number of requests and duration. These will enable much more complicated implementations as seen in the added error constraints.

## Details

- Added constraints.py module with the new implementations and constructs including Protocols for constraint and initializer implementations. Additionally added constraints for number of requests, benchmark duration, total number of errors, error rate (both global and window)
- Refactored result.py and types.py into the new objects.py and expanded their capabilities. The scheduler package itself is now fully generic and not tied to the backend package enabling more flexibility. Additionally, object types have been expanded out to include more details for benchmarking.
- Refactored strategy.py to introduce the concept of request timings and reimplemented the strategy types to work with the new pathways.

## Test Plan

## Related Issues

---

- [X ] "I certify that all code in this PR is my own, except as noted below."

## Use of AI

- [ X] Includes code generated by an AI application


src/guidellm/scheduler/objects.py
```diff
@@ -0,0 +1,446 @@
...
+        list[Union[RequestT, tuple[RequestT, float]]],
+        tuple[Union[RequestT, tuple[RequestT, float]]],
+    ],
+    type_params=(RequestT,),
+)
+"""Multi-turn request structure supporting conversation history with optional delays."""
+
+
+class RequestSchedulerTimings(StandardBaseModel):
+    """Scheduler-level timing measurements for request lifecycle tracking."""
```
jaredoconnell: Is the unit for all of these fields epoch time in seconds?
It may make sense to briefly mention that.

src/guidellm/scheduler/objects.py
```diff
@@ -0,0 +1,446 @@
...
+                "scheduler_timings": self.scheduler_timings.model_copy(),
+                "request_timings": (
+                    self.request_timings.model_copy() if self.request_timings else None
+                ),
+            },
+            deep=False,
+        )
+
+
+class BackendInterface(ABC, Generic[RequestT, MeasuredRequestTimingsT, ResponseT]):
```
sjmonson: This should be a `Protocol` not an `ABC` right? Since its not implemented by the Backend base class.

sjmonson (changes requested):
We have been over this section quite a few times before so looks good aside from one minor nit. I did not check the constraint calculations to closely @AlonKellner-RedHat if you could quickly run over them.

src/guidellm/scheduler/strategy.py
```diff
@@ -1,493 +1,700 @@
...
-            for immediate request scheduling.
+        :param local_rank: The rank of the worker process
+        :param local_world_size: Total number of worker processes
+        :param local_max_concurrency: Maximum concurrent requests for the worker
+        :return: NoDelayRequestTimings instance for immediate request scheduling
         """
-        start_time = time.time()
+        if self.startup_duration > 0:
+            # Vary offset by up to 5% of the startup duration for a bit of variance
+            offset = 0.05 * self.startup_duration * (local_rank / local_world_size)
```
jaredoconnell, line 574: Do we want this hard-coded?

jaredoconnell (approved):
Looks good. Just a few minor comments.

src/guidellm/scheduler/objects.py
markurtz: Yep, thanks, added a note

src/guidellm/scheduler/objects.py
markurtz: Yep, good catch, updated

src/guidellm/scheduler/strategy.py
markurtz: Yeah, we'll need to rework this logic in the future, but there's no way to enable it currently, so I'd say let's revisit it after this goes out

markurtz:
Resetting the refactor state with new PRs, closing this one out

PR #299, opened 2025-08-28
Automatically resolve pylock if check fails

sjmonson:
## Summary

Currently if the package requirements change in `pyproject.toml` the pylock pre-commit hook will fail and require the user to manually apply a pylock update. This change automates the pylock update so the user should only need to add and commit the updated pylock.

## Details

- Update pylock pre-commit hook to attempt to update the pylock file when `pdm lock --check` fails.

## Test Plan

1. Adjust package dependencies in `pyproject.toml`
2. Run `pdm run pre-commit run pdm-lock-check --all-files` and validate that the pylock changed
4. Run `pdm run pre-commit run pdm-lock-check --all-files` and validate that the check passes

---

## Use of AI

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,6 +11,7 @@ repos:
   hooks:
     - id: pdm-lock-check
       name: check lock file matches pyproject
+      entry: sh -c "pdm lock -v --check || pdm lock --update-reuse"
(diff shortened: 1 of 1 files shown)


jaredoconnell:
Could this ever trigger a lock change when dependencies do not change?

Issue #315, opened 2025-09-11
KeyError: 'content' when processing streaming chat completions


git-jxj:
**Describe the bug**

When running a benchmark against a streaming-capable, OpenAI-compatible chat model, guidellm crashes with a KeyError: 'content'.

This occurs because the final delta chunk sent by the server to signal the end of the stream does not contain a content key, as is standard for the API. The current implementation in guidellm does not handle this case and attempts to access the missing key directly, leading to an unhandled exception.

**Expected behavior**

guidellm should handle the end-of-stream signal gracefully without crashing. It should complete the benchmark and report the results, correctly processing all content-bearing chunks and ignoring the final, content-less chunk.

**Environment**
Include all relevant environment information:
1. OS [e.g. Ubuntu 20.04]: 15.6.1
2. Python version [e.g. 3.12.2]: 3.10.17
3. guidellm version: 0.4.0.dev2

**To Reproduce**
Exact steps to reproduce the behavior:

1. Set up an OpenAI-compatible API endpoint that serves a chat model with streaming enabled. (e.g., using litellm to proxy a model).
2. Set the environment variable to force guidellm to use the chat completions endpoint:
   1     export GUIDELLM__PREFERRED_ROUTE=chat_completions
3. Run the guidellm benchmark command targeting your endpoint. The command will run and appear to work, but will crash just as it is about to finish.

guidellm benchmark run \
  --target "http://your-api-endpoint/v1" \
  --model "your-chat-model" \
  --rate-type "synchronous" \
  --max-requests 1

**Errors**

The following exception is raised at the end of the benchmark run:

    1 Traceback (most recent call last):
    2   File "/path/to/your/guidellm/./.venv/bin/guidellm", line 7, in <module>
    3     sys.exit(cli())
    4   ...
    5   File "/path/to/your/guidellm/src/guidellm/benchmark/entrypoints.py", line 71, in benchmark_generative_text
    6     await backend.validate()
    7   File "/path/to/your/guidellm/src/guidellm/backend/backend.py", line 138, in validate
    8     async for _ in self.chat_completions(  # type: ignore[attr-defined]
    9   File "/path/to/your/guidellm/src/guidellm/backend/openai.py", line 393, in chat_completions
   10     raise ex
   11   File "/path/to/your/guidellm/src/guidellm/backend/openai.py", line 374, in chat_completions
   12     async for resp in self._iterative_completions_request(
[...]

PR #316, opened 2025-09-11
fix(backend): Handle missing content in streaming delta

git-jxj:
## Summary

This PR fixes a `KeyError: 'content'` that occurs when processing streaming chat completions.

## Details

When using the `chat_completions` endpoint with `stream=True`, the final `delta` chunk sent by the server may not contain a `content` key. This is part of the standard API behavior to signal the end of the stream.

The existing code in `_extract_completions_delta_content` did not account for this possibility and tried to access `delta['content']` directly, leading to a `KeyError` and causing the benchmark process to crash when the stream ended.

## Test Plan

This was discovered while running `guidellm benchmark` against an OpenAI-compatible API endpoint (via `litellm`) that correctly implements the streaming protocol.

guidellm benchmark \
  --target "http://10.64.1.62:4000/v1" \
  --model "qwen3-06b-2" \
  --processor "Qwen/Qwen3-0.6B" \
  --rate-type "synchronous" \
  --max-requests 1 \
  --data "prompt_tokens=32,output_tokens=32,samples=1"

## Related Issues

https://github.com/vllm-project/guidellm/issues/315

- Resolves #

---

## Use of AI

### DIFF
--- a/src/guidellm/backend/openai.py
+++ b/src/guidellm/backend/openai.py
@@ -688,7 +688,7 @@ class OpenAIHTTPBackend(Backend):
             return data["choices"][0]["text"]
 
         if type_ == "chat_completions":
-            return data["choices"][0]["delta"]["content"]
+            return data.get("choices", [{}])[0].get("delta", {}).get("content")
 
         raise ValueError(f"Unsupported type: {type_}")
 


sjmonson (dismissed):
Looks good to me but needs signoff.

sjmonson (approved):

PR #328, opened 2025-09-15
first benchark testing example

tukwila:
## Summary

image

## Details

- [ ]

## Test Plan

-

## Related Issues

- Resolves #

---

## Use of AI


docs/examples/practice_on_vllm_simulator.md
```diff
@@ -0,0 +1,112 @@
...
+    "model": "tweet-summary-0",
+    "stream": false,
+    "prompt": "Say this is a test!",
+    "max_tokens": 128
+}'
+```
+
+#### 1.5 Download Tokenizer
+
+Download Qwen/Qwen3-0.6B tokenizer.json from [Qwen/Qwen3-0.6B](https://modelscope.cn/models/Qwen/Qwen3-0.6B/files) save to local path.
```
DaltheCow: For this step I couldn't get things working without downloading both tokenizer.json and tokenizer_config.json. With only one or the other it would fail. Something weird with AutoTokenizer from my understanding.

I was also a bit confused in the next step since it says download tokenizer.json, but for the setting the processor it should be:
`--processor "${local_path}/Qwen3-0.6B"`

I am less familiar with how to run tokenizers or manually set a processor as I've always just directly set the model to the hugging face id, even when I use the llm-d inference simulator. I think this shows my lack of familiarity, but I realized after a little while that the files should be downloaded into a Qwen3-0.6B directory that contains the two files. Not sure if it's just me but might be helpful to explicitly say the files should be downloaded into the Qwen3-0.6B directory even though it is sort of implied in the next step.

DaltheCow: Above Qwen2.5-1.5B-Instruct was used but here and below Qwen3-0.6B is used. Is that intentional? I would guess they could all be Qwen3-0.6B?

DaltheCow: Or is the tokenizer for Qwen3-0.6B okay to use with the Qwen2.5-1.5B-Instruct model? I don't know know a lot about this topic.

tukwila: yes, i correct tokenizer files.

tukwila: For the sake of clarity, i correct tokenizer files to Qwen2.5-1.5B-Instruct

tukwila:
@DaltheCow Is this docs ok?

markurtz (approved):
Looks good, thanks @tukwila!

PR #362, opened 2025-09-23
[GuideLLM Refactor] Scenarios reenablement

sjmonson:
## Details

- [ ]

## Test Plan

-

## Related Issues

- Resolves #

---

## Use of AI

### DIFF
--- a/src/guidellm/benchmark/scenario.py
+++ b/src/guidellm/benchmark/scenario.py
@@ -84,7 +84,7 @@ class Scenario(StandardBaseModel):
             logger.error(f"Failed to parse {filename} as type {cls.__name__}")
             raise ValueError(f"Error when parsing file: {filename}") from e
 
-        data.update(overrides)
+        data.update(overrides or {})
         return cls.model_validate(data)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


src/guidellm/benchmark/scenario.py
```diff
@@ -78,29 +115,76 @@ class Config:
...
+    ) = None
+    warmup: Annotated[float | None, Field(gt=0, le=1)] = None
+    cooldown: Annotated[float | None, Field(gt=0, le=1)] = None
+    request_samples: PositiveInt | None = 20
+    # Constraints configuration
+    max_seconds: PositiveFloat | PositiveInt | None = None
+    max_requests: PositiveInt | None = None
+    max_errors: PositiveInt | None = None
+    max_error_rate: PositiveFloat | None = None
+    max_global_error_rate: PositiveFloat | None = None
```
jaredoconnell, line 139: With a lot of these inputs shared with other functions, it would reduce redundancy and reduce the risk of fragmentation and mismatched types if we create type aliases for the more complicated types here, and share them with the entrypoints file and other relevant areas. I created some type aliases in my PR that was merged yesterday. Feel free to relocate those if you find a better spot for them.

jaredoconnell (approved):
Looks good to me. With the fix I pushed it works for me, though I haven't had the time to let a full run complete.
I have one comment, but I won't block on that. I am also willing to do that work for you if you like the idea.

src/guidellm/benchmark/scenario.py
sjmonson: Yeah think it would be good to reuse the type aliases from your PR. If you have some cycles feel free to take over this PR.

jaredoconnell (approved):

PR #366, opened 2025-09-24
[GuideLLM Refactor] Fix from-file

jaredoconnell:
## Summary

This PR ports the new functionality from `benchmark run` to `benchmark from-file`, and does so in a way that reuses as much code as practical to have one source of truth.

## Details

- Fixes from-file by making it to use the new output format.
- Moves code related to the new output formats to separate functions that are called from both benchmark entrypoints.
- Moves additional chunks of code out of the large benchmark run entrypoint function for modularity.

## Test Plan

Run a benchmark with an output of json or yaml, and use `from-file` to re-import it and export it. You can select any output type supported by `benchmark run`.

`guidellm benchmark from-file ./result.json --output-formats console`
`guidellm benchmark from-file ./result.yaml --output-formats yaml`

## Related Issues

---

## Use of AI

### DIFF
--- a/src/guidellm/benchmark/entrypoints.py
+++ b/src/guidellm/benchmark/entrypoints.py
@@ -41,7 +41,7 @@ from guidellm.scheduler import (
     NonDistributedEnvironment,
     StrategyType,
 )
-from guidellm.utils import Console, InfoMixin
+from guidellm.utils import Console, InfoMixin, ConsoleUpdateStep
 
 __all__ = [
     "benchmark_generative_text",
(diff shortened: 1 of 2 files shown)


src/guidellm/benchmark/entrypoints.py
```diff
@@ -52,6 +52,90 @@
...
+async def finalize_outputs(
+    report: GenerativeBenchmarksReport,
+    resolved_output_formats: dict[str, GenerativeBenchmarkerOutput]
+):
+    output_format_results = {}
+    for key, output in resolved_output_formats.items():
+        output_result = await output.finalize(report)
+        output_format_results[key] = output_result
+    return output_format_results
+
```
sjmonson, line 142: Lets move all of this to a new file (`initializers.py` maybe); `entrypoints.py` is kind of our public interface for ABI usage so I think its best only have the front-facing functions in here. Also allows for some type reuse in `scenarios.py`

sjmonson (approved):
Above comment is non-blocking

sjmonson (approved):

src/guidellm/benchmark/entrypoints.py
jaredoconnell: To bring the conversation online: I thought so too, but it appears that Mark would prefer them to be in the same file. To clarify, I added comments.

PR #374, opened 2025-09-26
[POST GuideLLM Refactor] Multi-Turn Rework

sjmonson:
## Details

- [ ]

## Test Plan

-

## Related Issues

- Resolves #

---

## Use of AI


src/guidellm/scheduler/objects.py
```diff
@@ -71,6 +82,21 @@ class SchedulerMessagingPydanticRegistry(RegistryMixin[RegistryObjT]):
...
+
+    post_requeue_delay: float = Field(
+        description=(
+            "Delay in seconds to wait after a request to "
+            "queue the next request in the conversation."
+        ),
+        default=0.0,
+    )
+
+
```
sjmonson: This could be a part of `ScheduledRequestInfo` but I figured it was not really necessary to pass it back and forth with each request update. Plus I thought this would be a good interface for adjusting the scheduling of individual requests.

src/guidellm/backends/openai.py
```diff
@@ -500,6 +499,22 @@ async def chat_completions(
...
+        Apply conversation history to the current request.
+        """
+
+        def turn_to_text(turn: tuple[GenerationRequest, GenerationResponse]) -> str:
+            req, res = turn
+            return f"{req.content}{res.value}"
+
+        request.content = "".join(chain(map(turn_to_text, history), (request.content,)))
+        return request
+
```
sjmonson: Temporary hack until we land request templates.

PR #376, opened 2025-09-29
[GuideLLM Refactor] Edge case errors

jaredoconnell:
## Summary

This PR handles errors that occur when there are no successful requests. There will obviously still be an error, but it will be one that the user can get useful information from, rather than one that is the inner workings breaking.

## Details

- Adds default value for an inner data type to allow it to work in this edge case.
- Adds an error check that creates a runtime error with an explanation for the failure. The error message can be changed if you would like the wording changed.
- Fixes a type literal mismatch.

## Test Plan

- Run GuideLLM against a mock server in a way that results in all requests failing. Like setting the max token value way too small.

---

## Use of AI

### DIFF
--- a/src/guidellm/benchmark/profile.py
+++ b/src/guidellm/benchmark/profile.py
@@ -681,7 +681,7 @@ class SweepProfile(Profile):
                 prev_benchmark.metrics.requests_per_second.successful.mean
             )
             if self.synchronous_rate <= 0 and self.throughput_rate <= 0:
-                raise RuntimeError("Invalid rates in sweep; aborting. Where there any successful requests?")
+                raise RuntimeError("Invalid rates in sweep; aborting. Were there any successful requests?")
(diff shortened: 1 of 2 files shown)


src/guidellm/benchmark/profile.py
```diff
@@ -680,6 +680,8 @@ def next_strategy(
             self.throughput_rate = (
                 prev_benchmark.metrics.requests_per_second.successful.mean
             )
+            if self.synchronous_rate <= 0 and self.throughput_rate <= 0:
+                raise RuntimeError("Invalid rates in sweep; aborting. Where there any successful requests?")
```
sjmonson:
```suggestion
                raise RuntimeError("Invalid rates in sweep; aborting. Were there any successful requests?")
```

sjmonson (approved):

PR #377, opened 2025-09-29
[GuideLLM Refactor] Features/refactor fix html

jaredoconnell:
## Summary

Fixes HTML output in the refactor branch.

## Details

This PR updates the references for the new data, and corrects a problem that prevented the injection of the data into the HTML. They broke in the refactor.

Here is a screenshot of the current output using this branch:
image

There will be a follow up PR to make HTML and CSV outputs default in the CLI.

## Test Plan

To test this you need to locally host the web server as defined in the readme, then run the following command, with variables replaced for your setup:
`guidellm benchmark run --target <your target> --data "prompt_tokens=256,output_tokens=128" --rate-type constant --rate 2 --max-seconds 10 --output-path <your output path> --output-formats html`

---

## Use of AI

### DIFF
--- a/src/guidellm/benchmark/output.py
+++ b/src/guidellm/benchmark/output.py
@@ -724,9 +724,9 @@ class GenerativeBenchmarkerHTML(GenerativeBenchmarkerOutput):
 
         ui_api_data = {}
         for key, value in camel_data.items():
-            placeholder_key = f"window.{humps.decamelize(key)} = {{}};"
+            placeholder_key = f"window.{key} = {{}};"
             replacement_value = (
-                f"window.{humps.decamelize(key)} = {json.dumps(value, indent=2)};\n"
+                f"window.{key} = {json.dumps(value, indent=2)};\n"
             )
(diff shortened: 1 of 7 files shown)


src/guidellm/benchmark/output.py
```diff
@@ -36,6 +37,8 @@
     safe_format_timestamp,
     split_text_list_by_length,
 )
+from guidellm.utils.dict import recursive_key_update
+from guidellm.utils.text import camelize_str
```
sjmonson:
```suggestion
from guidellm.utils import recursive_key_update, camelize_str
```

When importing from another submodule, import the top level

sjmonson (changes requested):
Small fix; otherwise looks good

src/guidellm/benchmark/output.py
jaredoconnell: Sounds good. I just included the original code from main, but this is more compact.

sjmonson (approved):

Issue #369, opened 2025-09-25
Make base server endpoint configurable


sjmonson:
**Is your feature request related to a problem? Please describe.**
Sometimes OpenAI-compatible endpoints are hosted on differing paths from `/v1/...` (e.g. `/v3/...`, `/v1beta/...`)

**Describe the solution you'd like**
Support a config option for configuring the base endpoint. Something like `GUIDELLM__OPENAI__BASE_ENDPOINT=/v1`.

PR #384, opened 2025-10-01
Initial state for datasets rework to enable multimodal and more compl…

markurtz:
…icated combinations

## Details

- [ ]

## Test Plan

-

## Related Issues

- Resolves #

---

## Use of AI


src/guidellm/data/objects.py
```diff
@@ -47,66 +46,23 @@ def model_combine_dict(  # noqa: C901, PLR0912
...
+            )
+
+            if (json_body := args_dict.get("json_body")) is not None:
+                combined["json_body"] = combined.get("json_body", {}) + json_body
+            if (files := args_dict.get("files")) is not None:
+                combined["files"] = combined.get("files", {}) + files
+            if (params := args_dict.get("params")) is not None:
+                combined["params"] = combined.get("params", {}) + params
+            if (headers := args_dict.get("headers")) is not None:
+                combined["headers"] = combined.get("headers", {}) + headers
```
jaredoconnell: This could could be simplified with loops, if desired.

src/guidellm/backends/openai.py
```diff
@@ -10,36 +10,53 @@
...
+
+
+open_ai_paths: dict[str, str] = {
+    "health": "health",
+    "models": "v1/models",
+    "text_completions": "v1/completions",
+    "chat_completions": "v1/chat/completions",
+    "audio_transcriptions": "v1/audio/transcriptions",
+    "audio_translations": "v1/audio/translations",
+}
```
sjmonson: Eventually we should make the `v1` part of the endpoint separately configurable (see #369) but that can be a followup.

src/guidellm/backends/openai.py
```diff
@@ -300,350 +235,207 @@ async def resolve(
...
+                            prompt_stats, output_stats = self._extract_response_stats(
+                                data, request
+                            )
+                        else:
+                            if request_info.request_timings.first_iteration is None:
+                                request_info.request_timings.first_iteration = (
+                                    time.time()
+                                )
+                            request_info.request_timings.last_iteration = time.time()
+                            deltas.append(self._extract_response_text(data))
```
sjmonson: Depending on the path taken, time.time is called at different points in the processing.

Call time.time at the beginning of each loop and assign it as appropriate to `request_info.request_timings.request_end`, `request_info.request_timings.first_iteration` and/or `request_info.request_timings.last_iteration`.

src/guidellm/backends/openai.py
```diff
@@ -300,350 +235,207 @@ async def resolve(
...
-                {
-                    "max_tokens": max_tokens,
-                    "max_completion_tokens": max_tokens,
-                }
-            )
-            # Set stop conditions only for request-level limits
-            if max_output_tokens:
-                body.update({"stop": None, "ignore_eos": True})
+        if "delta" in data:
+            return data.get("delta", "")
```
sjmonson: After we rebase this lets make it a `match` statement. I checked the bytecode and we avoid a few load ops.

src/guidellm/backends/response_handlers.py
```diff
@@ -0,0 +1,283 @@
...
+import json
+from typing import Any, Protocol, cast
+
+from guidellm.schemas import GenerationRequest, GenerationResponse, UsageMetrics
+from guidellm.utils import RegistryMixin
+
+try:
+    import orjson
+except ImportError:
+    orjson = None
```
sjmonson: It seems like `orjson` is always used in place of `json` so maybe to save us some run time checks:


```suggestion
from typing import Any, Protocol, cast

from guidellm.schemas import GenerationRequest, GenerationResponse, UsageMetrics
from guidellm.utils import RegistryMixin

try:
    import orjson as json
except ImportError:
    import json
```

Might make sense to do a global helper for this too.

src/guidellm/data/objects.py
markurtz: Reworked and simplified

src/guidellm/backends/openai.py
markurtz: Currently updated it so that the open_ai_paths themselves can be passed into the backend and configured, so the user has an option to control the base url and the paths. Additionally, they can override from the request formatting. That will enable the most flexibility IMO. I'm wary of enabling smaller parts within, like specifying v1 or not, because I feel like that can get bloated and confusing pretty quickly

src/guidellm/backends/openai.py
markurtz: Updated to grab time.time at the beginning

src/guidellm/backends/openai.py
markurtz: This has been stripped out now and replaced with the response handler pathways which should limit the number of checks on each loop

src/guidellm/backends/response_handlers.py
```diff
@@ -0,0 +1,466 @@
...
+
+        if not line or not (line := line.strip()) or not line.startswith("data:"):
+            return 0
+
+        line = line[len("data:") :].strip()
+        data = cast(
+            "dict[str, Any]",
+            json.loads(line) if orjson is None else orjson.loads(line),
+        )
+        updated = False
```
sjmonson: This bit should probably be in a shared function since its identical for every streaming endpoint.

src/guidellm/backends/response_handlers.py
```diff
@@ -0,0 +1,466 @@
...
+        self, request: GenerationRequest, response: dict
+    ) -> GenerationResponse:
+        """
+        Process a complete text completion response.
+
+        :param request: The original generation request
+        :param response: Complete API response containing choices and usage data
+        :return: Standardized GenerationResponse with extracted text and metrics
+        """
+        choices = cast("list[dict]", response.get("choices", []))
```
sjmonson: Something to note: `cast` has runtime overhead ([ref](https://github.com/python/cpython/issues/137138)). I adapted the script in the reference to our use-case (replaced `Union` with `|` and specified types as strings) and saw about an 80% loss in performance:

```
Baseline (direct assignment): 0.0166s (1.00x)
Simple cast: 0.0302s (1.82x)
String cast: 0.0303s (1.82x)
Complex string cast: 0.0295s (1.78x)
```

src/guidellm/benchmark/entrypoints.py
```diff
@@ -44,108 +36,300 @@
...
+    | None,
+)
 
-# Helper functions
+ProcessorInputT = TypeAliasType("ProcessorInputT", str | Path | PreTrainedTokenizerBase)
 
+ProgressInputT = TypeAliasType(
+    "ProgressInputT", tuple[str, ...] | list[str] | list[BenchmarkerProgress]
+)
+
```
sjmonson: Some of these are duplicated in `types.py`

src/guidellm/data/deserializers/synthetic.py
```diff
@@ -0,0 +1,344 @@
...
+    )
+    source: str = Field(
+        description="The source of the text data to be used for generation.",
+        default="data:prideandprejudice.txt.gz",
+    )
+
+    @model_validator(mode="after")
+    def check_prefix_options(self) -> SyntheticTextDatasetConfig:
+        prefix_count = self.__pydantic_extra__.get("prefix_count", None)  # type: ignore[attr-defined]
+        prefix_tokens = self.__pydantic_extra__.get("prefix_count", None)  # type: ignore[attr-defined]
```
sjmonson: This one is my fault.


```suggestion
        prefix_count = self.__pydantic_extra__.get("prefix_count", None)  # type: ignore[attr-defined]
        prefix_tokens = self.__pydantic_extra__.get("prefix_tokens", None)  # type: ignore[attr-defined]
```

src/guidellm/data/preprocessors/formatters.py
```diff
@@ -0,0 +1,318 @@
...
+        ):
+            input_metrics.text_tokens = prompt_tokens
+
+        # Apply extra arguments
+        if self.extras:
+            arguments.model_combine(self.extras)
+
+        # Build prompt
+        prefix = "".join(pre for pre in columns.get("prefix_column", []) if pre)
+        text = "".join(txt for txt in columns.get("text_column", []) if txt)
```
sjmonson, line 79: Why are these lists? In the Jinja2 version of this patch it seemed to be for batch processing, but that is not how its used here.

src/guidellm/data/preprocessors/mappers.py
```diff
@@ -0,0 +1,191 @@
...
+                if "." in name:
+                    dataset, column_name = name.split(".", 1)
+                    dataset_index = (
+                        int(dataset)
+                        if dataset.isdigit()
+                        else datasets_named_indices.get(dataset)
+                    )
+                else:
+                    dataset_index = 0
+                    column_name = name
```
sjmonson, line 143: Can you add a comment explaining this?

src/guidellm/preprocess/dataset.py
```diff
@@ -271,9 +270,7 @@ def process_dataset(
         f"Starting dataset conversion | Input: {data} | Output directory: {output_path}"
     )
 
-    dataset, column_mappings = guidellm_load_dataset(
-        data, data_args, processor, processor_args
-    )
+    dataset, column_mappings = None, None
```
sjmonson, line 273: I'm guessing this code is non-functional?

src/guidellm/scheduler/worker.py
```diff
@@ -233,19 +232,18 @@ async def _processing_startup(self):
...
             self.startup_barrier,
             poll_interval=self.messaging.poll_interval,
         )
 
+        # Get messaging system ready
+        await self.messaging.start(
+            receive_stop_criteria=[self.requests_generated_event]
+        )
+        self.messaging_started = True
+
```
sjmonson, line 240: Why move this after the start barrier? Would it be better to utilize the barrier twice, once before messaging and once after?

src/guidellm/scheduler/worker_group.py
```diff
@@ -153,19 +154,23 @@ async def create_processes(self):
...
-        if max_conc == math.inf:
-            # if concurrency not specified, use settings
+        max_conc: int
+        if (
+            requests_limit := min(
+                self.strategy.requests_limit or math.inf,
+                self.backend.requests_limit or math.inf,
+            )
+        ) != math.inf:
+            max_conc = requests_limit  # type: ignore[assignment]
```
sjmonson:
```suggestion
            max_conc = int(requests_limit)
```

src/guidellm/schemas/info.py
```diff
@@ -0,0 +1,159 @@
...
+
+        :param kwargs: Additional keyword arguments for model copying
+        :return: New RequestInfo instance with independent timing objects
+        """
+        return super().model_copy(
+            update={
+                "timings": self.timings.model_copy(),
+            },
+            deep=False,
+        )
```
sjmonson, line 159: Wouldn't this also work?

```suggestion
        return super().model_copy(
            deep=True,
        )
```

src/guidellm/utils/statistics.py
```diff
@@ -255,6 +255,7 @@ def from_values(
     def from_request_times(
```
sjmonson, line 255: NOTE: Verify that this passes the tests from #266

src/guidellm/settings.py
```diff
@@ -46,7 +46,7 @@ class LoggingSettings(BaseModel):
 
     disabled: bool = False
     clear_loggers: bool = True
-    console_log_level: str = "WARNING"
+    console_log_level: str = "DEBUG"
```
sjmonson: `INFO` rather than `DEBUG`?

src/guidellm/settings.py
```diff
@@ -145,7 +145,7 @@ class Settings(BaseSettings):
     mp_max_pending_buffer_percent: float = 0.5
     mp_max_worker_buffer_percent: float = 0.2
     max_concurrency: int = 512
-    max_worker_processes: int = 10
+    max_worker_processes: int = 2
```
sjmonson: This seems too low, was it an accidental inclusion?

pyproject.toml
```diff
@@ -54,28 +61,26 @@ dependencies = [
     "httpx[http2]<1.0.0",
     "loguru",
     "msgpack",
-    "numpy",
+    "numpy<2.0.0",
```
sjmonson, line 64: What do we depend on that changed in `numpy` 2.0? Could we instead lock `numpy>=2.0.0,<2.3.0` as that range supports python 3.10.

pyproject.toml
```diff
@@ -54,28 +61,26 @@ dependencies = [
...
-    "msgspec",
-    "uvloop",
-]
+perf = ["orjson", "msgpack", "msgspec", "uvloop"]
 recommended = [
-    "tiktoken>=0.11.0",  # For OpenAI tokenizer
-    "blobfile>=3.1.0",   # For OpenAI tokenizer
+    "tiktoken>=0.11.0", # For OpenAI tokenizer
+    "blobfile>=3.1.0",  # For OpenAI tokenizer
 ]
```
sjmonson: Just a thought, lmk what you think. The name "recommended" was chosen because its part of an PEP that will eventually be supported in `pip` ([ref](https://peps.python.org/pep-0771)).

```suggestion
perf = ["orjson", "msgpack", "msgspec", "uvloop"]
openai = ["tiktoken>=0.11.0", "blobfile>=3.1.0"]
recommended = ["guidellm[perf,openai]"]
```

src/guidellm/backends/response_handlers.py
markurtz: Added a utility function

src/guidellm/backends/response_handlers.py
markurtz: extracted the core shared code to a helper across text and chat completions, audio pathways are slightly different since they don't have "data:" for the data returned

src/guidellm/backends/response_handlers.py
markurtz: Removed cast from the response_handlers code

src/guidellm/benchmark/entrypoints.py
markurtz: Imported from types

src/guidellm/data/preprocessors/formatters.py
markurtz: Lists in case there are multiple of any column specified across multiple datasets that are passed in with an assumed concatenation

src/guidellm/data/preprocessors/mappers.py
markurtz: added

src/guidellm/preprocess/dataset.py
markurtz: Ah yeah, thanks for pointing this out, I haven't migrated this yet. It's non functional because it relied on a helper function in the old data pathways, so need to rework it a bit

src/guidellm/scheduler/worker.py
markurtz: This was a left over from when I was debugging the race condition, moved back

src/guidellm/schemas/info.py
markurtz: It would, but I was trying to avoid doing a deep copy on the entire class for performance reasons

src/guidellm/settings.py
markurtz: Good catch, left over from debugging race condition

src/guidellm/settings.py
markurtz: Yep, left over from debugging, fixed

pyproject.toml
markurtz: The audio pathways through HF datasets and what it imports from torch require numy 1.X

pyproject.toml
markurtz: oh nice!

markurtz: It won't let me commit due to deleted lines, but fixed and pushing up

pyproject.toml
jaredoconnell: Is there a plan to migrate that?

src/guidellm/backends/openai.py
```diff
@@ -68,74 +51,54 @@ class OpenAIHTTPBackend(Backend):
...
+
+        :param target: Base URL of the OpenAI-compatible server
+        :param model: Model identifier for generation requests
+        :param api_routes: Custom API endpoint routes mapping
+        :param response_handlers: Custom response handlers for different request types
+        :param timeout: Request timeout in seconds
+        :param http2: Enable HTTP/2 protocol support
+        :param follow_redirects: Follow HTTP redirects automatically
+        :param verify: Enable SSL certificate verification
+        :param validate_backend: Backend validation configuration
```
jaredoconnell, line 77: Is this documented more completely anywhere? If so, documentation on the expected values would be helpful.

src/guidellm/backends/openai.py
```diff
@@ -281,363 +207,149 @@ async def default_model(self) -> str | None:
...
-                **request.params,
-            },
-            value="",
-            request_prompt_tokens=request.stats.get("prompt_tokens"),
-            request_output_tokens=request.constraints.get("output_tokens"),
-        )
-        request_info.request_timings = GenerationRequestTimings()
-        request_info.request_timings.request_start = time.time()
+        if history is not None:
+            raise NotImplementedError("Multi-turn requests  not yet supported")
```
jaredoconnell:
```suggestion
            raise NotImplementedError("Multi-turn requests not yet supported")
```

src/guidellm/backends/response_handlers.py
```diff
@@ -0,0 +1,456 @@
...
+            input_metrics=input_metrics,
+            output_metrics=output_metrics,
+        )
+
+    def extract_line_data(self, line: str) -> dict[str, Any] | None:
+        """
+        Extract JSON data from a streaming response line.
+
+        :param line: Raw line from the streaming response
+        :return: Parsed JSON data as a dictionary, or None if line is invalid
```
jaredoconnell, line 174: Is this correct? It looks like it returns None when it gets Done and it returns an empty dictionary if the line is invalid.

src/guidellm/benchmark/entrypoints.py
```diff
@@ -320,30 +424,27 @@ async def reimport_benchmarks_report(
 ) -> tuple[GenerativeBenchmarksReport, dict[str, Any]]:
     """
     The command-line entry point for re-importing and displaying an
-    existing benchmarks report. Can also specify an output format.
+    existing benchmarks report. Can also specify
```
jaredoconnell, line 415: This looks like a merge error that accidentally reverted a change.

src/guidellm/__main__.py
```diff
@@ -393,7 +423,45 @@ def benchmark():
     default=GenerativeTextScenario.get_default("max_global_error_rate"),
     help="Maximum global error rate allowed across all benchmarks",
 )
-def run(**kwargs):
```
jaredoconnell, line 396: This looks to revert the changes from #362. Is this intentional?

markurtz: This piece is intentional for now until we can reenable scenarios. Because the args changed, it's going to be more involved to reenable that

src/guidellm/backends/openai.py
markurtz: Good callout and we need to expand out documentation quite a bit, I'll make sure we address that in a follow up PR

src/guidellm/backends/response_handlers.py
markurtz: Updated docs

src/guidellm/benchmark/entrypoints.py
markurtz: Yep, fixed

PR #391, opened 2025-10-07
Features/add model size to the UI

DaltheCow:
## Summary

The run info will now grab the model size if the model is pulled from HF

## Details

- Model size added to run info


src/guidellm/presentation/data_models.py
```diff
@@ -71,11 +109,14 @@ def from_benchmarks(cls, benchmarks: list["GenerativeBenchmark"]):
         timestamp = max(
             bm.run_stats.start_time for bm in benchmarks if bm.start_time is not None
         )
+        response = httpx.get(f"https://huggingface.co/api/models/{model}")
+        model_json = response.json()
```
sjmonson: Is this safe from an error response?

sjmonson (changes requested):
One question, just resolve it and I'll approve

src/guidellm/presentation/data_models.py
DaltheCow: I'm splitting time between this team and another team now but I will get to this. Definitely need to handle an error response

sjmonson:
Stale

PR #392, opened 2025-10-07
add dots with tooltip to line chart

### LINKED ISSUE #277: Label points in UI with what benckmark they represent
The summary graph in the UI plots each benchmarks Responses Per Second against various latency metrics and interpolates values between the points. Because there are no points on the graph it is hard to understand where the measured values are.

Add dots at each measured value and a tooltip/label matching the benchmark to a row in the CLI summary. (e.g. `constant@0.5`)

DaltheCow:
## Summary

The Metrics summary section now includes the data point dots on the line and a tooltip including the strategy label, e.g.: synchronous, constant@32.32, concurrent@5

## Test Plan

- Unit test for backend change
- No tests written to check on tooltip addition, seems overflow complex to try and hover the dots on the chart but I didn't look into it.

## Related Issues

- Resolves #277 

---

- [x] "I certify that all code in this PR is my own."

### DIFF
--- a/src/guidellm/presentation/data_models.py
+++ b/src/guidellm/presentation/data_models.py
@@ -5,6 +5,8 @@ from typing import TYPE_CHECKING, Optional, Union
 
 from pydantic import BaseModel, computed_field
 
+from guidellm.scheduler.strategy import SchedulingStrategy
+
 if TYPE_CHECKING:
     from guidellm.benchmark import GenerativeBenchmark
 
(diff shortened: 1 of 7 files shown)


ivanbaldo:
Maybe it would be useful to add also the actual value?

For example "constant@26.81 1454.62 ms" (in the TTFT graph), "constant@26.81 204.35 tok/s" (in the Throughput graph), etc..

PR #393, opened 2025-10-08
[GuideLLM Refactor] fix util package types

jaredoconnell:
## Summary

This PR fixes all type errors in the utils package. Only a few were ignored.

## Details

- A lot of these changes are reflecting that values can be None, and the associated None checks.
- Others are incorrect type annotations
- Others are asserting with cast that we know for certain that the type is correct.
- Plus other minor changes

## Test Plan

Run the tests and look through the changes to make sure the logic is equivalent or better to the original code.

---

## Use of AI


src/guidellm/utils/messaging.py
```diff
@@ -216,9 +218,8 @@ async def start(
         self.buffer_receive_queue = culsans.Queue[ReceiveMessageT](
             maxsize=self.max_buffer_receive_size or 0
         )
-        self.tasks_lock = threading.Lock()
```
jaredoconnell, line 219: This was entirely unused. Is that a problem?

src/guidellm/utils/encoding.py
```diff
@@ -296,6 +302,15 @@ def _get_available_encoder_decoder(
         return None, None, None
 
 
+PayloadType = Literal[
+    "pydantic",
+    "python",
+    "collection_tuple",
+    "collection_sequence",
+    "collection_mapping",
+]
```
sjmonson, line 311: This needs to be a `typing_extensions.TypeAliasType` to be backwards compatible with python<3.12.

src/guidellm/utils/messaging.py
```diff
@@ -48,19 +48,21 @@
 ReceiveMessageT = TypeVar("ReceiveMessageT", bound=Any)
 """Generic type variable for messages received through the messaging system"""
 
+CheckStopCallableT = Callable[[bool, int], bool]
```
sjmonson, line 51: `typing_extensions.TypeAliasType`

src/guidellm/utils/messaging.py
sjmonson: Yeah I am guessing this is leftover from a previous version. cc: @markurtz

src/guidellm/utils/messaging.py
```diff
@@ -506,9 +542,9 @@ def create_worker_copy(
             "pending_queue": self.pending_queue,
             "done_queue": self.done_queue,
         }
-        copy_args.update(kwargs)
+        final_args = {**copy_args, **kwargs}
```
sjmonson, line 545: I haven't reviewed the purpose of `kwargs` here, but should this be

```suggestion
        final_args = copy_args | kwargs
```

to prevent failing if both dicts have the same key?

src/guidellm/utils/messaging.py
```diff
@@ -847,9 +918,10 @@ def create_worker_copy(
             "worker_index": worker_index,
             "pipe": self.pipes[worker_index],
         }
-        copy_args.update(kwargs)
 
-        return InterProcessMessagingPipe(**copy_args)
+        final_args = {**copy_args, **kwargs}
```
sjmonson, line 922: Likewise to prev kwargs comment.

src/guidellm/utils/registry.py
```diff
@@ -209,6 +211,8 @@ def get_registered_object(cls, name: str) -> RegistryObjT | None:
         if name in cls.registry:
             return cls.registry[name]
 
-        lower_key_map = {key.lower(): key for key in cls.registry}
+        for k, v in cls.registry.items():
+            if name.lower() == k.lower():
+                return v
```
sjmonson, line 217: Avoid calling lower more than we need to:

```suggestion
        name_casefold = name.lower()
        for k, v in cls.registry.items():
            if name_casefold == k.lower():
                return v
```

src/guidellm/utils/messaging.py
jaredoconnell: This shouldn't fail if both dicts have the same key. It should use the values from `kwargs` in the event of a duplicate key.

The method used here allows the type checker to remember which types are passed in, as opposed to making it unknown with `.update`

src/guidellm/utils/registry.py
jaredoconnell: Looks good. This will definitely be more efficient than the original code.

src/guidellm/utils/messaging.py
sjmonson: Oh your right. I've only ever tried this with function calls and it fails doing that with duplicate keys.

src/guidellm/utils/encoding.py
sjmonson: Wontdo, dropping 3.9 makes this unnecessary.

PR #394, opened 2025-10-08
[GuideLLM Refactor] Fixes for various failing unit tests

sjmonson:
## Summary

Fixes failing unit tests. Most were failing due to changes in functionality but a couple were regressions.

---

## Use of AI

### DIFF
--- a/src/guidellm/benchmark/types.py
+++ b/src/guidellm/benchmark/types.py
@@ -2,12 +2,13 @@ from __future__ import annotations
 
 from collections.abc import Iterable
 from pathlib import Path
-from typing import Any, TypeAliasType
+from typing import Any
 
 from datasets import Dataset, DatasetDict, IterableDataset, IterableDatasetDict
 from transformers import (  # type: ignore[import]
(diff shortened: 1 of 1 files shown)


tests/unit/test_main.py
```diff
@@ -19,10 +19,10 @@ def test_benchmark_run_with_backend_args():
             "--backend-args",
             '{"headers": {"Authorization": "Bearer my-token"}, "verify": false}',
             "--target",
-            "http://localhost:8000",
```
jaredoconnell, line 22: Is there a mock server on port 9?

sjmonson: No, based on this comment I believe this test is intended to point to a closed port. Port 9 is a safer bet than 8000 because its designated as a closed port ([ref](https://en.wikipedia.org/wiki/Discard_Protocol)). https://github.com/vllm-project/guidellm/blob/448ed7a866e379f3f3b32b00b2e53d0a56713c0d/tests/unit/test_main.py#L33

PR #397, opened 2025-10-09
[GuideLLM Refactor] Bump minimum python version to 3.10

sjmonson:
## Summary

Bumps the minimum python version to 3.10 and updates all dependent tooling (CI, tox, ruff, mypy).

---

## Use of AI


scripts/generate_pylock.sh
sjmonson, line 1: This script is not strictly necessary now that we have dropped python 3.9 as the issue was mostly with numpy not having a single version with support for all of 3.9-3.13. But its still nice to split the lock at 3.12 to allow for newer locked packages in more supported python versions. Location for this file doesn't really matter it only needs to be run if the version ranges change otherwise pdm is smart enough to reuse existing ranges.

PR #411, opened 2025-10-14
[GuideLLM Refactor] Replace librosa, pydub, and soundfile with torchcodec

sjmonson:
## TODO

- [ ] ~~More flexible version locking in multimodal extras group~~
  - Goal with this was to add locking for different torchcodec/torch versions but honestly its not worth the hassle
- [x] Check for multi-modal libs being installed
- [ ] More testing on `encode_audio`

## Summary

Replaces audio processing libraries with `torchcodec` which eliminates 19 dependencies and brings us inline with what HuggingFace `datasets` is doing.

## Details

- 

## Test Plan

- Run against audio server with

```bash
guidellm benchmark run \
    --profile "synchronous" \
    --max-requests 20 \
    --request-type "audio_transcriptions" \
    --data "openslr/librispeech_asr" \
    --data-args '{"name": "clean", "split": "test"}'
```

---

## Use of AI

### DIFF
--- a/pylock.toml
+++ b/pylock.toml
@@ -13,36 +13,31 @@ created-by = "pdm"
 
 [[packages]]
 name = "torch"
-version = "2.8.0+cpu"
-requires-python = ">=3.9.0"
+version = "2.9.0+cpu"
+requires-python = ">=3.10"
 wheels = [
-    {name = "torch-2.8.0+cpu-cp313-cp313-linux_s390x.whl",url = "https://download.pytorch.org/whl/cpu/torch-2.8.0%2Bcpu-cp313-cp313-linux_s390x.whl",hashes = {sha256 = "8b5882276633cf91fe3d2d7246c743b94d44a7e660b27f1308007fdb1bb89f7d"}},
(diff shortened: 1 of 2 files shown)


src/guidellm/data/preprocessors/formatters.py
```diff
@@ -18,8 +19,28 @@
...
+
+        return encode_image(*args, **kwargs)
+
+    @staticmethod
+    def encode_video(*args, **kwargs):
+        from guidellm.extras.multimodal import encode_video
+
+        return encode_video(*args, **kwargs)
+
+
```
sjmonson, line 41: I thought this would be a good way to surface import error, but it turns out that its caught by the request error handler. Thinking of adding a check where if the first `n` requests all fail we just exit with the exception from the first request.

jaredoconnell (dismissed):
Wrong PR.

src/guidellm/data/preprocessors/formatters.py
jaredoconnell: Yeah that on the surface does look good. Is there a step that happens before that that we can do the import check?

jaredoconnell (approved):
Looks good to me. And the code works fine in my testing.

src/guidellm/data/preprocessors/formatters.py
sjmonson: Not really, we could create one. The problem is its hard to know if its necessary before this step since there is no user provided flag, its just based on whether the data has multi-modal columns which doesn't seem to be checked till here:

https://github.com/vllm-project/guidellm/blob/eede90057a3142a1e2c89f14859af49f314f5efa/src/guidellm/data/preprocessors/formatters.py#L270-L272

jaredoconnell: Hm. Making a step to validate requirements may make sense. But maybe not in this PR.

PR #412, opened 2025-10-15
[GuideLLM Refactor] Fixes for constant rate and other minor bugs for audio flows

markurtz:
## Summary

Rework scheduling strategies and how workers/worker group uses them to both simplify and ensure race conditions won't happen between worker processes and grabbing timings. Additionally, fix stats accumulation bugs and data loading bugs that were causing audio to fail

## Details

- [ ]

## Test Plan

-

## Related Issues

- Resolves #

---

## Use of AI

### DIFF
--- a/src/guidellm/data/deserializers/deserializer.py
+++ b/src/guidellm/data/deserializers/deserializer.py
@@ -77,7 +77,7 @@ class DatasetDeserializerFactory(
         if dataset is None:
             raise DataNotSupportedError(
                 f"No suitable deserializer found for data {data} "
-                f"with kwargs {data_kwargs} and type_ {type_}."
+                f"with kwargs {data_kwargs} and deserializer type {type_}."
             )
 
         if resolve_split:
--- a/src/guidellm/data/deserializers/huggingface.py
+++ b/src/guidellm/data/deserializers/huggingface.py
@@ -12,6 +12,11 @@ from datasets import (
     load_dataset,
     load_from_disk,
 )
+from datasets.exceptions import (
+    DataFilesNotFoundError,
+    DatasetNotFoundError,
+    FileNotFoundDatasetsError,
+)
 from transformers import PreTrainedTokenizerBase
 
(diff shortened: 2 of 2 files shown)


src/guidellm/scheduler/strategies.py
```diff
@@ -43,308 +41,162 @@
...
         """
-        Handle request completion (no action needed for constant rate strategy).
+        Initialize shared timing state for multi-process coordination.
 
-        :param request_info: Information about the completed request (unused)
+        :param worker_count: Number of worker processes to coordinate
+        :param max_concurrency: Maximum number of concurrent requests allowed
+        :param startup_duration: Duration in seconds for request startup ramping
         """
+        self.worker_coount = worker_count
```
jaredoconnell: That looks like a typo. This will need to be changed in several spots.

markurtz: fixed

src/guidellm/scheduler/worker_group.py
```diff
@@ -186,9 +172,7 @@ async def create_processes(self):
         max_pending_size = max(
             1, math.floor(max_conc * settings.mp_max_pending_buffer_percent)
         )
-        per_proc_max_buffer_size = max(
-            1, math.floor(per_proc_max_conc * settings.mp_max_worker_buffer_percent)
-        )
+        per_proc_max_buffer_size = 1
```
sjmonson, line 175: So buffer is just `len(self.processes)`?

sjmonson (approved):
Tiny bit worried about lock contention on `Strategy` objects but we can check that later. Otherwise LGTM with one clarifying question.

src/guidellm/scheduler/worker_group.py
markurtz: Ah yeah, meant to make this a configurable setting, will get up a separate diff for that. This is specifically for how much the local threading queues are able to store. We still have the much larger cache in the multiprocessing queue

PR #414, opened 2025-10-16
[GuideLLM Refactor] Reenablement of scenarios and fixes for benchmark package and CLI pathways

markurtz:
## Summary

Changed the benchmarking entrypoint to take in an Args object which is now used to load scenarios. It enables a single source of truth in addition to being able to save the exact configurations in the report output.

## Details

- [ ]

## Test Plan

-

## Related Issues

- Resolves #

---

## Use of AI

### DIFF
--- a/src/guidellm/benchmark/schemas.py
+++ b/src/guidellm/benchmark/schemas.py
@@ -1820,7 +1820,9 @@ class BenchmarkGenerativeTextArgs(StandardBaseModel):
     # Required
     target: str = Field(description="Target endpoint URL for benchmark execution")
     data: list[Any] = Field(
-        description="List of dataset sources or data files", default_factory=list
+        description="List of dataset sources or data files",
+        default_factory=list,
+        min_length=1,
     )
     # Benchmark configuration
     profile: StrategyType | ProfileType | Profile = Field(
--- a/src/guidellm/__main__.py
+++ b/src/guidellm/__main__.py
@@ -28,6 +28,7 @@ import codecs
 from pathlib import Path
 
 import click
+from pydantic import ValidationError
 
 try:
     import uvloop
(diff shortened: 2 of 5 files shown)


src/guidellm/benchmark/scenarios/__init__.py
sjmonson, line 1: For some reason the builtin scenarios are not getting installed with the package. `pip install -e .` works since it just links to the source, but `pip install .` does not (this might be my fault but I swear this used to work).

src/guidellm/benchmark/schemas.py
```diff
@@ -1404,6 +1736,263 @@ def compile(
...
+        :raises ValueError: If scenario is not found or file format is unsupported
+        """
+        constructor_kwargs = {}
+
+        if scenario is not None:
+            if isinstance(scenario, str) and scenario in (
+                builtin_scenarios := get_builtin_scenarios()
+            ):
+                scenario_path = builtin_scenarios[scenario]
+            elif Path(scenario).exists() and Path(scenario).is_dir():
```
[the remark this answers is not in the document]
sjmonson: Yes

sjmonson (changes requested):
Few issues, but otherwise looks good. Tested with a synchronous text run.

src/guidellm/benchmark/schemas.py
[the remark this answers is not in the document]
markurtz: Fixed and pushing up with others

src/guidellm/benchmark/scenarios/__init__.py
markurtz: Added the files to the package-data in setuptools so they should be included now

src/guidellm/benchmark/schemas.py
```diff
@@ -1404,6 +1736,263 @@ def compile(
         )
 
 
+class BenchmarkGenerativeTextArgs(StandardBaseModel):
```
jaredoconnell, line 1739: This is overdue. Nice to see a class for this so we don't need to pass around so many arguments everywhere.

PR #422, opened 2025-10-18
Add priority feature to Registry

jaredoconnell:
## Summary

This feature is designed for use cases like deserialize, where we need to use some objects before others.

As a side effect of these changes, I lowered the code's reliance on class variables, and instead on functions which can provide a more stable API.

There are multiple ways of implementing a priority system, and I'm open to feedback if this is not the direction we want to go in.

Fixes an issue where JSON inputs would fail at the `InMemoryJsonStrDatasetDeserializer`. With this new design that deserializer has a priority of 2, so that the other deserializers run first.
This change also allows us to remove the huggingface special case in the code, and instead set it as priority 3.

## Details

- Adds a priority field to the registration of object
- Added functions to retrieve objects from the registry
- Added function to retrieve sorted groups of objects by priority.
- Adds relevant tests

## Test Plan

To test that this addresses the original concern, run GuideLLM with all types of data to ensure the proper priorities are set on the objects.

The tests were updated to reflect these changes.

---

## Use of AI

### DIFF
--- a/src/guidellm/data/deserializers/deserializer.py
+++ b/src/guidellm/data/deserializers/deserializer.py
@@ -51,9 +51,7 @@ class DatasetDeserializerFactory(
 
         if type_ is None:
             for priority, objects in cls.get_priority_grouped_objects() :
-                print("Deserializing priorities:", priority)
                 for deserializer in objects:
-                    print("Deserializing deserializer:", deserializer.__name__)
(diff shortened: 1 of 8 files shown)


jaredoconnell:
We're going to try a solution that doesn't depend on proper ordering.

PR #423, opened 2025-10-20
Try all data deserializers before failing

jaredoconnell:
## Summary

This pull request makes the code try all dataset deserializers before failing. This ensures that the data format not working with one deserializer doesn't cause premature failure.

## Details

It collects all errors, and if none of them succeed in deserializing a dataset it either prints the errors, if present, or it reaches the old code which prints an error stating that there are no suitable deserializers for the data.

Here is an example of the error message raised if I force it to fail:
`
guidellm.data.deserializers.deserializer.DataNotSupportedError: data deserialization failed; 2 errors occurred while attempting to deserialize data {"prompt_tokens": 1, "output_tokens": 100}: [HFValidationError('Repo id must use alphanumeric chars or \'-\', \'_\', \'.\', \'--\' and \'..\' are forbidden, \'-\' and \'.\' cannot start or end the name, max length is 96: \'{"prompt_tokens": 1, "output_tokens": 100}\'.'), TypeError('InMemoryDictDatasetDeserializer.__call__() takes 4 positional arguments but 5 were given')]
`

## Test Plan

Run GuideLLM with various formats of data to ensure the proper one is used, and test invalid inputs, too.

---

## Use of AI

### DIFF
--- a/src/guidellm/data/deserializers/deserializer.py
+++ b/src/guidellm/data/deserializers/deserializer.py
@@ -50,31 +50,32 @@ class DatasetDeserializerFactory(
         dataset = None
 
         if type_ is None:
+            errors = []
+            # Note: There is no priority order for the deserializers, so all deserializers
+            #  must be mutually exclusive to ensure deterministic behavior.
             for name, deserializer in cls.registry.items():
-                if name == "huggingface":
-                    # Save Hugging Face til the end since it is a catch-all.
-                    continue
-
                 deserializer_fn: DatasetDeserializer = (
                     deserializer() if isinstance(deserializer, type) else deserializer
                 )
 
-                with contextlib.suppress(DataNotSupportedError):
-                    dataset = deserializer_fn(
-                        data=data,
-                        processor_factory=processor_factory,
-                        random_seed=random_seed,
-                        **data_kwargs,
-                    )
+                try:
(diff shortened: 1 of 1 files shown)


sjmonson (approved):
Code looks good, but can you clarify two thing:
1. Does this solve the error you linked in the description? As in `{"prompt_tokens": 1, "output_tokens": 100}` is parsed as a synthetic dataset.
2. Can you quantify the delay to startup time caused by running though all of the options?

jaredoconnell:
> Code looks good, but can you clarify two thing:
> 
>     1. Does this solve the error you linked in the description? As in `{"prompt_tokens": 1, "output_tokens": 100}` is parsed as a synthetic dataset.
> 
>     2. Can you quantify the delay to startup time caused by running though all of the options?

1. Yes, it does resolve that error.
2. Good idea.

If it goes through every dataset, it normally around 0.5 seconds. The old code also went through every dataset. The new one can at least short-circuit if it finds it.

Looks like it takes around the order of 10<sup>-6</sup> seconds to complete each of the unsuccessful deserializations, so small enough that the measurement itself is probably too inaccurate to measure it.
But the call to synthetic data gen for the synthetic_text deserializer can take some time. Once it took 10 seconds. I'm assuming that's from the generation of synthetic data.

PR #428, opened 2025-10-21
Implement incremental rate type

albertoperdomo2:
## Summary

This PR adds a new rate type `incremental` to the suite of benchmarks. This new rate type starts at a specified initial rate `--start-rate` and linearly increases the request rate over time by `--increment-factor`, offering an optional rate limit `--rate-limit` to cap the maximum rate. This simulates, as the name implies, load ramp up along time. 

Reimplementation of [#291](https://github.com/vllm-project/guidellm/pull/291) after the new release. 

## Details

- [x] Added new CLI flags mentioned before: `--start-rate`, `--increment-factor` and `--rate-limit`. 
- [x] Added a new profile `IncrementalProfile` class. 
- [x] Implemented `AsyncIncrementalStrategy` scheduler strategy that handles the logic and includes the optional initial burst. 
 
## Test Plan

-

## Related Issues

- Resolves #

---

## Use of AI

### DIFF
--- a/src/guidellm/benchmark/schemas/generative/entrypoints.py
+++ b/src/guidellm/benchmark/schemas/generative/entrypoints.py
@@ -172,6 +172,9 @@ class BenchmarkGenerativeTextArgs(StandardBaseModel):
     rate: list[float] | None = Field(
         default=None, description="Request rate(s) for rate-based scheduling"
     )
+    profile_kwargs: dict[str, Any] | None = Field(
+        default=None, description="Additional profile-specific configuration arguments"
+    )
     # Backend configuration
     backend: BackendType | Backend = Field(
         default="openai_http", description="Backend type or instance for execution"


src/guidellm/__main__.py
sjmonson, line 1: Generally we don't like to add top-level arguments that only map to a specific use-case. Many of the other components have a "kwargs" argument so it might make sense to add a `--profile-kwargs`. Also `--rate` needs to map to something so maybe make it `increment_factor` (or `start_rate`; whichever it makes more sense to sweep over).

src/guidellm/scheduler/strategies.py
```diff
@@ -517,3 +521,114 @@ def request_completed(self, request_info: RequestInfo):
...
+        elapsed_time = current_time - start_time
+        next_rate = self.start_rate + (self.increment_factor * elapsed_time)
+
+        # Cap at rate limit if specified
+        if self.rate_limit and next_rate >= self.rate_limit:
+            increment = 1.0 / self.rate_limit
+        else:
+            increment = 1.0 / next_rate
+
+        self._process_offset += increment
```
sjmonson, line 734: Strategies are shared across threads/processes, any assignments need to be atomic or locked with `self._processes_lock`. See other Strategies for examples.

src/guidellm/__main__.py
sjmonson: Another option might be to modify `--rate` to take tuples of `(start_rate, increment_factor)`.

cc: @markurtz @jaredoconnell to weigh in on this.

PR #430, opened 2025-10-21
[Refactor Followup] Fix data package type and quality errors

jaredoconnell:
## Summary

This PR fixes type and quality errors in the data package.

This brings the total type error count down to 124.

Some assumptions had to be made, so it would be a good idea to look for accidental changes in logic during your reviews.

## Testing

Run benchmarks with various data input types to ensure they all work.

---

## Use of AI

### DIFF
--- a/src/guidellm/data/preprocessors/formatters.py
+++ b/src/guidellm/data/preprocessors/formatters.py
@@ -103,8 +103,7 @@ class GenerativeTextCompletionsRequestFormatter(RequestFormatter):
         if prefix or text:
             prompt = prefix + text
             arguments.body["prompt"] = prompt
-            input_metrics.text_characters = len(prompt)
-            input_metrics.text_words = len(prompt.split())
+            input_metrics.add_text_metrics(prompt)
 
         return GenerationRequest(
             request_type="text_completions",
@@ -197,25 +196,14 @@ class GenerativeChatCompletionsRequestFormatter(RequestFormatter):
             if not prefix:
                 continue
 
-            input_metrics.text_characters = (
-                input_metrics.text_characters or 0
-            ) + len(prefix)
-
-            input_metrics.text_words = (input_metrics.text_words or 0) + \
-                                       len(prefix.split())
-
+            input_metrics.add_text_metrics(prefix)
             arguments.body["messages"].append({"role": "system", "content": prefix})
 
         for text in columns.get("text_column", []):
             if not text:
                 continue
 
-            input_metrics.text_characters = (
-                input_metrics.text_characters or 0
-            ) + len(text)
-            input_metrics.text_words = (
-               input_metrics.text_words or 0
-            ) + len(text.split())
+            input_metrics.add_text_metrics(text)
 
             arguments.body["messages"].append(
                 {"role": "user", "content": [{"type": "text", "text": text}]}
@@ -394,8 +382,7 @@ class GenerativeAudioTranscriptionRequestFormatter(RequestFormatter):
         if prefix or text:
             prompt = prefix + text
             arguments.body["prompt"] = prompt
-            input_metrics.text_characters = len(prompt)
-            input_metrics.text_words = len(prompt.split())
+            input_metrics.add_text_metrics(prompt)
 
         return GenerationRequest(
             request_type="audio_transcriptions",
--- a/src/guidellm/schemas/request.py
+++ b/src/guidellm/schemas/request.py
@@ -169,6 +169,16 @@ class UsageMetrics(StandardBaseDict):
             self.video_tokens or 0
         ) + (self.audio_tokens or 0) or None
 
+    def add_text_metrics(self, text):
+        """
+        Adds the metrics from the given text to the fields
+        `text_characters` and `text_words`.
+
+        :param text: Text to add metrics from
+        """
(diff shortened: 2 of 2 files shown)


src/guidellm/data/deserializers/synthetic.py
```diff
@@ -174,14 +176,14 @@ def __iter__(self) -> Iterator[dict[str, Any]]:
...
-        prompt_token_ids = []
+        prompt_token_ids: list[int] = []
         avg_chars_per_token = 5
         margin_of_safety = 1.5
         attempts = 0
 
         while len(prompt_token_ids) < prompt_tokens_count:
             attempts += 1
-            num_chars = (
+            num_chars = math.ceil(
```
sjmonson: This should probably be an int cast not `ceil`.

src/guidellm/data/loaders.py
```diff
@@ -98,9 +100,13 @@ def generator(
                         continue
 
                     for preprocessor in self.preprocessors:
-                        row = preprocessor(row)
+                        processed_row = preprocessor(row)
+                        if isinstance(processed_row, GenerationRequest):
+                            yield processed_row
+                        else:
+                            row = processed_row
                     yield row
```
sjmonson: This is wrong, its yielding every row twice. I don't really like this code anyways since it makes multi-turn difficult; so maybe type ignore it for now and we revisit soon.

src/guidellm/data/utils/functions.py
```diff
@@ -5,14 +5,12 @@
...
     text: str,
-) -> dict[Literal["type", "text", "num_chars", "num_words"], str | int]:
+) -> dict[Literal["num_chars", "num_words"], int]:
     """Compute basic text statistics."""
     num_chars = len(text)
     num_words = len(text.split())
 
     return {
-        "type": "text",
-        "text": text,
```
sjmonson, line 15: We should probably remove this function entirely as it seems like a holdover from the Jinja2 version.

src/guidellm/data/deserializers/synthetic.py
jaredoconnell: Changed.

src/guidellm/data/utils/functions.py
jaredoconnell: Yeah this seemed really odd. Removed.

src/guidellm/data/loaders.py
jaredoconnell: I see. I could fix the logic, but given the upcoming change I left a comment mentioning the problem instead and type ignored it.

src/guidellm/data/deserializers/deserializer.py
```diff
@@ -107,3 +86,44 @@ def deserialize(
...
+                        **data_kwargs,
+                    )
+            except Exception as e:  # noqa: BLE001 # The exceptions are saved.
+                errors.append(e)
+
+            if dataset is not None:
+                break  # Found one that works. Continuing could overwrite it.
+
+        if dataset is None and len(errors) > 0:
+            raise DataNotSupportedError(
```
markurtz: This error message can get a bit unruly here, can we distinguish between not supported errors and dataset errors?

src/guidellm/data/preprocessors/formatters.py
```diff
@@ -191,27 +197,25 @@ def __call__(  # noqa: C901, PLR0912, PLR0915
...
                 continue
 
-            stats = text_stats(prefix)
-            if (num_chars := stats.get("num_chars")) is not None:
-                input_metrics.text_characters = (
-                    input_metrics.text_characters or 0
-                ) + num_chars
-            if (num_words := stats.get("num_words")) is not None:
-                input_metrics.text_words = (input_metrics.text_words or 0) + num_words
+            input_metrics.text_characters = (
```
markurtz: I would like to keep the counting logic unified in the text_stats function. We could update it so that it pushes to a UsageMetrics instance directly / make that a functoin of the UsageMetrics class, but this code is duplicated and if we add anymore stats, it will make it hard to keep updated

src/guidellm/data/preprocessors/formatters.py
```diff
@@ -387,10 +392,10 @@ def __call__(  # noqa: C901
         prefix = "".join(pre for pre in columns.get("prefix_column", []) if pre)
         text = "".join(txt for txt in columns.get("text_column", []) if txt)
         if prefix or text:
-            arguments.body["prompt"] = prefix + text
-            stats = text_stats(arguments.body["prompt"])
-            input_metrics.text_characters = stats.get("num_chars")
-            input_metrics.text_words = stats.get("num_words")
+            prompt = prefix + text
+            arguments.body["prompt"] = prompt
+            input_metrics.text_characters = len(prompt)
```
markurtz: Same note on handling text stats

src/guidellm/data/deserializers/deserializer.py
jaredoconnell: Done.

src/guidellm/data/preprocessors/formatters.py
jaredoconnell: Fixed. I think the new code is a much more elegant solution.

markurtz (approved):

PR #439, opened 2025-10-30
Fix test output xfail ut

tukwila:
## Summary

Fix two problems:
1. UT: tests/unit/benchmark/test_output:test_file_csv is marked as fail
2. some code quality check failure 

## Details

- [ ]

## Test Plan

-

## Related Issues

- Resolves #

---

## Use of AI

### DIFF
--- a/src/guidellm/benchmark/schemas.py
+++ b/src/guidellm/benchmark/schemas.py
@@ -24,6 +24,8 @@ from typing import Any, ClassVar, Literal, TypeVar, cast
 
 import yaml
 from pydantic import (
+    AliasChoices,
+    AliasGenerator,
     ConfigDict,
     Field,
     ValidationError,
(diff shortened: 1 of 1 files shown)


tukwila:
@markurtz @jaredoconnell pls help to check this fix is ok?

src/guidellm/benchmark/output.py
```diff
@@ -690,7 +690,7 @@ def _get_benchmark_extras_headers_and_values(
         values: list[str] = [
             benchmark.benchmarker.profile.model_dump_json(),
             json.dumps(benchmark.benchmarker.backend),
-            json.dumps(benchmark.benchmarker.requests["data"]),
+            json.dumps(benchmark.benchmarker.requests["attributes"]["data"]),
```
jaredoconnell: This is not a valid key. The only valid key is "data" when I run an example with this input:
```
guidellm benchmark run --profile constant --data "prompt_tokens=4096,prompt_tokens_stdev=512,prompt_tokens_min=2048,prompt_tokens_max=6144,output_tokens=512,output_tokens_stdev=128,output_tokens_min=1,output_tokens_max=1024" --rate 1 --target http://localhost:8000 --max-seconds 10 --output-formats csv --output-path ../guidellm_output/result_16.csv
```

If attributes does exist when you run it, which inputs are you using? If attributes is what we want to include, code will need to be put in place to account for it being a special case.

tukwila: yes, my fault. I found out 'attributes' in mock benchmarker; so i rollback this fix and udpated mock request dict. Then the testing cmd and UT test can be pass.

@markurtz Please help to confirm if i can update BenchmarkerDict.requests to delete 'attributes'?

PR #438, opened 2025-10-29
Add over saturation constraint

AlonKellner-RedHat:
## Summary

This PR adds over-saturation stopping to the GuideLLM CLI.
It's based on the OSD (Over-Saturation Detection) algorithm we developed and evaluated at Jounce.
Use `--stop-over-saturated` or `--stop-osd` to enable.

## Details

This PR adds:
- [x] Over-saturation stopping (`--stop-over-saturated`)
- [x] Comprehensive OSD unit tests

## Test Plan

- [x] Currently, only unit tests
- [x] When #440 lands, we'll enable its over-saturation stopping E2E test

---

- [x] "I certify that all code in this PR is my own, except as noted below."

## Use of AI

- [x] Includes AI-assisted code completion
- [x] Includes code generated by an AI application
- [x] Includes AI-generated tests (NOTE: AI written tests should have a docstring that includes `## WRITTEN BY AI ##`)

PR #440, opened 2025-10-30
test(e2e): basic E2Es

AlonKellner-RedHat:
## Summary

E2E tests which check basic GuideLLM functionality, using vLLM simulator.

## Details

- [x] Max requests test
- [x] Max duration test
- [ ] Over-saturation stopping test - skipped for now, will be enabled when #438 lands

## Test Plan

- [x] Local testing
- [x] GitHub action

---

## Use of AI


pyproject.toml
```diff
@@ -55,6 +55,7 @@ keywords = [
 dependencies = [
     "click>=8.0.0,<8.2.0",
     "culsans~=0.9.0",
+    "aiologic~=0.14.0",
```
jaredoconnell: Where is this new dependency used?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Set up Python
+        run: |
+          echo uv
+          curl -LsSf https://astral.sh/uv/install.sh | sh
+          chmod +x $HOME/.local/bin/uv $HOME/.local/bin/uvx
+          uv self update
+          uvx --version
+          echo python
+          uv python install 3.12 --default
+          python --version
```
jaredoconnell: During our call, we concluded that it would likely be a better practice to use the pylock to set dependencies in the E2E tests.

tests/e2e/vllm-sim.Dockerfile
sjmonson, line 1: llm-d-inference-sim publishes official container images [here](https://github.com/llm-d/llm-d-inference-sim/pkgs/container/llm-d-inference-sim). Could you reuse there image for the first stage rather than building it. Something like

```dockerfile
FROM ghcr.io/llm-d/llm-d-inference-sim:v0.3.0 as base

FROM scratch
COPY --from=base /app/llm-d-inference-sim/bin /bin
```

tests/e2e/utils.py
```diff
@@ -0,0 +1,336 @@
...
+        """
+        guidellm_exe = get_guidellm_executable()
+
+        # Build command components
+        cmd_parts = [
+            *([f"{k}={v}" for k, v in extra_env.items()] if extra_env else []),
+            "HF_HOME=/tmp/huggingface_cache",
+            f"{guidellm_exe} benchmark",
+            f'--target "{self.target}"',
+            f"--rate-type {rate_type}",
```
sjmonson:
```suggestion
            f"--profile {rate_type}",
```

pyproject.toml
AlonKellner-RedHat: It's not used, but weirdly it was breaking all tests (unit, integration & E2E).
`aiologic` released a new version today which breaks `culsans`, and that broke us - which is strange, because you'd expect our `pylock.toml` to guard us against something of the sort.
Well, I just found out why - we were using tox out of the box, and by default it uses the `pyproject.toml` for creating environments, not the `pylock.toml`. `tox` got the wrong `aiologic` version, and that made the tests fail.
I modified the `install_command` in `tox.ini`, which might fix this issue. I will try to remove this version pin once I verify my solution.

.github/workflows/development.yml
sjmonson, line 1: Thanks for this, I have been meaning to take a look at tox-pdm.

pyproject.toml
x42005e1f: > `aiologic` released a new version today which breaks `culsans`, and that broke us - which is strange, because you'd expect our `pylock.toml` to guard us against something of the sort.

`aiologic==0.15.0` is incompatible with `culsans==0.9.0`, but compatible with `culsans==0.10.0`.

sjmonson (changes requested):
Few small nits

pyproject.toml
AlonKellner-RedHat: That's cool, maybe we could update at some point, not sure why it's pinned right now, maybe @markurtz knows more.

x42005e1f: And yes, I apologize, I completely forgot that you use semantic versioning. The only reason why they are not compatible is that I removed a private module (`aiologic.lowlevel._thread`) that was used by `culsans<0.10.0`, but whose features are now exposed via public functions (`aiologic.lowlevel.create_thread_lock()`, etc.).

As for `0.x.y`, the patch version (`y`) has the same semantics as for `1.x.y`, and the minor version (`x`) includes both compatible and incompatible changes. But I usually keep track of the dependent repositories I know about and try to make sure that the code they use does not require changes (at least within one update).

There are quite a few changes in `aiologic==0.15.0`, but overall it has greatly improved its reliability for all use cases.

.github/workflows/e2e.yml
AlonKellner-RedHat: @jaredoconnell Unfortunately it ended up as a change to tox usage all across the board - since we never actually used the `pylock.toml` in our tests.
I added a tox plugin called `tox-pdm` which uses our lock file, then I integrated that modification to all tox related workflows.
From now on - every single tox based test environment will use pdm and `pylock.toml` by default.

pyproject.toml
AlonKellner-RedHat: Amazing! @jaredoconnell in any case - I just updated `culsans` to 0.10.0 and removed the explicit `aiologic` dependency - it works perfectly 💯

scripts/generate_pylock.sh
sjmonson, line 1: Revert this. It resolves to `python>=3.10,<4.0`, `python>=3.11,<4.0`, `python>=3.12,<4.0`; overlapping ranges do not make sense to lock separately. The previous version was `python>=3.12,<4.0`, `python>=3.10,<3.12`. Having a split is not strictly necessary anymore since we dropped 3.9 but it does allow us to lock newer packages for the modern python version.

sjmonson (changes requested):
One more nit. Also needs signoff, E.g. `git rebase --signoff main`.

scripts/generate_pylock.sh
AlonKellner-RedHat: I see. I can't really fully revert this, because the original generated `pylock.toml` does not work for 3.10 (for example  `torchcodec` revolves only 3.12 and 3.13), and the CI fails.
I am working on modifying my version to generate non-overlapping ranges that still work for both 3.10 and 3.13 to pass the CI. Will update

AlonKellner-RedHat: OK I see the issue - I was using the `generate_pylock.sh` script without `-f` which generated an invalid `pylock.toml` (not sure exactly why, seems like a strange interaction between `--append` and `--update-reuse`).
In any case, I am applying a full revert of my changes

PR #441, opened 2025-10-30
Support dashed arguments for benchmark args

sjmonson:
## Summary

For all arguments in `BenchmarkGenerativeTextArgs` support dashed versions of the name during validation to match the CLI. Additionally alias the argument `data_request_formatter` to `request_type`.

## Details

Alias are configured so that if both options are specified, the original name takes precedence. E.g. the order of precedence is `max_seconds > max-seconds`.

~~For `data_request_formatter` the request type alias takes precedence. E.g. `request_type > request-type > data_request_formatter`.~~ Changed this because it caused precedence issues with CLI arg. Order is now `data_request_formatter > data-request-formatter > request_type > request-type`.

## Test Plan

The following scenario file can be used to test. Modify which fields are uncommented to test precedence. 

```yaml
target: <hostname>
#request_type: text_completions
#request-type: text_completions
data_request_formatter: text_completions
profile: concurrent
rate: 2
max_seconds: 30
data:
  prompt_tokens: 256
  output_tokens: 128
```

## Related Issues

- Resolves #

---

## Use of AI

### DIFF
--- a/src/guidellm/benchmark/schemas.py
+++ b/src/guidellm/benchmark/schemas.py
@@ -1887,7 +1887,12 @@ class BenchmarkGenerativeTextArgs(StandardBaseModel):
     data_request_formatter: DatasetPreprocessor | dict[str, str] | str = Field(
         default="chat_completions",
         description="Request formatting preprocessor or template name",
-        validation_alias=AliasChoices("request_type", "request-type"),
+        validation_alias=AliasChoices(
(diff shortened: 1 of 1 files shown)


jaredoconnell (approved):
This seems fine. This may cause some fragmentation in usage, but this doesn't seem like something that would be a maintenance burden.

markurtz (approved):

PR #442, opened 2025-10-31
[GuideLLM Refactor] Updates and Fixes for benchmark outputs, schemas, and stats calculations

markurtz:
## Summary

Screenshot 2025-10-31 at 1 51 32 PM

## Details

- [ ]

## Test Plan

-

## Related Issues

- Resolves #

---

## Use of AI

### DIFF
--- a/src/guidellm/benchmark/schemas/base.py
+++ b/src/guidellm/benchmark/schemas/base.py
@@ -175,29 +175,46 @@ class TransientPhaseConfig(StandardBaseModel):
         duration_transition_time: float | None = None
         request_transition_time: float | None = None
 
+        # Calculate transition times for the phase based on phase limits and period
+        # Potential phases: start (warmup) -> active -> end (cooldown)
+        #   Warmup transition times: (start, start + duration)
+        #   Active transition times: (start + duration, end - duration)
+        #   Cooldown transition times: (end - duration, end)
         if period == "start":
             if phase_duration is not None:
+                # Duration was set and caculating for "warmup" / start phase
+                # Phase is active for [start, start + duration]
                 duration_transition_time = state.start_time + phase_duration
             if phase_requests is not None:
+                # Requests was set and calculating for "warmup" / start phase
+                # Phase is active for requests [0, phase_requests]
+                # Grab start time of the next request as transition time
+                # (all requests up to and including phase_requests are in warmup)
                 request_transition_time = (
                     info.started_at
                     if info.started_at is not None
-                    and state.processed_requests > phase_requests
+                    and state.processed_requests == phase_requests + 1
                     else -1.0
                 )
         elif period == "end":
             if phase_duration is not None:
+                # Duration was set and calculating for "cooldown" / end phase
+                # Phase is active for [end - duration, end]
                 duration_transition_time = (
                     state.start_time + state.progress.total_duration - phase_duration
                     if state.progress.total_duration is not None
                     else -1.0
                 )
             if phase_requests is not None:
+                # Requests was set and calculating for "cooldown" / end phase
+                # Phase is active for requests [total - phase_requests, total]
+                # Grab completion time of the request right before cooldown starts
+                # (all requests from that point onward are in cooldown)
                 request_transition_time = (
                     info.completed_at
                     if info.completed_at is not None
                     and state.progress.remaining_requests is not None
-                    and state.progress.remaining_requests < phase_requests + 1
+                    and state.progress.remaining_requests == phase_requests + 1
                     else -1.0
                 )
 
@@ -205,7 +222,7 @@ class TransientPhaseConfig(StandardBaseModel):
         transition_time: float | None = None
 
         if request_transition_time == -1.0 or duration_transition_time == -1.0:
-            # Transition defined but not yet reached, not enough info yet
+            # Transition defined but not yet reached or passed
             transition_active = True
             request_transition_time = None
         elif (
@@ -221,7 +238,7 @@ class TransientPhaseConfig(StandardBaseModel):
         elif (
             request_transition_time is not None or duration_transition_time is not None
         ):
-            # One limit defined; satisfy t hat one
+            # One limit defined; satisfy that one
             transition_active = True
             transition_time = request_transition_time or duration_transition_time
 


pyproject.toml
sjmonson, line 1: Can you run `tox -e lock` (or pre-commit) and commit the new lockfile?

src/guidellm/__main__.py
```diff
@@ -255,7 +255,7 @@ def benchmark():
     help="Number of worker processes for data loading.",
 )
 @click.option(
-    "--dataloader_kwargs",
+    "--dataloader-kwargs",
```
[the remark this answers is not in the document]
sjmonson: No

src/guidellm/benchmark/entrypoints.py
```diff
@@ -195,45 +216,74 @@ async def resolve_request_loader(
...
+    if isinstance(data_request_formatter, RequestFormatter):
+        data_request_formatter_instance = data_request_formatter
+    else:
+        if isinstance(data_request_formatter, str):
+            request_type = data_request_formatter
+            formatter_kwargs: dict[str, Any] = {}
+        else:
+            # Extract request_type from formatter dictionary
+            formatter_dict = dict(data_request_formatter)
+            request_type = formatter_dict.pop("request_type", "chat_completions")
```
sjmonson:
```suggestion
            request_type = formatter_dict.pop("request_type", settings.preferred_route)
```

Also need to update `preferred_route` to be `chat_completions`.

src/guidellm/benchmark/entrypoints.py
```diff
@@ -301,9 +355,10 @@ async def resolve_profile(
         if val is not None:
             constraints[key] = val
     if not isinstance(profile, Profile):
+        rate_list: list[float] | None = [rate] if isinstance(rate, float) else rate
```
sjmonson: After rebase should be able to drop this line (and change function sig to `rate: list[float] | None`) due to:

https://github.com/vllm-project/guidellm/blob/7b07866affdfa25305db8d3db9d5e850dfa36a39/src/guidellm/benchmark/schemas.py#L1956-L1974

src/guidellm/scheduler/worker_group.py
```diff
@@ -647,11 +647,13 @@ def _update_state_request_counts(self, info: RequestInfo):
             self._processing_request_ids.add(info.request_id)
             self._state.processing_requests = len(self._processing_request_ids)
         elif info.status == "completed":
+            info.timings.finalized = time.time()
```
sjmonson: Call `time.time()` once at top of function.

sjmonson (changes requested):
Aside from the max-seconds bug here are some high-level notes:

- Needs signoff
- Needs cleanup. To retroactively run pre-commit on only the changes: `pre-commit run --from $(git merge-base main@{u} HEAD) --to HEAD`
- Only glanced over the accumulator and statistics code. Will do a more in-depth review time permitting but don't block on it. 
- Is warmup/cooldown only seconds now and not sometimes a percent? Setting `--warmup .1` results in a table entry of `Warm Sec: 0.1`.

src/guidellm/benchmark/outputs/csv.py
```diff
@@ -0,0 +1,692 @@
...
+        self._add_field(
+            headers, values, "Run Info", "Backend", json.dumps(benchmark.config.backend)
+        )
+        self._add_field(
+            headers,
+            values,
+            "Run Info",
+            "Environment",
+            json.dumps(benchmark.config.environment),
+        )
```
jaredoconnell, line 225: The requests and environment values are not the most pretty, but it works and doesn't exclude any information.
The format doesn't preserve the format of the original input, like it does on main right now, but if that's okay then I'm okay with proceeding.

src/guidellm/scheduler/worker_group.py
```diff
@@ -228,11 +228,11 @@ async def create_processes(self):
 
             worker = WorkerProcess[RequestT, ResponseT](
                 worker_index=rank,
-                messaging=self.messaging.create_worker_copy(
+                messaging=self.messaging.create_worker_copy(  # type: ignore[arg-type]
```
jaredoconnell, line 226: It may make sense to explain this ignored type error. If someone tries to fix it, this one is a confusing one.

jaredoconnell (commented):
I can confirm that the max-seconds error is fixed. I'll do more testing tomorrow, but it works for me.
The new CLI table outputs are a little busy to look at, but work. Adding more horizontal padding may help, but isn't necessary.

src/guidellm/benchmark/outputs/csv.py
markurtz: Yeah, I was of the opinion that for the csv side, this is more one off, manual debugging in spreadsheets if someone needs it. If they want actual access to the data, to query/manipulate/act on it, then they should be interacting with either the json file or loading it into a python object. I could be moved the otherway to try and format it better, but I think I would consider that a follow up to this PR

src/guidellm/scheduler/worker_group.py
markurtz: Added the comment back in that you had

sjmonson (approved):
LGTM Either land #440 first for test fixes or ignore.

jaredoconnell (commented):
For the throughput+rampup case, is it expected that it goes from very few requests (like one or two per second), to full throughput, rather than scaling linearly?

No issues stood out during my most recent code review. We just need to clarify this and fix the tests.

src/guidellm/benchmark/schemas/base.py
```diff
@@ -120,127 +120,112 @@ def create_from_value(
...
+                    else -1.0
+                )
+            if phase_requests is not None:
+                request_transition_time = (
+                    info.completed_at
+                    if info.completed_at is not None
+                    and state.progress.remaining_requests is not None
+                    and state.progress.remaining_requests < phase_requests + 1
+                    else -1.0
+                )
```
jaredoconnell, line 219: Some brief comments in each of these cases may make it easier to read the code.

src/guidellm/benchmark/schemas/base.py
```diff
@@ -120,127 +120,112 @@ def create_from_value(
...
+            transition_active = True
+            transition_time = (
+                min(request_transition_time, duration_transition_time)
+                if period == "end"
+                else max(request_transition_time, duration_transition_time)
+            )
+        elif (
+            request_transition_time is not None or duration_transition_time is not None
+        ):
+            # One limit defined; satisfy t hat one
```
jaredoconnell: Extra space

src/guidellm/benchmark/progress.py
```diff
@@ -682,7 +661,7 @@ def _update_processing_states(
 
     def _update_request_stats(
         self,
-        request_concurrency: int | None = None,
+        request_concurrency: float | None = None,
```
jaredoconnell, line 664: Is this a float because it's a statistic of average integer concurrency over a period?

src/guidellm/benchmark/schemas/base.py
markurtz: Fixed

src/guidellm/benchmark/schemas/base.py
markurtz: Added

src/guidellm/benchmark/progress.py
markurtz: Yep, correct, it's the time weighted average of concurrency across all possible intervals (request start, request stop) for all requests

src/guidellm/benchmark/schemas/base.py
```diff
@@ -175,37 +175,54 @@ def compute_transition_time(
...
             if phase_requests is not None:
+                # Requests was set and calculating for "warmup" / start phase
+                # Phase is active for requests [0, phase_requests]
+                # Grab start time of the next request as transition time
+                # (all requests up to and including phase_requests are in warmup)
                 request_transition_time = (
                     info.started_at
                     if info.started_at is not None
-                    and state.processed_requests > phase_requests
+                    and state.processed_requests == phase_requests + 1
```
jaredoconnell, line 196: Using a == here means that the variable will be set to -1.0 if it's not on the exact request value. Is this function called in such a way that it will always hit that, or in a way that it's okay if it's skipped?

jaredoconnell (approved):
Looks good. Ready for merge from my end unless my comment brings up any need for changes.

src/guidellm/benchmark/schemas/base.py
markurtz: it's something that I caught when going through the updated code comments. The accumulator is constantly updating the measurement times (in case there are multiple constraints that are generating differing values). In that case, we can end up in a place where if it was set by requests, we are then reporting a differing time each time once we get into the cooldown period. This way ensures we either report a time that should be updated or we tell the caller the time is active but we don't have an update yet.

PR #463, opened 2025-11-14
Ensure synthetic text datasets remain random across benchmarks

### LINKED ISSUE #460: Request repetition across sweep rounds leads to 100% prefix-hit ratio (possible bug)
**Describe the bug**
We noticed that request payloads are repeating across sweep rounds when running a benchmark with `rate-type=sweep`.

Example configuration:
- max_req = 5
- rate-type = sweep
- rates = [sync, tput, 2 req/s, 4 req/s] (4 rounds total)
- Single benchmark 

Observed Behavior: Within a single benchmark run, the same 5 requests are reused in every sweep iteration. For instance:
- Iteration 1 (sync): req0, req1, req2, req3, req4
- Iteration 2 (tput): req0, req1, req2, req3, req4
- Iteration 3 (2 req/s): req0, req1, req2, req3, req4
- Iteration 4 (4 req/s): req0, req1, req2, req3, req4

This repetition continues for all rounds.

Because the same requests are reused after the first iteration, the prefix cache is fully warm, resulting in a 100% prefix-hit ratio. This makes comparisons across sweep rates unfair,

**Expected behavior**

Each sweep iteration should ideally use unique requests, or there should be an option to disable request reuse. This would avoid prefix-cache artifacts and allow accurate measurement of prefill performance across rate variations.

markurtz:
## Summary

Fix synthetic dataset so it can preserve randomness across benchmarks run in the same session by enforcing set_epoch across the data loader and iterators chain. This was happening due to a new iterator being created from the DataLoader for each benchmark. The solution was to preserve epoch information across datasets so they can make use of that, if needed, and in the case of SyntheticTextGeneration, to increment a random seed.

## Test Plan

Automation tests

## Related Issues

- Resolves #460 

---

- [ X] "I certify that all code in this PR is my own, except as noted below."

## Use of AI


src/guidellm/data/loaders.py
```diff
@@ -74,18 +75,29 @@ def __iter__(self) -> Iterator[DataT]:
...
+        epoch: int = 0,
     ) -> Iterator[DataT]:
         gen_count = 0
 
         with contextlib.suppress(StopIteration):
-            dataset_iters = [iter(dataset) for dataset in self.datasets]
+            dataset_iters = []
+            for dataset in self.datasets:
+                if hasattr(dataset, "set_epoch"):
+                    with contextlib.suppress(Exception):
```
sjmonson, line 98: What Exception could happen here?

markurtz: Mainly wasn't sure if other IterableDatasets would complain about that not being supported or not work with it, so just wanted a catch all just in case

PR #465, opened 2025-11-18
update build versions in settings, workflow versioned build fix

DaltheCow:
## Summary

Updating the UI build version default in settings to v4.0.0. Fixing the release tag issue. Setting the directory path with ${TAG} is not working.

## Details

- prod and staging build defaults updated to v0.4.0
- release/staging workflows versioned build directory in gh-pages branch fixed

## Test Plan

- Manually tested the build version updates
- Trial tested the workflow change in the dev workflow to confirm methodology

---

### DIFF
--- a/.github/workflows/release-candidate.yml
+++ b/.github/workflows/release-candidate.yml
@@ -226,8 +226,8 @@ jobs:
           export $(grep -v '^#' .env.staging | xargs)
 
           # Set asset prefix and base path with git tag
-          ASSET_PREFIX=https://blog.vllm.ai/guidellm/ui/release/${{ env.TAG }}
-          BASE_PATH=/ui/release/${{ env.TAG }}
+          ASSET_PREFIX=https://blog.vllm.ai/guidellm/ui/${{ env.TAG }}
+          BASE_PATH=/ui/${{ env.TAG }}
           GIT_SHA=${{ github.sha }}
           export ASSET_PREFIX=${ASSET_PREFIX}
           export BASE_PATH=${BASE_PATH}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -221,8 +221,8 @@ jobs:
           export $(grep -v '^#' .env.production | xargs)
 
           # Set asset prefix and base path with git tag
-          ASSET_PREFIX=https://blog.vllm.ai/guidellm/ui/${TAG}
-          BASE_PATH=/ui/${TAG}
+          ASSET_PREFIX=https://blog.vllm.ai/guidellm/ui/${{TAG}}
```
sjmonson: This probably needs to be `${{ env.TAG }}` and likewise for any other usage in the job.

DaltheCow: @sjmonson Is that preferable to github.ref_name? That is what I had been using prior and seemed to work, but this was an issue where the TAG variable defined above (with a value of github.ref_name) seemed to not be interpolated correctly so it was leading to improper prefixes where the value was not being parsed. Just thought I'd ask in case it doesn't matter so I don't have to test everything again lol. Happy to change it though

sjmonson: Before it was `${TAG}` which works within the bash context that the variable was set, but I don't think it works when used as a workflow variable or in any future bash sections. `${{ env.TAG }}` is the appropriate place to pull GitHub variables from.

DaltheCow: Sweet thank you!

.github/workflows/release-candidate.yml
sjmonson, line 219: One more thing. For release-candidate `${{ github.ref_name }}` is the branch name (e.g. "release/v0.5.0") so you probably want to cut out the `release/` part.

sjmonson (changes requested):

.github/workflows/release-candidate.yml
DaltheCow: Sweet!

PR #467, opened 2025-11-18
Indicate max_concurrency for throughput and disallow running standalone without --rate

sjmonson:
## Summary

Add a indicator to throughput similar to concurrent's `concurrent@CONC`. Additionally require specifying `--rate` for throughput when running standalone rather than defaulting to `MAX_CONCURRENCY`.

---

## Use of AI


markurtz (approved):
LGTM

PR #468, opened 2025-11-18
Unmask StopIteration in DataLoader

sjmonson:
## Summary

Fixes bug where dataloader will not recognize the end of a dataset due to the StopIteration Exception being caught and masked.

---

## Use of AI


jaredoconnell (approved):

jaredoconnell:
It was agreed offline that follow up work would be a good idea regarding this section of code. The exception handling may be overly-broad, and if a non-flaky test is possible, we should add one.

PR #469, opened 2025-11-18
Fixes and Cleanup of CI

sjmonson:
## Summary

Deduplicates much of the CI code by defining shared workflows and composite actions. Included in this is fixing some breakage introduced with #440

## Details

This PR mostly keeps the same functionally as before with the main exception being pytest markers (when set) are now uniformly applied to all tests jobs.

After this is approved the required CI list needs to be updates to enable merging.

Some follow-up work (post this PR):

- Deduplicate build and publish actions
- Enable caching for shared workflow actions (such as pip installs, container builds)
- Reduce coverage for certain stages
  - E.g. Run subset of tests on PRs

## Test Plan

- See CI runs on this PR for results

---

## Use of AI

### DIFF
--- a/.github/actions/run-tox/action.yml
+++ b/.github/actions/run-tox/action.yml
@@ -24,4 +24,6 @@ runs:
     - name: Run tox
       run: |
         tox run -e "${{ inputs.tox-env }}" -- ${{ inputs.tox-args }}
+        # errno 5 means all tests were filtered out, ignore
+        ret=$? && [[ $ret -eq 5 ]] && exit 0; exit $ret
       shell: bash


.github/workflows/main.yml
```diff
@@ -6,191 +6,30 @@ on:
       - main
 
 jobs:
-  quality-checks:
-    runs-on: ubuntu-latest
+  tests:
     strategy:
       matrix:
         python: ["3.10", "3.11", "3.12", "3.13"]
```
jaredoconnell, line 12: It may be a good idea to switch this to a GitHub variable.

.github/workflows/development.yml
```diff
@@ -5,191 +5,31 @@ on:
...
-    permissions:
-      contents: "read"
-    runs-on: ubuntu-latest
-    steps:
-      - name: Check out code
-        uses: actions/checkout@v4
+    uses: ./.github/workflows/testing.yml
+    with:
+      python: ${{ matrix.python }}
+      args: -m "smoke or sanity"
```
jaredoconnell: Are the regression tests run anywhere in CI after this change?

sjmonson: Yes, everywhere except PRs.

.github/workflows/main.yml
sjmonson: Maybe, but its different for each stage (PR, main, release, rc, nightly).

.github/workflows/development.yml
jaredoconnell: The concern I have is that in a PR, someone may unknowingly break regression tests in their PR, breaking other CI. Is it at least run for PRs into main?

sjmonson: For PRs I would be nice to reduce the total number of tests since it runs so frequently. This is just reverting to the behavior before #255 to match integration and e2e. Maybe not the right play tho. Perhaps instead we can just run tests for python 3.10 and not both 3.13 and 3.10 on PRs.

jaredoconnell: Yeah. That's an option. The E2E tests are the most important ones to run on the latest version.
I just checked and the unit-tests on 3.13 on another PR took around 2m 9s to complete. And this PR has it took 2m 9s. So that's literally identical. So I don't see a value disabling regression tests for a 0s improvement.

.github/actions/run-tox/action.yml
```diff
@@ -0,0 +1,27 @@
...
+      uses: pdm-project/setup-pdm@v4
+      with:
+        python-version: ${{ inputs.python-version }}
+    - name: Install dependencies
+      run: |
+        pip install tox tox-pdm
+      shell: bash
+    - name: Run tox
+      run: |
+        tox run -e "${{ inputs.tox-env }}" -- ${{ inputs.tox-args }}
```
sjmonson, line 26: `pytest` will return with errono 5 if the markers filter out all tests. Need to write a simple fix for this before merge.

markurtz (approved):
LGTM

PR #472, opened 2025-11-19
Reenable and improve preprocess dataset

jaredoconnell:
## Summary
This PR re-enables, tests, and documents the `preprocess dataset` command.
Also changes the format that prompt and output sizes are specified, and makes the code aware of prefixes.

## Details

- Uses the post-refactor code to re-enable the command.
- Switches over to the same format used by `benchmark run`'s synthetic data for the data config to enable more features and make the command more cohesive with the rest of GuideLLM.
- Adds options for prefixes. I added an option to include prefixes in the count, since prefixes are included in input and output tokens, and affect performance.

## Test Plan

- Run with a known dataset, or create one as a simple CSV.
- New tests are added that should cover everything except huggingface uploads. They are all at least in part generated by AI, but I went through each one iteratively to ensure they do what they need to do.

---

## Use of AI

### DIFF
--- a/src/guidellm/extras/audio.py
+++ b/src/guidellm/extras/audio.py
@@ -107,8 +107,13 @@ def _decode_audio(  # noqa: C901, PLR0912
             raise ValueError(
                 f"Audio dict must contain either 'data' or 'url' keys, got {audio}"
             )
+        audio_data = audio["data"] if "data" in audio else audio.get("url")
+        if audio_data is None:
+            raise ValueError(
+                f"Audio dict must contain either 'data' or 'url' keys, got {audio}"
+            )
         return _decode_audio(
-            audio=audio.get("data") or audio.get("url"),
+            audio=audio_data,
             sample_rate=sample_rate,
             max_duration=max_duration,
         )


src/guidellm/data/entrypoints.py
```diff
@@ -0,0 +1,601 @@
...
+
+    if prefix_buckets is None:
+        return None  # No configuration
+
+    # Use prefix buckets with weighted sampling
+    from random import Random
+    rand = Random(random_seed)
+    bucket_choices = []
+    for bucket in prefix_buckets:
+        bucket_choices.extend([bucket.prefix_tokens] * bucket.bucket_weight)
```
sjmonson: ~~This does not account for `prefix_count` see implementation in `synthetic.py`.~~

Ignore this comment see top-level one

src/guidellm/data/entrypoints.py
```diff
@@ -0,0 +1,601 @@
...
+
+
+class ShortPromptStrategy(str, Enum):
+    IGNORE = "ignore"
+    CONCATENATE = "concatenate"
+    PAD = "pad"
+    ERROR = "error"
+
+
+def handle_ignore_strategy(
```
sjmonson: For better organization I think these `handle_*_strategy` methods should be defined on a static class.

src/guidellm/data/entrypoints.py
```diff
@@ -0,0 +1,601 @@
...
+        "output_tokens_count_column", []
+    )
+    output_column = (
+        output_mappings[0][1] if output_mappings else "output_tokens_count"
+    )
+
+    return prompt_column, prefix_column, output_column
+
+
+def _create_token_samplers(
```
sjmonson: ~~I don't love how this is duplicated from the synthetic dataset methods. Preferably this is moved into its own class/functions under `synthetic.py` and can be reused both places.~~

Ignore this comment see top-level one

src/guidellm/__main__.py
```diff
@@ -486,6 +488,150 @@ def preprocess():
...
+    help=(
+        "Delimiter for concatenating short prompts (used with 'concatenate' strategy)."
+    ),
+)
+@click.option(
+    "--prefix-tokens",
+    type=int,
+    default=None,
+    help="Single prefix token count (alternative to prefix_buckets in config).",
+)
```
sjmonson: ~~Maybe drop this argument. You can already do `--config prefix_tokens=10,...` to achieve the same thing. The underlying logic in `SyntheticTextDatasetConfig` will automatically convert it to a single bucket.~~

Ignore this comment see top-level one

src/guidellm/data/entrypoints.py
```diff
@@ -0,0 +1,601 @@
...
+                max_value=prefix_tokens,
+                random_seed=random_seed,
+            )
+        )
+
+    if prefix_buckets is None:
+        return None  # No configuration
+
+    # Use prefix buckets with weighted sampling
+    from random import Random
```
sjmonson: Should not import random inline

sjmonson (changes requested):
I think the reuse of `SyntheticTextDatasetConfig` is not a great idea. `prefix_buckets` are defined differently here. For synthetic data a prefix bucket of `prefix_tokens=10,prefix_count=1` means you get one identical prefix for the entire dataset. As implemented here `prefix_tokens=10,prefix_count=1` will only ensure that every row has a prefix of length 10. It does not guarantee any shared prefix between rows. 

Rather then reuse `SyntheticTextDatasetConfig` I think the best option is to create a new config format that is similar only where it makes sense. For example:

```yaml
prompt_tokens:
prompt_tokens_stdenv:
prompt_tokens_min:
prompt_tokens_max:
output_tokens:
output_tokens_stdenv:
output_tokens_min:
output_tokens_max:
prefix_tokens:
prefix_tokens_stdenv:
prefix_tokens_min:
prefix_tokens_max:
```

Treat prefix the same as prompt and output.

src/guidellm/data/entrypoints.py
jaredoconnell: Relevant code no longer present.

src/guidellm/__main__.py
jaredoconnell: This input is no longer present.

src/guidellm/data/entrypoints.py
jaredoconnell: Done.

src/guidellm/data/schemas.py
```diff
@@ -11,3 +23,110 @@
...
+
+    prefix_buckets: list[SyntheticTextPrefixBucketConfig] | None = Field(
+        description="Buckets for the prefix tokens distribution.",
+        default=None,
+    )
+
+    source: str = Field(
+        description="The source of the text data to be used for generation.",
+        default="data:prideandprejudice.txt.gz",
+    )
```
sjmonson: Can you drop this field? It no longer does anything.

src/guidellm/data/schemas.py
```diff
@@ -11,3 +23,110 @@
     "video_column",
     "audio_column",
 ]
+
+class DataNotSupportedError(Exception):
+    """
+    Exception raised when the data format is not supported by deserializer or config.
+    """
+
+class TokenCountConfig(StandardBaseModel):
```
sjmonson: I am not a huge fan of how it moves an arbitrary subset of fields to a superclass. Can you remove `TokenCountConfig` and lets revisit in a future release? Can leave `PreprocessDatasetConfig` and `SyntheticTextDatasetConfig` in schema.py for now. 

For v0.5.0 (or v0.6.0) I think we could have all of the deserializers follow a similar pattern to synthetic (config file + initialize class); in which case it would make more sense for preprocess to support the same `--data` argument as benchmark and maybe a separate custom config for preprocess args.

sjmonson (changes requested):

src/guidellm/data/schemas.py
jaredoconnell: Done.

sjmonson (dismissed):
Need to double-check that `benchmark run` is unaffected but LGTM.

sjmonson (approved):
LGTM

PR #478, opened 2025-11-21
Move request formatting to backend

sjmonson:
## Summary

Splits the `RequestFormatter` into multiple separate components across data loading, data preprocessing, and the backend. Necessary step for next attempt at multi-turn.

## Details

- Removed the `RequestFormatter` preprocessor
- Added "Encode" preprocessors for audio, video, and image
- Added a "Finalizer" stage after preprocessors to convert from column ordered data to request object
- Added most of previous formatter code to backend

## Test Plan

- Run all know benchmark well-lit pathways

## Related Issues

- Resolves #

---

## Use of AI

### DIFF
--- a/src/guidellm/data/preprocessors/encoders.py
+++ b/src/guidellm/data/preprocessors/encoders.py
@@ -53,7 +53,7 @@ class MediaEncoder(DatasetPreprocessor):
                     continue
 
                 encoded_audio.append(
-                    self.encode_audio(audio, b64encode=True, **self.encode_audio_kwargs)
+                    self.encode_audio(audio, **self.encode_audio_kwargs)
                 )
             columns["audio_column"] = encoded_audio
 
--- a/src/guidellm/backends/openai/request_handlers.py
+++ b/src/guidellm/backends/openai/request_handlers.py
@@ -10,6 +10,7 @@ extract usage metrics, and convert results into standardized GenerationResponse.
 
 from __future__ import annotations
 
+import base64
 from typing import Any, Protocol, cast
 
 from guidellm.schemas import GenerationRequest, GenerationResponse, UsageMetrics
@@ -466,7 +467,9 @@ class ChatCompletionsRequestHandler(TextCompletionsRequestHandler):
                         {
                             "type": "input_audio",
                             "input_audio": {
-                                "data": audio.get("audio"),
+                                "data": base64.b64encode(
+                                    audio.get("audio", b"")
+                                ).decode("utf-8"),
                                 "format": audio.get("format"),
                             },
                         }
--- a/src/guidellm/extras/audio.py
+++ b/src/guidellm/extras/audio.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import base64
 from pathlib import Path
 from typing import Any, Literal
 
@@ -33,7 +32,6 @@ def encode_audio(
     | np.ndarray
     | torch.Tensor
     | dict[str, Any],
-    b64encode: bool = False,
     sample_rate: int | None = None,
     file_name: str = "audio.wav",
     encode_sample_rate: int = 16000,
@@ -71,12 +69,8 @@ def encode_audio(
     )
 
     return {
-        "type": "audio_base64" if b64encode else "audio_file",
-        "audio": (
-            base64.b64encode(encoded_audio).decode("utf-8")
-            if b64encode
-            else encoded_audio
-        ),
+        "type": "audio_file",
+        "audio": encoded_audio,
         "file_name": get_file_name(audio)
         if isinstance(audio, str | Path)
         else file_name,
--- a/tests/unit/backends/openai/test_request_handlers.py
+++ b/tests/unit/backends/openai/test_request_handlers.py
@@ -734,7 +734,7 @@ class TestChatCompletionsRequestHandler:
         data = GenerationRequest(
             columns={
                 "audio_column": [
-                    {"audio": "base64data", "format": "wav"},
+                    {"audio": b"base64data", "format": "wav"},
                 ]
             },
         )
@@ -746,7 +746,7 @@ class TestChatCompletionsRequestHandler:
         assert result.body["messages"][0]["content"][0]["type"] == "input_audio"
         assert (
             result.body["messages"][0]["content"][0]["input_audio"]["data"]
-            == "base64data"
+            == "YmFzZTY0ZGF0YQ=="
         )
         assert (
             result.body["messages"][0]["content"][0]["input_audio"]["format"] == "wav"
--- a/tests/unit/extras/test_audio.py
+++ b/tests/unit/extras/test_audio.py
@@ -56,7 +56,6 @@ def test_encode_audio_with_tensor_input(sample_audio_tensor):
         sample_rate=16000,
         audio_format="mp3",
         bitrate="64k",
-        b64encode=False,
     )
 
     assert result["type"] == "audio_file"
@@ -67,20 +66,6 @@ def test_encode_audio_with_tensor_input(sample_audio_tensor):
     assert result["audio_seconds"] == 1.0
     assert isinstance(result["audio_bytes"], int)
     assert result["audio_bytes"] > 0
-
-
-def test_encode_audio_with_base64(sample_audio_tensor):
-    result = encode_audio(audio=sample_audio_tensor, sample_rate=16000, b64encode=True)
-
-    assert result["type"] == "audio_base64"
-    assert isinstance(result["audio"], str)
-    import base64
-
-    try:
-        decoded = base64.b64decode(result["audio"])
-        assert len(decoded) > 0
-    except (base64.binascii.Error, ValueError) as e:
-        pytest.fail(f"Invalid base64 encoding: {e}")
 
 
 def test_encode_audio_with_numpy_array(sample_audio_tensor):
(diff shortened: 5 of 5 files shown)


src/guidellm/backends/openai/http.py
```diff
@@ -13,15 +13,32 @@
...
+)
 
-__all__ = ["OpenAIHTTPBackend"]
+__all__ = [
+    "OpenAIHTTPBackend",
+    "OpenAIRequestType",
+]
+
+
+OpenAIRequestType = Literal[
```
jaredoconnell: So the proposal is to make this enum per-backend?
In my pull request I use this enum to validate required fields within the requests, like "prompt" for text completions, and "messages" for chat completions.
Maybe I can work around not having it based on the content itself. But this field is required on the data generation side for that, too. How are you planning on handling that? Or should it be as simple as passing the correct data types to the synthetic data generation tools (when they exist for non-text inputs)?

sjmonson: The idea is to decouple this field entirely from data generation and make it an implementation detail of the backends that need it. Each backend should implement it as it sees fit or completely ignore the field.

sjmonson: I don't understand the second part of this question. The data generation does not depend on `request_type`, the only place it was used (pre this PR) was in the last preprocess stage, which is now handled in the backend. If you need a field to distinguish between different ways of formatting the data then we could rename `--request-type` to something like `--query-format` (with an alias back to `--request-type`) and treat it as a special field that each backend can implement one or more of.

sjmonson: See https://github.com/vllm-project/guidellm/pull/478/commits/1cceb3534faef053af06fb7b898139c29e9cbdfd

src/guidellm/data/loaders.py
```diff
@@ -113,12 +116,11 @@ def generator(
                     ):
                         continue
 
+                    # Apply preprocessors in sequence
                     for preprocessor in self.preprocessors:
-                        # This can assign a GenerationRequest, which would then be
-                        # passed into the preprocessor, which is a type violation.
-                        # This should be fixed at some point.
-                        row = preprocessor(row)  # type: ignore[assignment]
```
jaredoconnell, line 120: I'm glad to see this type violation fixed.

src/guidellm/__main__.py
```diff
@@ -230,6 +223,19 @@ def benchmark():
...
+    callback=cli_tools.parse_json,
+    multiple=True,
+    help="JSON string of preprocessors to apply to the dataset.",
+)
+@click.option(
+    "--data-finalizer",
+    default=BenchmarkGenerativeTextArgs.get_default("data_finalizer"),
+    callback=cli_tools.parse_json,
+    help="JSON string of finalizer to convert dataset rows to requests.",
+)
```
jaredoconnell, line 256: These fields would benefit from some documentation in the docs files on how to format the JSON.

src/guidellm/settings.py
```diff
@@ -72,6 +72,9 @@ class DatasetSettings(BaseModel):
     preferred_data_splits: list[str] = Field(
         default_factory=lambda: ["test", "tst", "validation", "val", "train"]
     )
+    default_column_mapper: Literal["generative_column_mapper"] = (
+        "generative_column_mapper"
+    )
```
jaredoconnell: Let's use BenchmarkGenerativeTextArgs's default feature instead, as discussed in our DMs.

src/guidellm/backends/openai/request_formatter.py
```diff
@@ -0,0 +1,309 @@
...
+        :return: The formatted request arguments
+        """
+        ...
+
+
+class GenerationRequestFormatterFactory(
+    RegistryMixin[type[GenerationRequestFormatter]]
+):
+    """
+    Factory for registering and creating response handlers by backend type.
```
jaredoconnell: Is this registry intended to be used only in the OpenAI backend? If so, I think the docstring should specify that. If not, why is it in this file?

It looks to me like each backend will need one of these with the current design, if desired, to process `request_type`.

jaredoconnell (commented):
Looks good so far. I tested the default paths with a simple local benchmark, and that didn't break anything.

I have a few minor comments.

I'll give a final approving review once the tests are passing.

src/guidellm/data/loaders.py
sjmonson: Note to self: remove `# type: ignore[assignment]` to make sure its actually fixed :stuck_out_tongue:.

src/guidellm/backends/openai/request_formatter.py
sjmonson: Yeah the intent is that its a per-backend implementation detail. I don't think every backend needs to follow the same design pattern but they will all need a way to augment on `request_type` . Will do an internal docs pass since I think there are other areas where the docs are now out of date.

src/guidellm/data/loaders.py
sjmonson: Oh nvm, the diff confused me. It is removed.

src/guidellm/backends/openai/http.py
```diff
@@ -82,23 +112,38 @@ def __init__(
         self.model = model
         self.api_key = api_key
 
+        # Resolve request format
+        if request_format is None:
+            request_format = "/chat/completions"
```
jaredoconnell: My preference would be to locate the default in a variable alongside the other constants.

src/guidellm/data/preprocessors/encoders.py
```diff
@@ -0,0 +1,78 @@
...
+        return encode_image(*args, **kwargs)
+
+    @staticmethod
+    def encode_video(*args, **kwargs):
+        from guidellm.extras.vision import encode_video
+
+        return encode_video(*args, **kwargs)
+
+    def __call__(self, columns: dict[str, list[Any]]) -> dict[str, list[Any]]:
+        if columns.get("audio_column"):
```
jaredoconnell, line 49: Is there any chance these constants could get changed? It's not important but it may make sense to use constants. I see that they're also used in the finalizer file.

src/guidellm/backends/openai/http.py
```diff
@@ -18,10 +18,35 @@
...
+]
+
+
+DEFAULT_API_PATHS = {
+    "/health": "health",
+    "/models": "v1/models",
+    "/completions": "v1/completions",
+    "/chat/completions": "v1/chat/completions",
+    "/audio/transcriptions": "v1/audio/transcriptions",
+    "/audio/translations": "v1/audio/translations",
```
jaredoconnell: How would we want to handle a theoretical switch to v2? This design may not be ideal for that.

sjmonson: OpenAI seems to prefer adding new endpoints over versioning so I am not sure that will ever happen. I did consider this initially but wanted to avoid adding too many characters since the use has to type out the full name. However I can see an additional argument that its more clear for users.

src/guidellm/backends/openai/http.py
sjmonson: Done

src/guidellm/data/preprocessors/encoders.py
sjmonson: You mean the column names? I don't expect we will change them for anything less then a full redesign. The column mapper also supports custom columns which we would not be able to set constants for; so I would rather keep the design as-is to be an example for custom backends.

jaredoconnell: I won't block on this, but I would say that using constants by passing them in shouldn't mislead developers to think that constants are required. These specific ones seem like fairly constant names that could benefit from constants.

src/guidellm/__main__.py
```diff
@@ -179,20 +181,14 @@ def benchmark():
...
+    "--request-format",
     "--request-type",
-    default=BenchmarkGenerativeTextArgs.get_default("data_request_formatter"),
-    type=click.Choice(list(get_literal_vals(GenerativeRequestType))),
+    default=BenchmarkGenerativeTextArgs.get_default("request_format"),
     help=(
-        f"Request type to create for each data sample. "
-        f"Options: {', '.join(get_literal_vals(GenerativeRequestType))}."
+        "Format to use for requests. Options depend on backend. "
+        "If not provided, uses backend default."
```
jaredoconnell, line 189: Where can users get the documentation for this?

src/guidellm/__main__.py
```diff
@@ -228,7 +224,35 @@ def benchmark():
...
+    help="JSON string of arguments to pass to all preprocessors.",
+)
+@click.option(
+    "--data-finalizer",
+    default=BenchmarkGenerativeTextArgs.get_default("data_finalizer"),
+    callback=cli_tools.parse_json,
+    help=(
+        "JSON string of finalizer to convert dataset rows to requests."
+        " E.g., 'generative' or '{\"type\": \"generative\"}'"
+    ),
```
jaredoconnell, line 255: Where can users get exhaustive lists of these options?

jaredoconnell (approved):
I have no blocking comments. It looks good and I re-tested a simple benchmark.

src/guidellm/__main__.py
sjmonson: None, needs a followup.

src/guidellm/__main__.py
sjmonson: `guidellm benchmark run --help`? We do need better documentation across the board for CLI.

Issue #479, opened 2025-11-24
test_encode_audio_with_dict_input fail


tukwila:
**Describe the bug**

The following ut fail to test guidellm.extras.audio with dict input parameters:

```
import pytest
import torch
from guidellm.extras.audio import encode_audio, _decode_audio, _encode_audio

def test_encode_audio_with_dict_input():
        audio_dict = {
            "data": torch.randn(1, 16000),
            "sample_rate": 16000
        }
        
        result = encode_audio(audio=audio_dict)
        
        assert result["type"] == "audio_file"
        assert result["audio_bytes"] > 0
        assert result["audio_samples"] == 16000
        assert result["audio_seconds"] == 1.0


if __name__ == "__main__":
    pytest.main([__file__, "-v"])
```

save the ut into file such as:   test_dict_input.py,
then run pytest cmd: pytest test_dict_input.py

failure info:
```
____________________________________________________________________ test_encode_audio_with_dict_input ____________________________________________________________________

    def test_encode_audio_with_dict_input():
            audio_dict = {
                "data": torch.randn(1, 16000),
                "sample_rate": 16000
            }
    
>           result = encode_audio(audio=audio_dict)

tests/integration/test_dict_input.py:11: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
src/guidellm/extras/audio.py:58: in encode_audio
    samples = _decode_audio(audio, sample_rate=sample_rate, max_duration=max_duration)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

audio = {'data': tensor([[-0.1090, -0.5127,  0.6118,  ..., -0.2266,  0.4273,  1.2508]]), 'sample_rate': 16000}, sample_rate = 16000, max_duration = None

[...]

PR #480, opened 2025-11-24
fix encode_audio with dict input failure

tukwila:
## Details

- [ ]

https://github.com/vllm-project/guidellm/issues/479
## Test Plan

-

## Related Issues

- Resolves #

---

## Use of AI

### DIFF
--- a/src/guidellm/extras/audio.py
+++ b/src/guidellm/extras/audio.py
@@ -107,7 +107,11 @@ def _decode_audio(  # noqa: C901, PLR0912
             raise ValueError(
                 f"Audio dict must contain either 'data' or 'url' keys, got {audio}"
             )
-        audio_data = audio["data"] if "data" in audio else audio["url"]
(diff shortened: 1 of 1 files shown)


sjmonson:
I am not sure what this fix is accomplishing. The check on line 106 already makes sure either the data or url key exists.

src/guidellm/extras/audio.py
```diff
@@ -107,8 +107,9 @@ def _decode_audio(  # noqa: C901, PLR0912
             raise ValueError(
                 f"Audio dict must contain either 'data' or 'url' keys, got {audio}"
             )
+        audio_data = audio["data"] if "data" in audio else audio["url"]
```
sjmonson: Lets save a check here and get the field before the ValueError:

```python
audio_data = audio["data"] if "data" in audio else audio.get("url")
if audio_data is None:
    raise ValueError(
        f"Audio dict must contain either 'data' or 'url' keys, got {audio}"
    )
```

Issue #483, opened 2025-11-24
Switch from PDM to uv for advanced dependency management


sjmonson:
**Is your feature request related to a problem? Please describe.**
When we initially adopted PDM, `uv` was not capable of many of the advanced features we use such as multi-platform locks, pulling dependencies from non-standard repositories, and native `pylock.toml` support. Most of these problems have been solved and now function better in `uv` than PDM.

**Describe the solution you'd like**
Switch to `uv` in CI and container builds

**Additional context**
See [this guide](https://docs.astral.sh/uv/guides/integration/pytorch/#installing-pytorch) on managing the CPU version of pytorch which is one of our main reasons for using PDM in the build process.

PR #484, opened 2025-11-24
Split the lockfile between MacOS and Linux

### LINKED ISSUE #481: Pylock tooling does not explicitly lock platform
**Problem**

`pdm` is not properly configured to lock for each supported platform, thus the lockfile that is generated depends on the developer's machine.

**Solution**

Add explicit split-locks for each supported platform (Linux, MacOS) to `scripts/generate_pylock.sh` and regenerate the lockfile using these splits.

sjmonson:
## Summary

Adds platform specific targets to ensure the same lockfile is generated ~across platforms.~

**Edit:** This is seemingly impossible to do cross-platform in the pylock format as the wheels specifier is always for the last locked version. For now just return to a linux only split lock. In v0.5.0 will look into switch to `uv` (#483).

**Edit 2:** Drop the split-lock entirely for the same reason as above. Each append overrides "wheels".

**Edit 3:** Filled issue with pdm https://github.com/pdm-project/pdm/issues/3691. Will wait a bit to see if this gets resolved.

## Test Plan

CI tests running will indicate if Linux works, MacOS will have to be tested by someone with a Mac.

## Related Issues

- Resolves #481 

---

## Use of AI

### DIFF
--- a/pylock.toml
+++ b/pylock.toml
@@ -3,7 +3,10 @@
 lock-version = "1.0"
 requires-python = "<4.0,>=3.10.0"
 environments = [
-    "python_full_version >= \"3.10.0\" and python_version < \"4.0\" and sys_platform == \"linux\" and platform_system == \"Linux\" and os_name == \"posix\" and platform_machine == \"x86_64\"",
+    "python_version ~= \"3.12\" and sys_platform == \"linux\" and platform_system == \"Linux\" and os_name == \"posix\" and platform_machine == \"x86_64\"",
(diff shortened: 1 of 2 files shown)


jaredoconnell:
Would it make sense to just include a lockfile for each of these four scenarios? That may fix the issues we're getting.

sjmonson:
> Would it make sense to just include a lockfile for each of these four scenarios? That may fix the issues we're getting.

Yeah this will work, was just trying to avoid it because you will have to manually specify the lockfiles that are non-default named.

PR #491, opened 2025-12-01
feat(backend): add authentication support to OpenAIHTTPBackend

didlawowo:
## Summary

This PR adds authentication support to `OpenAIHTTPBackend`, enabling GuideLLM to work with OpenAI-compatible servers that require authentication on all endpoints (including `/health`).

### Problem

When using GuideLLM with LiteLLM proxy or similar OpenAI-compatible servers that require authentication, the benchmark fails during backend validation because the `/health` endpoint check doesn't include authentication headers:

```
RuntimeError: Backend validation request failed. Could not connect to the server or validate the backend configuration.
```

The `OpenAISettings` class already defines `api_key`, `bearer_token`, and `headers` fields, but these were never used in the actual HTTP requests.

### Solution

- Add `api_key`, `bearer_token`, and `headers` parameters to `OpenAIHTTPBackend.__init__()`
- Read default values from `settings.openai` when not explicitly provided
- Apply authentication headers to:
  - Backend validation requests (`/health` endpoint)
  - Available models requests (`/v1/models` endpoint)
  - All generation requests (merged with request-specific headers)

### Usage

```bash
# Via environment variables (recommended)
export GUIDELLM__OPENAI__API_KEY=sk-xxx
guidellm benchmark run --target http://litellm:4000 ...

# Via backend kwargs
guidellm benchmark run --target http://litellm:4000 \
  --backend-kwargs '{"api_key": "sk-xxx"}' ...
```

### Priority Order for Headers

1. Request-specific headers (highest priority)
2. Explicit `api_key`/`bearer_token`/`headers` parameters
3. Settings from `settings.openai` (lowest priority)

### Testing

- Tested with LiteLLM proxy requiring master key authentication
- `/health` endpoint now returns 200 instead of 401
- Benchmarks complete successfully against authenticated endpoints

### Related

This makes the documented example in the docstring actually work:
```python
backend = OpenAIHTTPBackend(
    target="http://localhost:8000",
    model="gpt-3.5-turbo",
    api_key="your-api-key"  # This param was in the example but didn't exist!
)
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,49 @@
+# Python
+__pycache__/
+*.py[cod]
+*$py.class
+*.so
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+
+# Virtual environments
+.venv
+.venv_temp
+venv/
+ENV/
+env/
+
+# IDEs
+.vscode/
+.idea/
+*.swp
+*.swo
+*~
+
+# Git
+.git/
+.gitignore
+
+# Dataset cache (will be copied separately)
+dataset_cache/
+
+# Logs
+*.log
+
+# OS
+.DS_Store
+Thumbs.db
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,10 @@
 # GuideLLM with uv, CUDA 13, and embedded dataset
 FROM nvidia/cuda:13.0.0-base-ubuntu22.04
 
+# Configure timezone non-interactively
+ENV DEBIAN_FRONTEND=noninteractive
+ENV TZ=UTC
+
 # Install system dependencies
 RUN apt-get update && apt-get install -y --no-install-recommends \
     git \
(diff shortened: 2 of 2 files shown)


src/guidellm/backends/openai.py
```diff
@@ -79,6 +85,29 @@ def __init__(
         self.target = target.rstrip("/").removesuffix("/v1")
         self.model = model
 
+        # Build default headers with authentication
+        from guidellm.settings import settings
```
sjmonson, line 89: Don't late import. This should be in import section at top.

src/guidellm/backends/openai.py
```diff
@@ -79,6 +85,29 @@ def __init__(
...
+        if settings.openai.headers:
+            self._default_headers.update(settings.openai.headers)
+
+        # Add explicit headers parameter (medium priority)
+        if headers:
+            self._default_headers.update(headers)
+
+        # Resolve API key (highest priority): explicit param > settings
+        resolved_api_key = api_key or settings.openai.api_key
+        resolved_bearer_token = bearer_token or settings.openai.bearer_token
```
sjmonson, line 103: Can you actually remove all references to bearer_token. I don't think we need both.

.dockerignore
sjmonson, line 1: Drop this we have an OCI compliant `.containerignore`:

https://github.com/vllm-project/guidellm/blob/main/.containerignore

Dockerfile
sjmonson, line 1: Drop this we have an OCI compliant `Containerfile`:

https://github.com/vllm-project/guidellm/blob/main/Containerfile

src/guidellm/data/deserializers/huggingface.py
sjmonson, line 1: Can you move these changes to a new PR? We have had some internal discussion on making arguments to `--data` more explicit.

PR #509, opened 2025-12-08
Container Build Caching

sjmonson:
## Summary

Store `dnf` and `uv` caches in a GitHub Action cache.

## Details

Adds mount caches for `dnf` and `uv` to container image and stores their output into a action cache. Layer caching would probably be more efficient but requires a container repository for storing the intermediate layers.

## Test Plan

Rerun the container build job and observe that build takes less time and skips download steps.

---

## Use of AI

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -142,6 +142,12 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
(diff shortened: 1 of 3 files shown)


jaredoconnell (approved):
The change to use GUIDELLM_BUILD_EXTRAS alone is helpful. I didn't see any problems with the workflow changes.

PR #519, opened 2025-12-18
Record output_tokens for incomplete requests

### LINKED ISSUE #514: Up to 9% decrease in output tokens per second between v0.3 and v0.4
## **Describe the bug**

The measured output token throughput in GuideLLM v0.4.0 and later is significantly lower despite all other metrics being near identical.

## **Expected behavior**

Output token throughput should not differ from v0.3.x when all else is equal.

## **Environment**
1. OS [e.g. Ubuntu 20.04]: Fedora 43 container on OCP 4.19
2. Python version [e.g. 3.12.2]: 3.13.9
3. GuideLLM version: 0.4.0 and 0.5.0

## **To Reproduce**

Run the following GuideLLM test on v0.3.0 and v0.4.0, then compare metrics in output JSON:

```sh
# Ensure we use the same endpoint across versions
export GUIDELLM_REQUEST_TYPE=text_completions
export GUIDELLM_TARGET=http://localhost:8080

guidellm benchmark run \
            --target="${GUIDELLM_TARGET}" \
            --model=RedHatAI/Llama-3.3-70B-Instruct-FP8-dynamic \
            --processor=RedHatAI/Llama-3.3-70B-Instruct-FP8-dynamic \
            --rate-type=concurrent \
            --data=prompt_tokens=1000,output_tokens=1000 \
            --max-seconds=600 \
            --rate=1,50,100,200,300,500,650
```

### Total Output Tokens per second over intended concurrency

Image

### Percent Change from v0.3.0 baseline

Image

sjmonson:
## Summary

Sets `continuous_usage_stats` to get token usage on incomplete requests. If usage is still unavailable fall back to iteration count.

## Details

In v0.3.0 and earlier the number of iterations was used as proxy for output token count in incomplete requests that did not return usage metrics. In v0.4.0 this behavior was removed which lead to large discrepancies in output token count based on the percentage of the benchmark consisting of incomplete requests.

This PR restore the original behavior of falling back to number of iterations. Additionally it sets the `continuous_usage_stats` flag to enable usage metrics on every iteration, when available.

## Test Plan

- Run a long-generation, high concurrency benchmark using a `max-seconds` constraint. For incomplete requests check that output_tokens is greater than 0 for some requests.

## Related Issues

- Resolves #514 

---

## Use of AI

### DIFF
--- a/tests/unit/schemas/test_request.py
+++ b/tests/unit/schemas/test_request.py
@@ -397,6 +397,7 @@ class TestUsageMetrics:
         ("metrics_kwargs", "expected_total"),
         [
             ({}, None),
+            ({"text_tokens": 0}, 0),
             ({"text_tokens": 100}, 100),
             ({"text_tokens": 100, "image_tokens": 50}, 150),
             (
@@ -412,6 +413,7 @@ class TestUsageMetrics:
         ],
         ids=[
             "no_tokens",
+            "zero_tokens",
             "text_only",
             "text_and_image",
             "all_modalities",
--- a/tests/unit/schemas/test_request_stats.py
+++ b/tests/unit/schemas/test_request_stats.py
@@ -24,6 +24,7 @@ class TestGenerativeRequestStats:
             "short_completion",
             "long_chat",
             "single_token_output",
+            "no_usage_fallback",
             "no_iterations_fallback",
         ],
     )
@@ -42,8 +43,8 @@ class TestGenerativeRequestStats:
         if case_id == "short_completion":
             # Quick completion with moderate tokens
             request_type = "text_completions"
-            prompt_tokens = 30
-            output_tokens = 20
+            usage_prompt_tokens = prompt_tokens = 30
+            usage_output_tokens = output_tokens = 20
             request_start = 0.0
             # Time to first token: ~200ms (typical for small models)
             ttft_ms = rng.uniform(150, 250)
@@ -58,8 +59,8 @@ class TestGenerativeRequestStats:
         elif case_id == "long_chat":
             # Long conversation with many tokens
             request_type = "chat_completions"
-            prompt_tokens = 400
-            output_tokens = 256
+            usage_prompt_tokens = prompt_tokens = 400
+            usage_output_tokens = output_tokens = 256
             request_start = 10.0
             # Longer TTFT for larger context
             ttft_ms = rng.uniform(2000, 3000)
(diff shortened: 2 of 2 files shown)


src/guidellm/schemas/request_stats.py
```diff
@@ -123,6 +123,11 @@ def output_tokens(self) -> int | None:
         """
         :return: Number of tokens in the generated output, or None if unavailable
         """
+        # Fallback if we did not get usage metrics from the server
+        # NOTE: This assumes each iteration is one token
+        if self.output_metrics.total_tokens is None:
+            return self.info.timings.token_iterations or None
```
dbutenhof, line 129: OK -- I get that this is restoring old/convenient behavior: but (again, question rather than criticism), why it is expected that we might not get token counts back, and why is assuming we (would have) got[ten] one for each iteration a fair fallback rather than reporting an unexpected condition?

Seems like this could be potentially confusing and (I assume this is what Jared meant in an earlier comment), could this be an option (maybe even "don't assume iteration count" rather than "assume iteration count" if it's rare that someone would care)?

dbutenhof (dismissed):

src/guidellm/schemas/request_stats.py
jaredoconnell: The main setting I meant was to include tokens for incomplete requests.

Would it make sense to also make this fallback a setting?

sjmonson: > OK -- I get that this is restoring old/convenient behavior: but (again, question rather than criticism), why it is expected that we might not get token counts back, and why is assuming we (would have) got[ten] one for each iteration a fair fallback rather than reporting an unexpected condition?

One of the challenging things about working with the OpenAI API is that its not actually a standard. Its the API OpenAI came up with for their models and everyone else copied it to be compatible with existing clients. some model servers are less compliant then others so we need a fallback for the ones that don't report token usage. 

This issue is a blocker for internal use of GuideLLM >=v0.4.0 so I was thinking we restore basic functionality here and maybe focus on a better solution in the next release.

> The main setting I meant was to include tokens for incomplete requests.

Probably does not make sense to go back to every incomplete request being counted as 0 tokens. If a benchmark has a lot of incomplete requests it will cause the total throughput to be super inaccurate due to the giant hole it leaves in the data.

dbutenhof: Yeah, maybe someone has a use case for dropping incomplete request counts; but it seems a bit odd and it's probably best to wait for such a use case to be defined.

For the fallback, though -- if I were looking for metrics, and I might get a "lie" because the server doesn't report actual token counts, I suspect I'd rather know it was happening. In other places I've seen the code attach errors to the output -- this may not quite be considered "an error", but would it make sense to report that the server didn't produce token counts and GuideLLM made it up? That might even be better than an option to disable the fallback ... because "0" is unlikely to be right, either. (Though if it were a floating point count I might even whimsically suggest a NAN value...)

PR #525, opened 2026-01-12
Make image_url/video_url send dictionaries

Vinno97:
## Summary

This PR changes how GuideLLM sends image/video urls to the backend using the chat_completions endpoint. The current behaviour doesn't seem to follow spec, which was giving me trouble benchmarking a multimodal dataset.

Both vLLM and the OpenAI spec expect message content of type `image_url` to be a dictionary like `{"url": <url>}`, though the previous behaviour sent the url as a plain string. `video_url` is vLLM-only as far as I know though it follows the same schema.

## Details

- [ ] Updated the chat_completions request to match the spec
   * Old behaviour:
      ```
          {
              "role": "user",
              "content": [
                  {
                      "type": "image_url",
                      "image_url": <image_url>
                  }
              ]
          }
      ```
    * New behaviour:
         ```
                {
                    "role": "user",
                    "content": [
                        {
                            "type": "image_url",
                            "image_url": {"url": <image_url>}
                        }
                    ]
                }
         ```

## Test Plan

-

## Related Issues

- Resolves #

---

## Use of AI

### DIFF
--- a/src/guidellm/data/preprocessors/formatters.py
+++ b/src/guidellm/data/preprocessors/formatters.py
@@ -234,7 +234,10 @@ class GenerativeChatCompletionsRequestFormatter(RequestFormatter):
                 {
                     "role": "user",
                     "content": [
-                        {"type": "image_url", "image_url": image_dict.get("image")}
+                        {
+                            "type": "image_url",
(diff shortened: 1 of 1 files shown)


sjmonson (approved):
LGTM

PR #528, opened 2026-01-13
Fix strategy initialization deadlock

### LINKED ISSUE #526: Some benchmark runs have a long delay before starting
## Describe the bug

Occasionally benchmarks have a long delay between "Start time" and sending the first request. In a time constrained benchmark this can lead to massive difference in the number of requests sent.

## Expected behavior

Requests should start sending nearly instantly after the benchmark starts.

## Environment

1. OS [e.g. Ubuntu 20.04]: 
2. Python version [e.g. 3.12.2]: 3.13.9
3. GuideLLM version: v0.4.0 and later

## To Reproduce

Unknown, occurs often on certain hardware in stress testing.

### Request concurrency for various GuideLLM versions at `rate = 100` and `max-seconds = 600`

Image

### First and last request time (reqs) compared to benchmark start and end time (sched)

Image

sjmonson:
## Summary

Fixed an issue where applying the initial start time to a `Stratagy` could results in a deadlock. Also ensure that the recorded benchmark `start_time` is relative to the first request and not the scheduling start time.

## Details

At the start of a benchmark the main thread will set the start time in shared memory. Each worker will also attempt to poll the start time. This behavior can cause transient deadlocks that prevent the benchmark from progressing to sending requests. Additionally the recorded "start_time" used for constraint calculations and stored as the benchmark start time was relative to the scheduler start time and not the first request time. This PR implements separate fixes for each issue.

## Test Plan

The following patch is needed to observe the issue however it can be very difficult to reproduce depending on the hardware:

```diff
diff --git a/src/guidellm/scheduler/worker_group.py b/src/guidellm/scheduler/worker_group.py
index a2ee246f..4af4f710 100644
--- a/src/guidellm/scheduler/worker_group.py
+++ b/src/guidellm/scheduler/worker_group.py
@@ -22,7 +22,7 @@ from multiprocessing.process import BaseProcess
 from multiprocessing.synchronize import Barrier, Event
 from typing import Generic, NamedTuple
 
-from guidellm.logger import logger
+from loguru import logger
 from guidellm.scheduler.constraints import Constraint, RequestsExhaustedConstraint
 from guidellm.scheduler.schemas import (
     BackendInterface,
@@ -278,7 +278,11 @@ class WorkerProcessGroup(Generic[RequestT, ResponseT]):
         ):
             raise RuntimeError("create_processes() must be called before start()")
 
+        logger.critical(f"Starting in {start_time - time.time():.2f} seconds...")
+        initial_time = time.time()
         self.strategy.init_processes_start(start_time=start_time)
+        init_time = time.time()
+        logger.critical(f"Initialized strategy in {init_time - initial_time:.2f} seconds.")
         stop_send_requests_event = threading.Event()
         send_requests_stopped_event = threading.Event()
         self.state = WorkerGroupState[RequestT, ResponseT](
@@ -300,6 +304,8 @@ class WorkerProcessGroup(Generic[RequestT, ResponseT]):
             send_stop_criteria=[stop_send_requests_event],
             receive_stop_criteria=[self.shutdown_event],
         )
+        logger.critical(f"Started messaging in {time.time() - init_time:.2f} seconds.")
+        logger.critical(f"Finished startup in {time.time() - initial_time:.2f} seconds.")
 
         if (wait_time := start_time - time.time()) > 0:
             await asyncio.sleep(wait_time)

```

The following test would reliably reproduce the issue in at least one concurrency level on H200 hardware:

```sh
guidellm benchmark run \
            --target="${GUIDELLM_TARGET}" \
            --rate-type=concurrent \
            --data=prompt_tokens=1000,output_tokens=1000 \
            --max-seconds=600 \
            --rate=1,50,100,200,300,500,650 \
            --disable-progress \
            --output-path="output.json" 2>&1 | tee "output.log"
```

Example in `output.log` where strategy sync hit issue:

```
26-01-13 21:16:33|CRITICAL         |guidellm.scheduler.worker_group:start:281 - Starting in 1.00 seconds...
26-01-13 21:19:53|CRITICAL         |guidellm.scheduler.worker_group:start:285 - Initialized strategy in 200.15 seconds.
26-01-13 21:19:53|CRITICAL         |guidellm.scheduler.worker_group:start:307 - Started messaging in 0.00 seconds.
26-01-13 21:19:53|CRITICAL         |guidellm.scheduler.worker_group:start:308 - Finished startup in 200.15 seconds.
```

## Related Issues

- Resolves #526 

---

## Use of AI

### DIFF
--- a/src/guidellm/benchmark/schemas/generative/accumulator.py
+++ b/src/guidellm/benchmark/schemas/generative/accumulator.py
@@ -168,7 +168,7 @@ class GenerativeBenchmarkTimings(StandardBaseModel):
         :param config: Benchmark configuration with warmup/cooldown settings
         """
         # First update non terminal timestamps
-        self.request_start = scheduler_state.start_time
+        self.request_start = scheduler_state.start_requests_time
         self.last_update = self.current_update
         if (current_time := info.timings.last_reported) is not None:
             self.current_update = (
--- a/src/guidellm/scheduler/constraints/request.py
+++ b/src/guidellm/scheduler/constraints/request.py
@@ -221,11 +221,12 @@ class MaxDurationConstraint(PydanticConstraintInitializer):
             else self.max_duration[min(current_index, len(self.max_duration) - 1)]
         )
 
+        start_time = state.start_requests_time or state.start_time
         current_time = time.time()
-        elapsed = current_time - state.start_time
+        elapsed = current_time - start_time
         duration_exceeded = elapsed >= max_duration
         remaining_duration = min(max(0.0, max_duration - elapsed), max_duration)
-        stop_time = None if not duration_exceeded else state.start_time + max_duration
+        stop_time = None if not duration_exceeded else start_time + max_duration
 
         return SchedulerUpdateAction(
             request_queuing="stop" if duration_exceeded else "continue",
@@ -234,7 +235,7 @@ class MaxDurationConstraint(PydanticConstraintInitializer):
                 "max_duration": max_duration,
                 "elapsed_time": elapsed,
                 "duration_exceeded": duration_exceeded,
-                "start_time": state.start_time,
+                "start_time": start_time,
                 "current_time": current_time,
                 "stop_time": stop_time,
             },
--- a/src/guidellm/scheduler/schemas.py
+++ b/src/guidellm/scheduler/schemas.py
@@ -301,6 +301,12 @@ class SchedulerState(StandardBaseModel):
     end_time: float | None = Field(
         default=None, description="Unix timestamp when the scheduler stopped"
     )
+    start_requests_time: float | None = Field(
+        default=None, description="Unix timestamp of the first sent request"
+    )
+    end_requests_time: float | None = Field(
+        default=None, description="Unix timestamp of the last finalized request"
+    )
     end_queuing_time: float | None = Field(
         default=None, description="Unix timestamp when request queuing stopped"
     )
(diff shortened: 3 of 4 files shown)


jaredoconnell (dismissed):
Using an event looks to be an improvement to the design. I just have one comment.

src/guidellm/scheduler/strategies.py
```diff
@@ -577,15 +572,12 @@ def init_processes_start(self, start_time: float):
...
-        if self._processes_lock is None:
-            raise RuntimeError("_processes_lock is None in init_processes_start")
         if self._offset is None:
             raise RuntimeError(
                 "_offset is None in init_processes_start; was "
                 "init_processes_timings not called?"
             )
-        with self._processes_lock:
-            self._offset.value = start_time
+        self._offset.value = start_time
```
jaredoconnell: Are we sure it's safe to set the value without the lock?
I do see that later on, the lock is used to increment this value.

sjmonson: [multiprocessing.Value](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Value) will automatically lock for assignments. The only times it needs to be explicitly locked is multiple or non-atomic operations. I did consider explicitly locking here anyways as its not obvious at first glance how this works. Since you pointed it out I'll make the change.

jaredoconnell: Sounds good. Either that or a comment.

jaredoconnell (approved):
Looks good to me. I tested it with a basic setup and checked the relevant timing values in the output.

PR #531, opened 2026-01-15
Fix some outdated docs examples

sjmonson:
## Summary

Updates some outdated examples

---

## Use of AI

### DIFF
--- a/docs/guides/datasets.md
+++ b/docs/guides/datasets.md
@@ -1,5 +1,7 @@
 # Datasets
 
+> [!WARNING] 🚧 This documentation is in the process of being updated. Some information may be outdated. 🚧
+
 GuideLLM supports various dataset configurations to enable benchmarking and evaluation of large language models (LLMs). This document provides a comprehensive guide to configuring datasets for different use cases, along with detailed examples and rationale for choosing specific pathways.
 
 ## Data Arguments Overview


docs/guides/datasets.md
```diff
@@ -1,5 +1,8 @@
 # Datasets
 
+> [!WARNING]
+> 🚧 This documentation may contain outdated information 🚧
+
```
sjmonson: This whole file needs work. Will revisit later.

dbutenhof: Probably always a good warning, but way too generic to be particularly valuable in this case!

We might even want something a bit more specific like "references to URLs and names (e.g., of datasets and models) that are outside the GuideLLM documentation are subject to change--always be cautious when applying examples!" ... but I think we should be better about diagnosing this sort of problem cleanly before someone has to raise the issue with us.

README.md
```diff
@@ -182,16 +182,18 @@ guidellm benchmark \
 GuideLLM supports HuggingFace datasets, local files, and synthetic data. This example loads the CNN DailyMail dataset from HuggingFace and maps the article column to prompts while using the summary token count column to determine output lengths.
 
 ```bash
-guidellm benchmark \
+guidellm benchmark run \
-  --data "hf:cnn_dailymail" \
-  --data-args '{"prompt_column":"article","output_tokens_count_column":"summary_tokens"}'
+  --data "abisee/cnn_dailymail" \
```
dbutenhof, line 187: So the "hf:" prefix is ... unnecessary, obsolete? 🤔

dbutenhof (approved):
Lots more improvement than just the trigger issue; great.

README.md
sjmonson: Yeah, I don't think it was every necessary; it just used to be available as a hint.

docs/guides/datasets.md
sjmonson: The the first half of this document need to be rewritten as it contains a lot of dated explanations. I did not feel like taking that task on today so this is an indicator that work needs to be done.

sjmonson: Reworded a bit to be clearer

dbutenhof (approved):
I won't quibble with the new warning (although "in the process of" is really a redundant phrase ... and not entirely accurate as I think the implied rewrite is future rather than current work 😆) ... 

This looks like an improvement, and the warning is possibly a decent short term disclaimer although I hesitate to have our documentation start out insisting that readers not trust it! (We should have an issue laying out details of the revision necessary scheduled at least for 0.6 if not 0.5.2 to limit the impact.)

Issue #97, opened 2025-03-11
Define effective hierarchy of debug logging


sjmonson:
The INFO log level is way to verbose about the inner config of objects. We should push many of the messages to higher log levels and/or rewrite some messages to be less verbose.

Here is a proposal for what we could focus on in each log level:

* `ERROR`: Anything that leads to complete failure of the test
* `WARNING`: Things might be going wrong, or not happening in the expected way
    * Requests erroring
    * Concurrency saturated
* `INFO`: Things that should be happening that help explain the state
    * Info about run configuration, dataset...
    * What RPS settings are calculated in the sweep
    * Infrequent status updates
    * Info about output calculations / files that its dumped to
* `DEBUG`: Stuff like prompts and results
* `TRACE`: token-level events from the requests

PR #534, opened 2026-01-15
Added option to log errors from backends

jaredoconnell:
## Summary

I've found that it would be a lot more convenient to see the backend's errors in the logs for several tasks I've done.

This simply adds an environment variable to enable this logging.

## Details

- Intended only for development
- This should not add any measurable overhead, because the logging only happens when the environment variable is true.
- It prints a full stacktrace to the console, ensuring all info is available to the developer
- I've used this for two of my tasks so far, and I've found it useful.

## Test Plan

Run a benchmark with failing requests with the environment variable set to true. The errors should print out to the console.

---

## Use of AI

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,16 +182,18 @@ guidellm benchmark \
 GuideLLM supports HuggingFace datasets, local files, and synthetic data. This example loads the CNN DailyMail dataset from HuggingFace and maps the article column to prompts while using the summary token count column to determine output lengths.
 
 ```bash
-guidellm benchmark \
+guidellm benchmark run \
   --target http://localhost:8000 \
-  --data "hf:cnn_dailymail" \
-  --data-args '{"prompt_column":"article","output_tokens_count_column":"summary_tokens"}'
+  --data "abisee/cnn_dailymail" \
+  --data-args '{"name": "3.0.0"}' \
+  --data-column-mapper '{"text_column":"article"}'
 ```
 
 **Key parameters:**
 
 - `--data`: Data source specification - accepts HuggingFace dataset IDs (prefix with `hf:`), local file paths (`.json`, `.csv`, `.jsonl`, `.txt`), or synthetic data configs (JSON object or `key=value` pairs like `prompt_tokens=256,output_tokens=128`)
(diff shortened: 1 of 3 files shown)


src/guidellm/scheduler/worker.py
```diff
@@ -381,11 +383,16 @@ async def _process_next_request(self, target_start: float):
...
             if request is not None and request_info is not None:
-                request_info.error = repr(exc)
+                request_info.error = repr(e)
                 request_info.traceback = traceback.format_exc()
                 request_info.timings.resolve_end = time.time()
+                # Log backend exception if enabled
+                if settings.logging.log_backend_exceptions:
+                    logger.exception(
+                        f"Backend exception for request {request_info.request_id}: {e}"
+                    )
```
sjmonson: Just add this as a debug log. The only gate for log messages should be log level. Also see #97.

```suggestion
                logger.debug(
                    f"Backend exception for request {request_info.request_id}",
                    exc_info=e,
                )
```

jaredoconnell: There would probably be a small performance penalty for doing this. I can use https://loguru.readthedocs.io/en/stable/api/logger.html#loguru._logger.Logger.opt to reduce overhead. This line isn't inherently expensive, but it's executed so many times in a benchmark.

sjmonson: It will only be executed once per failed requests. We can time it to make sure but I think we should only start worrying about logging performance impact if its more then once per request or a significant amount of data such as when we used to log the entire request/response. Additionally this is DEBUG so it should be close to a noop under defaults.

jaredoconnell: I did some testing, and can confirm that there is no measurable overhead when log level is higher than Debug, so I'll switch to that. When it's printing the errors, there is substantial overhead from the logging itself, with throughput dropping to around 5% with Debug mode compared to Info.

Since this will only apply when debug logging is enabled and requests are failing, I'm okay with moving forward without a dedicated option unless/until someone finds a specific use case where they need high performance with debug logging enabled while requests are failing.

The suggestion you have here doesn't print the stacktrace. I'm going to push a commit that reverts the prior proposal and logs it as debug with the stacktrace.

sjmonson: SGTM

PR #461, opened 2025-11-13
Update docs for 0.4.0 release including README and docs folder updates

markurtz:
## Summary

Extensive updates for docs for 0.4.0 release centered around the capabilities added from the refactor and expected usage. 

## Test Plan

N/A

---

- [ X] "I certify that all code in this PR is my own, except as noted below."

## Use of AI

- [ X] Includes AI-assisted code completion
- [ ] Includes code generated by an AI application
- [ ] Includes AI-generated tests (NOTE: AI written tests should have a docstring that includes `## WRITTEN BY AI ##`)

PR #536, opened 2026-01-19
Add documentation how to use with llama.cpp

rgerganov:
## Summary

Add documentation on how to run guidellm with llama.cpp server

NOTE: I have added this with PR #318 and then it was, I believe accidentally, removed with PR #461. I have also verified the steps still work fine with the latest guidellm version.

## Details

guidellm can run against a llama.cpp server when the model metadata is prefetched and the server is started with the right arguments.

## Test Plan

Verified that guidellm runs successfully when following the documented steps

## Related Issues

n/a
---

## Use of AI


sjmonson (approved):
LGTM

PR #569, opened 2026-01-30
Replace various key=value string parsers with a single utility

sjmonson:
## Summary

Replaces various config string handlers with a new `ArgString` format. 

## Details

Replaces the dataset config parser with a more advance "arg_string" parser with support for nested and list config structures. For example:

```
prefix_buckets[0].bucket_weight=10,prefix_buckets[0].prefix_count=10,prefix_buckets[1].bucket_weight=20,prefix_buckets[1].prefix_count=4
```

Renders to

```python
{
    "prefix_buckets": [
        {"bucket_weight": 10, "prefix_count": 10},
        {"bucket_weight": 20, "prefix_count": 4},
     ]
}
```

Also added some glue logic to allow this format anywhere on the CLI that supports JSON. This code should be considered temporary as it will mostly be replaced through the rest of the CLI refactor.

Also open to suggestions for a better name then "arg_string"; a 2-4 letter acronym like JSON or YAML would be good.

## Test Plan

The following benchmark should run without errors and have ~384 ISL per request (make sure to check startup logs confirm `backend-kwargs` and `warmup` are configured):

```sh
guidellm benchmark run \
    --profile concurrent \
    --data "prompt_tokens=256,output_tokens=256,prefix_buckets[0].bucket_weight=10,prefix_buckets[0].prefix_tokens=128,prefix_buckets[0].prefix_count=10" \
    --request-format /v1/completions \
    --backend-kwargs '{"timeout":120}' \
    --warmup percent=0.1,value=20 \
    --max-seconds 30 \
    --rate 30
```

---

## Use of AI

### DIFF
--- a/src/guidellm/data/config.py
+++ b/src/guidellm/data/config.py
@@ -7,6 +7,7 @@ import yaml
 from pydantic import ValidationError
 
 from guidellm.data.schemas import DataConfig, DataNotSupportedError
+from guidellm.utils import arg_string
 
 ConfigT = TypeVar("ConfigT", bound=DataConfig)
 
@@ -85,37 +86,24 @@ def _load_config_str(data: str, config_class: type[ConfigT]) -> ConfigT | None:
     if not isinstance(data, str):
         return None
 
-    data_str = data.strip()
-    error = None
-
-    if (data_str.startswith("{") and data_str.endswith("}")) or (
-        data_str.startswith("[") and data_str.endswith("]")
-    ):
-        try:
-            return config_class.model_validate_json(data_str)
-        except Exception as err:  # noqa: BLE001
-            error = err
-
-    if data_str.count("=") >= 1:
-        # key=value pairs separated by commas
-        try:
-            config_dict = {}
-            items = data_str.split(",")
-            for item in items:
-                key, value = item.split("=")
-                config_dict[key.strip()] = (
-                    int(value.strip()) if value.strip().isnumeric() else value.strip()
-                )
-
-            return config_class.model_validate(config_dict)
-        except Exception as err:  # noqa: BLE001
-            error = err
-
     err_message = (
         f"Unsupported string data for {config_class.__name__}, "
         f"expected JSON or key-value pairs, got {data}"
     )
-    if error is not None:
-        err_message += f" with error: {error}"
-        raise DataNotSupportedError(err_message) from error
-    raise DataNotSupportedError(err_message)
+
+    try:
+        data_parsed = yaml.safe_load(data)
+    except yaml.YAMLError:
+        data_parsed = data
+
+    # If no change from YAML parsing, try arg_string parsing
+    if data_parsed == data:
+        try:
+            data_parsed = arg_string.loads(data_parsed)
+        except arg_string.ArgStringParseError as e:
+            raise DataNotSupportedError(err_message) from e
+
+    try:
+        return config_class.model_validate(data_parsed)
+    except ValidationError as err:
+        raise DataNotSupportedError(err_message) from err
--- a/src/guidellm/utils/arg_string.py
+++ b/src/guidellm/utils/arg_string.py
@@ -57,6 +57,8 @@ class ArgStringParser:
         skip_invalid: bool = False,
         fill_value: Callable[[], Any] | None = None,
         allow_overwrite: bool = False,
+        key_delimiter: str = ".",
+        split_delimiter: str = ",",
     ):
         """
         Initialize the argument string parser.
@@ -70,6 +72,8 @@ class ArgStringParser:
         self.skip_invalid = skip_invalid
         self.fill_value = fill_value() if fill_value is not None else None
         self.allow_overwrite = allow_overwrite
+        self.key_delimiter = key_delimiter
+        self.split_delimiter = split_delimiter
 
     def decode(self, s: str) -> dict[str, Any]:
         """
@@ -110,7 +114,7 @@ class ArgStringParser:
             return result
 
         # Split by commas
-        pairs = s.split(",")
+        pairs = s.split(self.split_delimiter)
 
         for pair_raw in pairs:
             pair_stripped = pair_raw.strip()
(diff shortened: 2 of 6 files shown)


src/guidellm/data/config.py
```diff
@@ -85,37 +86,24 @@ def _load_config_str(data: str, config_class: type[ConfigT]) -> ConfigT | None:
...
         f"Unsupported string data for {config_class.__name__}, "
         f"expected JSON or key-value pairs, got {data}"
     )
-    if error is not None:
-        err_message += f" with error: {error}"
-        raise DataNotSupportedError(err_message) from error
-    raise DataNotSupportedError(err_message)
+
+    try:
+        data_parsed = yaml.safe_load(data)
```
dbutenhof, line 95: Since you do this first everywhere you call `arg_string.loads`, why not bring that inside `loads`? Since the purpose is parsing nested dictionaries, I don't offhand see a plausible use where that would be inappropriate, and it would avoid the duplication.

src/guidellm/utils/arg_string.py
```diff
@@ -0,0 +1,441 @@
...
+        # Split by commas
+        pairs = s.split(self.split_delimiter)
+
+        for pair_raw in pairs:
+            pair_stripped = pair_raw.strip()
+            # Handle invalid pairs
+            if not pair_stripped or "=" not in pair_stripped:
+                if self.skip_invalid:
+                    continue
+                raise ArgStringParseError(f"Invalid key=value pair: '{pair_raw}'")
```
dbutenhof: I'd avoid the "pair" terminology, since that's not really a user-facing concept. Maybe "expression" or "term"?

src/guidellm/utils/arg_string.py
```diff
@@ -0,0 +1,441 @@
...
+
+            key, value = pair_stripped.split("=", 1)
+            key = key.strip()
+            value = value.strip()
+
+            # Handle empty keys
+            if not key:
+                if self.skip_invalid:
+                    continue
+                raise ArgStringParseError(f"Empty key in pair: '{pair_raw}'")
```
dbutenhof: Maybe "Empty key in key=value term"? A reminder of what's expected, and avoiding the rather loose "pair" terminology.

src/guidellm/utils/arg_string.py
```diff
@@ -0,0 +1,441 @@
...
+        :return: List of (name, index) tuples representing the parsed key segments
+        """
+        segments: list[tuple[str, int | None]] = []
+
+        # Split by dots first
+        parts = key.split(self.key_delimiter)
+
+        for part in parts:
+            # Check if part has array notation: name[index]
+            match = re.match(r"^([^\[]+)\[(\d+)\]$", part)
```
dbutenhof, line 224: You're pretty liberal in your allowance for whitespace, so a `strip()` here would probably be in order.

dbutenhof (commented):
Minor comments from a first pass without getting deep into the algorithms ... but it looks good.

src/guidellm/data/config.py
sjmonson: The main reason why is currently we have a CLI fields like `--data-column-mapper` which can be a string name or a dict, so the error handling needs to be different (See 5402c02aca149aa6fa2df6879dcf4d5dc15d3df3). Once we do away with that bad pattern we can make this change.

sjmonson: I.e. currently the default for column mapper is `generative_column_mapper` rather then `type=generative_column_mapper`

src/guidellm/utils/arg_string.py
sjmonson: Something else I just noticed, if the condition is `if not pair_stripped` then this error reads:

`"Invalid key=value pair: ''"`

So maybe something like: `Invalid expression at 16 in 'config=truncated,=to_last_keypair'`?

src/guidellm/utils/arg_string.py
dbutenhof: The change is good -- but the unit tests are feeling left out. 😢

src/guidellm/utils/arg_string.py
dbutenhof: Not sure I'm thrilled about the newlines in the error strings: if they get logged this might confuse readers (maybe including conmon?) -- but I won't object.

dbutenhof (dismissed):

src/guidellm/utils/arg_string.py
sjmonson: I'll drop the newline. I originally had planned on writing some logic to point out where in the string the error occurred and a newline made that easier, but dropped that code since it was overly complex.

jaredoconnell (dismissed):
Looks good to me. And basic data args are working for me.

jaredoconnell:
A test is failing. It looks like maybe you changed the error message under that condition.

src/guidellm/utils/arg_string.py
```diff
@@ -0,0 +1,446 @@
...
+
+            # Convert value to appropriate type
+            parsed_value = yaml.safe_load(value)
+
+            # Set the value in the nested structure
+            self._value_set_nested(result, segments, parsed_value)
+
+        return result
+
+    def set(self, obj: dict, path: str, value: Any) -> None:
```
dbutenhof, line 152: You don't seem to use this public method internally, nor is it tested by the unit test module.

While I kinda like the generality of the public get/set methods here, they should either be tested or removed.

dbutenhof (approved):
This looks great. Minor issue: you added the external get/set methods in the arg string parser class, which aren't used, and aren't tested. If we're going to have them, they should be tested.

src/guidellm/utils/arg_string.py
sjmonson: Sorry forgot I committed these, they will be used later. I'm not 100% sure of the interface yet, so can we land them now and I'll add tests in the PR where they get used?

dbutenhof: Yeah, that's enough of an excuse ... I just don't really want to carry around code we know isn't used or tested.

PR #589, opened 2026-02-06
Drop various depricated settings and remove the default OpenAI request timeout

sjmonson:
## Summary

Drops the global  `GUIDELLM__REQUEST_TIMEOUT` as it was not wired up and adjusts the OpenAI request timeout to infinte for reads and 5 seconds for all other operations, with a tunable for connect timeouts.

## Details

The `GUIDELLM__REQUEST_TIMEOUT` setting used to configure HTTP timeouts for the OpenAI backend to a default of 5 minutes. However since v0.4.0 the backend has had its own default value of 1 minute. Rather then wire this back up I have opted to remove the global setting entirely since timeout can be configured with `--backend-kwargs '{"timeout": 6000}'`.

For the backend default httpx supports setting separate timeouts for connect, read, write, etc. `timeout` backend argument now only configures `read` timeouts and defaults to no timeout. The `connect` timeout can be configured with the backend argument `timeout_connect` and defaults to 5 seconds (httpx default). All other timeouts are set to 5 seconds (httpx default)

One major risk with this change is benchmarks that do not set any time based constraints can potentially deadlock if the the server stalls. This is very unlikely since it will require the server to hold connections open while stalled. It is also partially mitigated by setting a timeout for connect as in this case the only connection that can stall are in-progress ones where GuideLLM is waiting for a response. A functioning server should also close connections when they stall rather then holding them open.

## Test Plan

Run oversaturated server tests and confirm that timeout errors do not occur or only occur on server crash.

---

## Use of AI

### DIFF
--- a/src/guidellm/backends/openai/http.py
+++ b/src/guidellm/backends/openai/http.py
@@ -50,6 +50,9 @@ LEGACY_API_ALIASES = {
     "audio_translations": "/v1/audio/translations",
 }
 
+# NOTE: This value is taken from httpx's default
+FALLBACK_TIMEOUT = 5.0
+
 
 @Backend.register("openai_http")
 class OpenAIHTTPBackend(Backend):
@@ -84,7 +87,7 @@ class OpenAIHTTPBackend(Backend):
         api_routes: dict[str, str] | None = None,
         request_handlers: dict[str, Any] | None = None,
         timeout: float | None = None,
-        timeout_connect: float | None = 5.0,  # 5.0 is httpx default
+        timeout_connect: float | None = FALLBACK_TIMEOUT,
         http2: bool = True,
         follow_redirects: bool = True,
         verify: bool = False,
@@ -186,7 +189,7 @@ class OpenAIHTTPBackend(Backend):
         self._async_client = httpx.AsyncClient(
             http2=self.http2,
             timeout=httpx.Timeout(
-                5.0,  # default httpx timeout
+                FALLBACK_TIMEOUT,
                 read=self.timeout,
                 connect=self.timeout_connect,
             ),


src/guidellm/backends/openai/http.py
```diff
@@ -182,7 +185,11 @@ async def process_startup(self):
 
         self._async_client = httpx.AsyncClient(
             http2=self.http2,
-            timeout=self.timeout,
+            timeout=httpx.Timeout(
+                5.0,  # default httpx timeout
+                read=self.timeout,
+                connect=self.timeout_connect,
+            ),
```
sjmonson, line 195: Before we were setting all of these values to 60. I think that `read` is the only one that needs to be increased under normal usage since it is the timeout that is hit when waiting on first token. Needs more high-stress testing to confirm.

src/guidellm/backends/openai/http.py
```diff
@@ -83,7 +83,8 @@ def __init__(
         api_key: str | None = None,
         api_routes: dict[str, str] | None = None,
         request_handlers: dict[str, Any] | None = None,
-        timeout: float = 60.0,
+        timeout: float | None = None,
+        timeout_connect: float | None = 5.0,  # 5.0 is httpx default
```
dbutenhof: Not really an objection; but if we really need the line comment tying this to httpx, is there an httpx constant we can use rather than *assuming* the default **and** attributing it to httpx??

(OK -- so far as I can tell, httpx doesn't export or even use a constant for this. While none of this invalidates the line here, I question the value of the comment. *We're* choosing a default, I'm not sure it matters that, at the time the code was written, it was also the httpx default. 😁 )

src/guidellm/backends/openai/http.py
```diff
@@ -182,7 +185,11 @@ async def process_startup(self):
 
         self._async_client = httpx.AsyncClient(
             http2=self.http2,
-            timeout=self.timeout,
+            timeout=httpx.Timeout(
+                5.0,  # default httpx timeout
```
dbutenhof: Even though httpx doesn't seem to be using a defined constant, since we're justifying the literal 5.0 in two separate places as the "default httpx timeout", can *we* at least use a common constant so they don't end up accidentally diverging?

src/guidellm/utils/text.py
```diff
@@ -232,7 +231,7 @@ def load_text(data: str | Path, encoding: str | None = None) -> str:
 
     # check URLs
     if isinstance(data, str) and data.strip().startswith(("http", "ftp")):
-        with httpx.Client(timeout=settings.request_timeout) as client:
+        with httpx.Client() as client:
```
dbutenhof, line 234: So we're just trusting the httpx defaults here? (Which is probably fine as we're just trying to `GET` a text stream; but at one point someone thought an explicit timeout was justified.)

src/guidellm/backends/openai/http.py
jaredoconnell: Yeah, I agree that httpx's default is irrelevant. Just pick what makes sense for the backend.

sjmonson: Mostly the comment is to show where we get the `5.0` so if necessary someone can make an educated decision on changing it in the future.

src/guidellm/utils/text.py
sjmonson: This code is basically just a curl command. The reason we need to adjust the timeout on the backend is LLMs don't act like normal http traffic. They can hold open a connection for a large amount of time without responding. This code is just downloading a file, I doubt we'll hit the same issue here. Also the only code that uses this function is the HTTP UI to grab the most recent template.

src/guidellm/backends/openai/http.py
sjmonson: Done

PR #590, opened 2026-02-06
Multiturn Benchmarking

### LINKED ISSUE #138: support for multi-turn qa dataset
Hi Team,
I was wondering if we can add support for a multi-turn qa on lines of what lmcache team has - https://github.com/LMCache/LMCache/tree/dev/benchmarks/multi-round-qa - so that we don't have to run their benchmarking script for this scenario which is becoming quite an important one to evaluate agentic workflows.  
cc @sjmonson @dagrayvid @markurtz

### LINKED ISSUE #347: [Refactor] multi-turn support built on top of the new rewrite

sjmonson:
## Summary

Adds support for multiturn benchmarking. This is required for emulating advanced application flows such as agentic solutions.

## Details

Implements multiturn support in three areas:

1. Support in `data` for creating synthetic multiturn dataset as well as parsing user-provided datasets
2. Support in `scheduler` for handling requests in turn groups and persisting history across requests
3. Support in `backends/http` for formatting requests with history and responses

Some things missing from this PR that may be in a followup:

1. Support for varying turn count on synthetic data with `turns_stdev`, `turns_min`, and `turns_max`.
2. Configurable delay before sending a next turn (implemented just not configurable)
3. "users" variant of `concurrent` profile where conversations hold on to the concurrency semaphore (only truly useful when combined with delay)

## Test Plan

See `docs/guides/multiturn.md` for some good examples.

## Related Issues

- Resolves #347
- Resolves #138
- Replaces #211 

---

## Use of AI

### DIFF
--- a/src/guidellm/scheduler/worker.py
+++ b/src/guidellm/scheduler/worker.py
@@ -395,7 +395,7 @@ class WorkerProcess(Generic[RequestT, ResponseT]):
             await self._schedule_request(request, request_info, target_start)
 
             async for resp, info in self.backend.resolve(  # type: ignore[attr-defined]
-                request, request_info, history
+                request, request_info, history or None
             ):
                 request_info = info
                 if request_info is None:


src/guidellm/data/deserializers/synthetic.py
```diff
@@ -75,35 +75,31 @@ def __iter__(self) -> Iterator[tuple[int, dict[str, Any]]]:
...
-                        f"{self.iteration_count} {samples_count} ",
-                    ),
-                    "prompt_tokens_count": prompt_tokens_count,
-                    "output_tokens_count": output_tokens_count,
-                },
-            )
-            samples_count += 1
+            row: dict[str, Any] = {"prefix": next(prefix_iter)}
+            for turn in range(self.config.turns):
+                row[f"prompt_{turn}"] = self._create_prompt(
```
dbutenhof, line 80: Wow. Stringifying the turn, and then parsing it back out with regex injects a lot of complication. Maybe it's justified if that'll be transparent to external or even "many" internal consumers who don't care ... but I wonder if changing to a `(name, turn)` tuple key would ultimately be easier and cleaner. This would line up with the `DatasetColumnKey` type you use later on.

dbutenhof (commented):
A first pass to try to grok the landscape ... one comment; feel free to dismiss it if it's impractical for "reasons" not obvious to me from the context.

src/guidellm/data/deserializers/synthetic.py
sjmonson: The complication here is that we are massaging the synthetic data into a [`datasets.Dataset`](https://huggingface.co/docs/datasets/en/package_reference/main_classes#datasets.Dataset) which only supports string keys.

dbutenhof: Ah. (Though, judging from the `_BaseExamplesIterable`, it's technically `int | str`, internally typed as `Key` ... not that that helps since it doesn't look like we could get away with just using the turn, either. 😁)

Anyway, got it -- dependency on an external class. Bummer.

src/guidellm/data/preprocessors/mappers.py
```diff
@@ -13,6 +14,11 @@
 
 __all__ = ["GenerativeColumnMapper"]
 
+# (dataset_column_type, turn)  # noqa: ERA001
```
dbutenhof: Why not avoid the ugly `# noqa` code (which I had to look up!) by making these very sparse comments a bit more usefully English, like "DatasetColumnKey tuple groups a dataset column type and turn number"??

src/guidellm/data/builders.py
```diff
@@ -219,9 +219,7 @@ def process_dataset(
     Main method to process and save a dataset with sampled prompt/output token counts.
     """
     _validate_output_suffix(output_path)
-    logger.info(
-        f"Starting dataset conversion | Input: {data} | Output: {output_path}"
-    )
+    logger.info(f"Starting dataset conversion | Input: {data} | Output: {output_path}")
```
dbutenhof, line 222: I've been considering a PR to defer all debug formatting. It's hard to say exactly how much that'll really gain us; but the loggers are designed to support it, and it really makes sense at least for any log we don't expect to be formatted every time anyway. (E.g., we default to WARNING; so we could be "lazy" about WARNING, ERROR, CRITICAL ... but at least INFO and below should be deferred to save a little time.)

E.g., I'd prefer statements like this to be

```
    logger.info("Starting dataset conversion | Input: {} | Output: {}", data, output_path)
```

(Unlike the stdlib Logger, Loguru defaults to `.format` style `{}` rather than `%s` placeholders.)

src/guidellm/scheduler/worker.py
```diff
@@ -398,26 +434,52 @@ async def _process_next_request(self, target_start: float):
...
+        history: HistoryT[RequestT, ResponseT],
+        conversation: ConversationT[RequestT],
+        settings: RequestSettings,
+    ):
+        try:
+            if settings.next_turn_delay > 0:
+                await asyncio.sleep(settings.next_turn_delay)
+        except asyncio.CancelledError:
+            # If we are cancelled, dump straight to queue
+            raise
```
dbutenhof: What's the point of handling it and immediately re-raising? It might be interesting to have a `logger.debug` here, although even at `DEBUG` level that might be more noise than value; but catching and re-raising is basically an expensive comment -- is that really justified here?

src/guidellm/scheduler/worker_group.py
```diff
@@ -495,40 +490,55 @@ def requests_generator(
...
+            ) -> Generator[RequestDataT[RequestT], None, None]:
+                nonlocal count, stop_queueing
+                for request, delay in requests_chain:
+                    count += 1
+
+                    request_id: str
+                    if hasattr(request, "request_id"):
+                        request_id = str(request.request_id)
+                    elif hasattr(request, "id"):
+                        request_id = str(request.id)
```
dbutenhof, line 510: This is one of the things I hate about Python (and maybe particularly the IDE Python language plugin 🤔) -- do we really have various GuideLLM request types "masquerading" as the `RequestT` TypeVar prototype that loosely use all three variations of `id`, `request_id` and no ID field at all? That's kinda ugly... as well as almost impossible to track when reading the codebase. (Sigh. 😦)

(I know, you just shuffled this code around and it's not new, and I've seen it other places; it's just too weird not to comment on...)

src/guidellm/schemas/info.py
```diff
@@ -104,24 +104,6 @@ def last_reported(self) -> float | None:
         return max(valid_timings) if valid_timings else None
 
 
-class RequestSettings(StandardBaseDict):
```
dbutenhof: An example of where the commit-by-commit review breaks down. You added this in a previous commit, now it goes away. Not blaming you in any way for trying to break it down ... just a bit of evidence for my suspicion that it's almost impossible to really isolate independent "chunks" in most real case...

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -152,9 +155,11 @@ def __init__(self):
         self.streaming_usage: dict[str, int | dict[str, int]] | None = None
         self.streaming_response_id: str | None = None
 
-    def format(
+    def format(  # noqa: C901
```
dbutenhof, line 158: I won't actually complain here, since I think the complexity checker metrics are often just annoying ... but if we're not going to globally disable C901, shouldn't we be considering refactoring any method that triggers it rather than just locally disabling it?

docs/guides/multiturn.md
```diff
@@ -0,0 +1,405 @@
...
+
+For text completions, the conversation history is concatenated:
+
+```text
+prefix content prompt_0 content response to prompt_0 prompt_1 content response to prompt_1 prompt_2 content
+```
+
+## The TurnPivot Preprocessor
+
+GuideLLM supports passing multiple `--data` argument, each pointing to a separate dataset. Normally this is useful for layering columns from different datasets within the same request. For example adding a text column from one dataset to another with images or combining multiple normally-distributed synthetic datasets into a multimodal distribution. In multiturn we can use the **TurnPivot** preprocessor to transpose each matched dataset with our turn columns.
```
dbutenhof: `--data` would normally be considered an "option" (`--`) rather than an "argument" (positional), (even though "data" is a slightly ambiguous case as it's a "required option" 😆); and multiple instances would imply the plural "arguments" or "options" anyway.

And while I get what you mean by "transpose each matched dataset" I'm not sure the sentence really clarifies the transformation.

docs/guides/multiturn.md
```diff
@@ -0,0 +1,405 @@
...
+- `/v1/chat/completions` - Utilizing chat template formatting
+- `/v1/completions` - with basic concatenated history
+
+Audio endpoints (`/v1/audio/transcriptions`, `/v1/audio/translations`) do not support multiturn benchmarking.
+
+### Column Naming Requirements
+
+Turn-indexed columns must follow the naming convention:
+
+- All turn columns must use the same base name. E.g. `<prompt>_0`, `<prompt>_2`, etc.
```
dbutenhof: "E.g." implies a concrete example: I think you should use `prompt_0` and `prompt_2` here, in conjunction with the `{"text_column": "prompt"}` specification below. In fact, moving the "Column mapping applies ..." bullet *first* might help clarify that. This bullet should probably be last.

docs/guides/multiturn.md
```diff
@@ -0,0 +1,405 @@
...
+
+- Turn 1: 150 tokens in; 200 tokens out
+- Turn 2: (150 + 200) + 100 = 450 tokens in; 200 tokens out
+- Turn 3: (450 + 200) + 100 = 750 tokens in; 200 tokens out
+- Turn 4: (750 + 200) + 100 = 1050 tokens in; 200 tokens out
+- Turn 5: (1050 + 200) + 100 = 1350 tokens in; 200 tokens out
+
+### Additional Considerations
+
+Multi-turn benchmarking has a couple extra characteristics to consider when compared to single-turn:
```
dbutenhof: Maybe "additional characteristics", which I think sounds more clear than "a couple extra".

src/guidellm/data/builders.py
sjmonson: Totally agree but I think defer that to a dedicated PR. This log message is unrelated to the content of this PR.

src/guidellm/scheduler/worker.py
sjmonson: Weird pythonism, the finally statement underneath will run before the re-raise.

src/guidellm/scheduler/worker_group.py
sjmonson: Yeah I also hate this. Not even sure it is necessary but have avoided touching it just in-case it breaks something.

src/guidellm/data/builders.py
dbutenhof: Fair enough. Somehow I apparently failed to notice that you only changed the line breaks here. Hmm.

src/guidellm/scheduler/worker.py
dbutenhof: But the `finally` will run anyway, and you're not *doing* anything here, so I don't see how "finally before reraise" matters. If I'm missing some subtlety, I'd love to understand.

src/guidellm/scheduler/worker_group.py
dbutenhof: Ah ... and with the looseness of the `RequestT` prototype, verifying it never matters would be ridiculously tedious. So we can wonder how & why it got here in the first place, but it's probably impractical to worry about it. Sad.

docs/guides/multiturn.md
sjmonson: > `--data` would normally be considered an "option" (--) rather than an "argument" (positional), (even though "data" is a slightly ambiguous case as it's a "required option" 😆); and multiple instances would imply the plural "arguments" or "options" anyway.

Yeah "options" seems good.

> And while I get what you mean by "transpose each matched dataset" I'm not sure the sentence really clarifies the transformation.

Howbout, 

```suggestion
GuideLLM supports passing multiple `--data` options, each pointing to a separate dataset. If there are matches for the same column type across multiple datasets, they are treated as separate batches. Normally this is useful for layering columns from different datasets within the same request. For example adding a text column from one dataset to another with images or combining multiple normally-distributed synthetic datasets into a multimodal distribution. We can use the **TurnPivot** preprocessor to transpose turn columns and dataset batches.
```

I spent a lot of time struggling to convey this part in English which is why I gave the example below it. So I don't think it needs to be 100% perfect, but if you have a better idea feel free to suggest.

dbutenhof: Yeah -- agree this isn't really easy to describe in words, but I think the new alternative is a definite improvement. More importantly, no feasible improvements immediately occur to me, so let's go with that. 😁

src/guidellm/scheduler/worker.py
sjmonson: Yeah my bad. I was thinking that the finally statement would only run if no exception occurs or we explicitly catch the exception. I think I ported this code from the first multi-turn PR so very long ago and it used to work a little differently [(link)](https://github.com/vllm-project/guidellm/pull/211/changes#diff-a2646f8d413b5691fb552d608e2f652695326bc23ea881bc9ab2f3ce666dc857R240-R243). Will remove.

src/guidellm/data/builders.py
sjmonson: I think this file slipped though a merge without having the formatter run.

src/guidellm/backends/openai/request_handlers.py
sjmonson: I usually use code complexity as a warning more then a blocking error. In most cases it does serve as a reminder to check if logic can be cleaned up. IMO this case it would be less readable to break this into functions. Unfortunately ruff does not have the concept of a lint warning.

dbutenhof: Yeah, I get that -- the problem is that the `# noqa` permanently blocks reporting of that issue while effectively concealing whether it even still exists. (The comment tags often end up lingering even if the code is refactored.)

If C901 can't be reconfigured as a "warning" in ruff, maybe we should be using a better linter. 😆 I kinda like the idea of a complexity metric as a warning; I'm not thrilled about it as a blocking error, and I'm not thrilled about these local disables.

docs/guides/multiturn.md
```diff
@@ -0,0 +1,408 @@
...
+Multiturn benchmarking in GuideLLM uses indexed columns to represent conversational exchanges. Each turn in a conversation is represented by a set of columns with a numeric suffix indicating the turn index. Turn indexes can be any numerical value but we recommend ascending from 0 for simplicity.
+
+### Turn-indexed Column Format
+
+For a 3-turn conversation, the dataset could contain the following columns:
+
+- **prefix**: Optional system prompt that becomes a system message (applies to first turn)
+- **prompt_0**, **output_tokens_count_0**: Turn 1 prompt and requested output tokens
+- **prompt_1**, **output_tokens_count_1**: Turn 2 prompt and requested output tokens
+- **prompt_2**, **output_tokens_count_2**: Turn 2 prompt and requested output tokens
```
jaredoconnell:
```suggestion
- **prompt_2**, **output_tokens_count_2**: Turn 3 prompt and requested output tokens
```

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -419,6 +446,13 @@ def format(
         :param **kwargs: Additional keyword arguments for request formatting
         :return: The formatted request arguments
         """
+        prev_requests: list[GenerationRequestArguments] = []
+        if history:
+            # NOTE: Don't include history to avoid infinite recursion
```
jaredoconnell: As in don't include it at all, or don't include it under some conditions?

jaredoconnell (commented):
Looks good so far. I will continue reviewing on Monday.

src/guidellm/backends/openai/request_handlers.py
sjmonson: Should be "Does not" instead of "Don't". We do not include history in the recursive calls.

src/guidellm/scheduler/worker_group.py
```diff
@@ -46,6 +50,20 @@
...
+        ConversationT[RequestT],
+        tuple[
+            ResponseT | None,
+            RequestT,
+            RequestInfo,
+            SchedulerState,
+        ],
+    ],
+    type_params=(RequestT, ResponseT),
+)
```
jaredoconnell, line 65: This type is so nested with type that a brief description on what it is might be nice.

PR #595, opened 2026-02-11
Fix /v1/chat/completions formatting

### LINKED ISSUE #594: /v1/chat/completions using wrong format for batched data.
### Bug Description

When formatting for `/v1/chat/completions` the backend formats each batch item as a separate message.

### Expected Behavior

Instead of separate messages, batches should be added to the `content` array of a single message.

### Steps to Reproduce

1. Run guidellm with `--request-format /v1/chat/completions` and some form of multi-item batch (multi-modal, multiple datasets, etc)
2. Observe requests to server are formatted with each batch item separately.

### Operating System

N/A

### Python Version

N/A

### GuideLLM Version

v0.5.3

### Installation Method

pip install guidellm

### Installation Details

_No response_

```shell

```

### Additional Context

**Current format:**

```json
{
    "messages": [
        {
            "role": "system",
            "content": "<prefix_text>"
        },
        {
            "role": "user",
            "content": [
                {"type": "text", "text": "<prompt_text>"}
            ]
        },
        {
            "role": "user",
            "content": [
                {"type": "text", "text": "<prompt2_text>"}
            ]
        },
        {
            "role": "user",
            "content": [
                {"type": "image_url", "image_url": {"url": <image_data>}}
            ]
        }
    ],
    ...
}
```

**Expected format:**

```json
{
    "messages": [
        {
            "role": "system",
            "content": "<prefix_text>"
        },
        {
            "role": "user",
            "content": [
                {"type": "text", "text": "<prompt_text>"},
                {"type": "image_url", "image_url": {"url": <image_data>}},
                {"type": "text", "text": "<prompt2_text>"}
            ]
        }
    ],
    ...
}
```

sjmonson:
## Summary

Format each data column as content sections rather than completely separate messages.

## Details

The chat completions interface consists of a list of messages. Each message is intend to represent a turn in the conversation. Within each message there is a content section that can be a list if multiple types of content are being passed. See #594 for more complete detail of the problem.

## Test Plan

Run GuideLLM with `--request-format /v1/chat/completions` against a server endpoint then compare request_args format before and after patch/

## Related Issues

- Resolves #594 

---

## Use of AI

### DIFF
--- a/tests/unit/backends/openai/test_request_handlers.py
+++ b/tests/unit/backends/openai/test_request_handlers.py
@@ -649,13 +649,13 @@ class TestChatCompletionsRequestHandler:
 
         result = instance.format(data)
 
-        assert len(result.body["messages"]) == 2
+        assert len(result.body["messages"]) == 1
         assert result.body["messages"][0]["role"] == "user"
+        assert len(result.body["messages"][0]["content"]) == 2
         assert result.body["messages"][0]["content"][0]["type"] == "text"
         assert result.body["messages"][0]["content"][0]["text"] == "Hello"
-        assert result.body["messages"][1]["role"] == "user"
-        assert result.body["messages"][1]["content"][0]["type"] == "text"
-        assert result.body["messages"][1]["content"][0]["text"] == "How are you?"
+        assert result.body["messages"][0]["content"][1]["type"] == "text"
+        assert result.body["messages"][0]["content"][1]["text"] == "How are you?"
 
     @pytest.mark.sanity
     def test_format_messages_prefix(self, valid_instances):
@@ -670,9 +670,11 @@ class TestChatCompletionsRequestHandler:
 
         result = instance.format(data)
 
-        assert len(result.body["messages"]) == 1
+        assert len(result.body["messages"]) == 2
         assert result.body["messages"][0]["role"] == "system"
         assert result.body["messages"][0]["content"] == "You are a helpful assistant."
+        assert result.body["messages"][1]["role"] == "user"
+        assert result.body["messages"][1]["content"] == []
 
     @pytest.mark.sanity
     def test_format_messages_image(self, valid_instances):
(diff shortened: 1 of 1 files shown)


src/guidellm/backends/openai/request_handlers.py
```diff
@@ -13,6 +13,8 @@
 import base64
 from typing import Any, Protocol, cast
 
+from more_itertools import roundrobin
```
sjmonson, line 16: I've added `more-itertools` here as a dependency since I think there are more places in the codebase where it can be used. However it is not strictly necessary; the `roundrobin` function it implements is defined in the [python docs](https://docs.python.org/3/library/itertools.html#itertools-recipes) and could just be copied into a local utils file. Wdyt?

dbutenhof: Normally, why write it if it exists. This is a little less clear if we're requiring a new package just for this one special case (and I've never really looked at `more-itertools` before, but there are some freaky weird iterators here). I'd have said it's 50/50 if we'd had this discussion before you pulled it; but having already done the work, it'd be silly to undo it just to clone a function that already exists.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -363,7 +365,49 @@ class ChatCompletionsRequestHandler(TextCompletionsRequestHandler):
...
+        """
+        formatted_data = []
+        for item in column_data:
+            if column_type == "text_column":
+                formatted_data.append({"type": "text", "text": item})
+            elif column_type == "image_column":
+                formatted_data.append(
+                    {
+                        "type": "image_url",
+                        "image_url": {"url": item.get("image")},
```
dbutenhof, line 383: Do you really want to pass on `None` values in each of these cases if the expected field is missing from `item`, or should it complain?

It'd be easy to just put this whole `if`/`elif` chain with direct `item["image"]` field references into a `try/except ValueError` and report (roughly) `ValueError(f"{column_type} data is missing {exc!r} field")`

sjmonson: Might do a followup for this because all of the formatters (not just /v1/chat/completions) need this treatment and I am trying to keep this PR to a minimal scope.

PR #596, opened 2026-02-12
vLLM Python Backend

### LINKED ISSUE #178: Add support for a Python vLLM backend
**Description:**
Right now, GuideLLM only works w/ GuideLLM through the OpenAI compatible API backend to take the target URI and perform benchmarking on a vLLM server. We want to also add support for the python vLLM backend so that we can create scripts for benchmarks w/ vLLM and use it in-line with Python. 

**Acceptance Criteria:**
- Add support for the Python vLLM Backend
- Potentially helpful: https://github.com/triton-inference-server/vllm_backend - maybe not

### LINKED ISSUE #348: [Refactor] Python vLLM backend support built on top of the new rewrite

jaredoconnell:
## Summary

The VLLM Python backend is an alternative to the traditional client-server setup by utilizing the Python API rather than the HTTP API.

## Details

- Implemented using the shared backend protocol
- Uses vLLM's AsyncLLMEngine
- Started by using the arg `--backend vllm_python`
- vLLM isn't expected to be included in the CI due to the weight of it.

## Test Plan

- The unit tests ensure that the code uses vLLM's code correctly, mostly using mocks.
- Running the code in a local or lab environment is necessary to fully test it out. I would recommend starting with something simple, like the suggestion in the documentation.

## Related Issues

- Resolves #348 
- Resolves #178 

---

## Use of AI

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,7 +6,6 @@ env_list = py3{10,11,12,13}
 [testenv:tests]
 description = Run all tests
 dependency_groups = dev
-extras = vllm
 commands =
     python -m pytest {posargs:tests/}
 


src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,963 @@
...
+    from vllm.engine.arg_utils import AsyncEngineArgs
+    from vllm.outputs import RequestOutput
+
+    HAS_VLLM = True
+except ImportError:
+    AsyncLLMEngine = None  # type: ignore[assignment, misc]
+    AsyncEngineArgs = None  # type: ignore[assignment, misc]
+    SamplingParams = None  # type: ignore[assignment, misc]
+    RequestOutput = None  # type: ignore[assignment, misc]
+    HAS_VLLM = False
```
sjmonson: Move this under `src/guidellm/extras/vllm.py` and `from guidellm.extras.vllm import SamplingParams, AsyncLLMEngine, ..., HAS_VLLM` here.

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1033 @@
...
+        pipeline works. Mode is inferred from body and files (audio/chat/text).
+        """
+        arguments = getattr(request, "arguments", None)
+
+        if arguments is not None:
+            body = getattr(arguments, "body", None) or {}
+            stream_override = getattr(arguments, "stream", None)
+            stream = self.stream if stream_override is None else bool(stream_override)
+            files = getattr(arguments, "files", None) or {}
+        else:
```
sjmonson: A GenerationRequest will never have an `arguments` field as of #478; drop this.

src/guidellm/backends/__init__.py
```diff
@@ -19,6 +19,17 @@
...
+try:
+    from .vllm_python.vllm import VLLMPythonBackend
+    from .vllm_python.vllm_response import VLLMResponseHandler
+
+    HAS_VLLM_BACKEND = True
+except ImportError:
+    VLLMPythonBackend = None  # type: ignore[assignment, misc]
+    VLLMResponseHandler = None  # type: ignore[assignment, misc]
+    HAS_VLLM_BACKEND = False
+
```
sjmonson, line 30: Actually the above comment on vLLM extras is better here:

```suggestion
from guidellm.extras.vllm import HAS_VLLM

# Conditionally import VLLM backend if available
if HAS_VLLM:
    from .vllm_python import VLLMPythonBackend
```

src/guidellm/backends/__init__.py
```diff
@@ -19,6 +19,17 @@
     TextCompletionsRequestHandler,
 )
 
+# Conditionally import VLLM backend if available
+try:
+    from .vllm_python.vllm import VLLMPythonBackend
+    from .vllm_python.vllm_response import VLLMResponseHandler
```
sjmonson: Should import from next level. E.g. `from .vllm_python import VLLMPythonBackend, VLLMResponseHandler`.

src/guidellm/benchmark/schemas/generative/entrypoints.py
```diff
@@ -325,6 +379,52 @@ def single_to_list(
...
+            raise ValueError(
+                f"Backend '{backend_type}' requires a target parameter. "
+                "Please provide --target with a valid endpoint URL."
+            )
+
+        if not requires_target and self.target is not None:
+            raise ValueError(
+                f"Backend '{backend_type}' does not support a target parameter. "
+                "Please remove --target as this backend runs locally."
+            )
```
sjmonson: Don't do this, just ignore target if its not needed.

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1033 @@
...
+    @classmethod
+    def requires_model(cls) -> bool:
+        """
+        VLLM Python backend requires a model parameter.
+
+        The model must be specified to know which model to load for inference.
+
+        :return: True, as this backend requires a model parameter
+        """
+        return True
```
sjmonson: I don't really like this, its not scalable and requires us to implement all of the validation external to the backend. I think we need something equivalent to `info()` for `__init__` arguments. Maybe something like this to start and we can fully plan it out later:

```python
@classmethod
def schema(cls) -> dict[str, dict[str, Any]]:
    return {
        "target": {
            "required": False,
            "type": str,
        },
        "model": {
            "required": True,
            "type": str,
        },
        ...
    }
```

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1033 @@
...
+                if p:
+                    messages.append({"role": "system", "content": str(p)})
+            for t in text_parts:
+                if t:
+                    messages.append(
+                        {
+                            "role": "user",
+                            "content": [{"type": "text", "text": str(t)}],
+                        }
+                    )
```
sjmonson: Needs to match changes in #595.

src/guidellm/benchmark/entrypoints.py
```diff
@@ -103,8 +104,16 @@ async def resolve_backend(
...
         else None
     )
+
+    # Only pass target if it's provided (some backends don't need it)
+    create_kwargs = {**(backend_kwargs or {})}
+    if target is not None:
+        create_kwargs["target"] = target
+    if model is not None:
+        create_kwargs["model"] = model
+
```
sjmonson, line 113: Always pass target and model, just make the backend capable of ignoring arbitrary arguments.

src/guidellm/scheduler/worker.py
```diff
@@ -387,6 +388,11 @@ async def _process_next_request(self, target_start: float):
                 request_info.error = repr(exc)
                 request_info.traceback = traceback.format_exc()
                 request_info.timings.resolve_end = time.time()
+                # Log backend exception if enabled
+                if settings.logging.log_backend_exceptions:
+                    logger.exception(
+                        f"Backend exception for request {request_info.request_id}: {exc}"
+                    )
```
sjmonson: Wasn't this a part of a separate PR and we decided to not require a setting?

src/guidellm/__main__.py
```diff
@@ -460,6 +469,40 @@ def run(**kwargs):  # noqa: C901
...
+
+    # Validate model parameter
+    if requires_model and model is None:
+        raise click.BadParameter(
+            f"Backend '{backend_type}' requires a model parameter. "
+            "Please provide --model with a valid model identifier.",
+            ctx=click.get_current_context(),
+            param_hint="--model",
+        )
+
```
sjmonson: Duplicate of the `BenchmarkGenerativeTextArgs` validation. Should not validate here.

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1033 @@
...
+        return True
+
+    def __init__(
+        self,
+        model: str,
+        vllm_config: dict[str, Any] | None = None,
+        request_format: str | None = None,
+        target: str | None = None,  # Backend.create API; unused (local)
+        stream: bool = True,
+    ):
```
sjmonson, line 178: Ignore any unused args:

```suggestion
    def __init__(
        self,
        model: str,
        vllm_config: dict[str, Any] | None = None,
        request_format: str | None = None,
        stream: bool = True,
        **_,
    ):
```

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1049 @@
...
+        self,
+        mode: _Mode,
+        body: dict[str, Any],
+        stream: bool,
+        files: dict[str, Any],
+    ) -> None:
+        self.mode = mode
+        self.body = body
+        self.stream = stream
+        self.files = files
```
sjmonson: Couldn't this be a dataclass or pydantic model?

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1049 @@
...
+
+        if arguments is not None:
+            body = getattr(arguments, "body", None) or {}
+            stream_override = getattr(arguments, "stream", None)
+            stream = self.stream if stream_override is None else bool(stream_override)
+            files = getattr(arguments, "files", None) or {}
+        else:
+            columns = getattr(request, "columns", {}) or {}
+            prefix_parts = list(columns.get("prefix_column", []) or [])
+            text_parts = list(columns.get("text_column", []) or [])
```
sjmonson: Missing `audio_column`, `image_column`, and `video_column`. Multimodal could be moved to a followup PR but the rest of the code seems to assume it exists in the old `request.arguments` format.

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1049 @@
...
+                        ctx,
+                        response_handler,
+                        request_info,
+                        state["accumulated_text"],
+                        previous_token_count,
+                        iter_time,
+                    )
+                )
+                if token_delta > 0:
+                    state["total_output_tokens"] += token_delta
```
sjmonson: Have you observed `token_delta` being less than 0 or is this just catching the 0 case?

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1049 @@
...
+                "Install with: pip install 'guidellm[audio]'"
+            ) from e
+        try:
+            audio_samples = _decode_audio(audio_bytes)
+            if hasattr(audio_samples.data, "numpy"):
+                audio_array = audio_samples.data.numpy()
+            elif hasattr(audio_samples.data, "cpu"):
+                audio_array = audio_samples.data.cpu().numpy()
+            else:
+                audio_array = np.asarray(audio_samples.data)
```
sjmonson: Be careful with `hasattr`, it will hide invalid code paths. The only valid option is the first one. `audio_samples.data` is a Tensor and all tensors have a [numpy()](https://docs.pytorch.org/docs/stable/generated/torch.Tensor.numpy.html) function.

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1049 @@
...
+        :param max_tokens_override: Optional max_tokens from request (e.g. benchmark)
+        :return: Configured SamplingParams instance
+        """
+        max_tokens = (
+            body.get("max_tokens")
+            if body.get("max_tokens") is not None
+            else (max_tokens_override if max_tokens_override is not None else 16)
+        )
+        if max_tokens == 0:
+            max_tokens = 16
```
sjmonson: Why set `max_tokens` to 16 by default. Should be uncapped so we let the model decide when its done.

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1049 @@
...
+            "ignore_eos": ignore_eos,
+            "frequency_penalty": body.get("frequency_penalty", 0.0),
+            "presence_penalty": body.get("presence_penalty", 0.0),
+        }
+
+        params = {k: v for k, v in params.items() if v is not None}
+
+        return SamplingParams(**params)  # type: ignore[misc]
+
+    def _convert_vllm_output_to_openai_format(
```
sjmonson: Shouldn't the response handler be written to take `RequestOutput` natively? Seems like a lot of extra code/work to convert to this format first and then parse.

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1049 @@
...
+    ) -> str:
+        """Format messages as plain 'Role: content' lines."""
+        parts = []
+        for msg in formatted_messages:
+            role = msg.get("role", "user")
+            content = msg.get("content", "")
+            if role and content:
+                parts.append(f"{role.capitalize()}: {content}")
+        parts.append("Assistant: ")
+        return "\n".join(parts)
```
sjmonson: Plain should be equivalent to `/v1/completions`. Need to read vLLM code and double check but there should not be any "role" here. Prompts should be concated with `" ".join(...)`.

sjmonson (changes requested):
Core handling of vLLM seems good, the rest is mostly cleanup and minor fixes. There are a lot of dead codepaths that likely exist due to this PR being started before #478. A lot of converting to/from `/v1/chat/completions` format can be removed. Multimodel probably needs bit more work. Audio should be easy to fix but image/video don't seem implemented; this can be a later follow-up PR. Both plain and chat template formats need fixes, see respective comments; neither should be difficult.

docs/getting-started/benchmark.md
```diff
@@ -4,7 +4,7 @@ weight: -6
 
 # Run a Benchmark
 
-After [installing GuideLLM](install.md) and [starting a server](server.md), you're ready to run benchmarks to evaluate your LLM deployment's performance.
+After [installing GuideLLM](install.md) and [starting a server](server.md), you're ready to run benchmarks to evaluate your LLM deployment's performance. Alternatively, you can run benchmarks with the vLLM Python backend (`--backend vllm_python`) without a separate server; see [vLLM Python backend](../guides/vllm-python-backend.md).
```
dbutenhof: I'd break this up: "after installing and starting a server [...] or run the Python backend" obscures the latter case. E.g.,

1. [Install GuideLLM](install.md)
2. You can run GuideLLM two ways:
   a. targeting a [running OpenAI-compatible LLM server](server.md) or
   b. with an installed vLLM package using the vLLM Python backend (...)

docs/guides/vllm-python-backend.md
```diff
@@ -0,0 +1,68 @@
+# vLLM Python Backend
+
+The **vLLM Python backend** (`vllm_python`) runs inference in the **same process** as GuideLLM using vLLM's [AsyncLLMEngine](https://docs.vllm.ai/). No HTTP server is involved, reducing overheat and variables. This is useful for isolating performance bottlenecks or simplifying your benchmark setup. You do **not** pass `--target`; you **must** pass `--model`.
```
dbutenhof: The bolding on "same process" feels like overkill here.

Also, I think you want "overhead", not "overheat"; plus I think the "and variables" doesn't mean much in this context. Maybe "variability" would be better... or "variability due to network bandwidth and latency" or whatever else you have in mind here.

docs/guides/vllm-python-backend.md
```diff
@@ -0,0 +1,68 @@
...
+  
+  **Note:** Using [uv](https://github.com/astral-sh/uv) is not recommended for the vLLM Python backend because of potentially incompatible requirements between the two projects. Prefer pip or the container / existing vLLM environment.
+
+
+It is also possible to install GuideLLM and vLLM via pip using `pip install guidellm[vllm]`. This method may make **hardware acceleration** (e.g. CUDA) harder to get working. See [vLLM installation](https://docs.vllm.ai/en/latest/getting_started/installation) and GPU/hardware-specific docs there. For production or GPU use, the container or existing-install path is recommended.
+
+
+## Basic example
+
+Run a benchmark with the vLLM Python backend (no `--target`):
```
dbutenhof: This wording might suggest that it's the lack of `--target` that triggers the Python backend, when it's really the `--backend`. You've already said that we don't use `--target`, and you're not showing it below, so I'm not sure that the explicit reminder that "we don't use `--target` here" is really necessary.

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1049 @@
...
+        self.files = files
+
+
+def _check_vllm_available() -> None:
+    """Check if vllm is available and raise helpful error if not."""
+    if not HAS_VLLM:
+        raise ImportError(
+            "vllm is not installed. Please install it using "
+            "'pip install guidellm[vllm]' or 'pip install vllm>=0.6.0'"
+        )
```
dbutenhof, line 132: Except that since the dependency morass requires installing vLLM *first* if not *together*, suggesting at this point that they just `pip install vllm` seems potentially unhelpful ...

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1049 @@
...
+        user_config is left to vLLM's AsyncEngineArgs defaults. When
+        request_format is a custom template (not plain or default-template),
+        adds chat_template to config.
+
+        :param user_config: User-provided configuration dictionary
+        :return: Config dict for AsyncEngineArgs (model set; chat_template if set)
+        """
+        config = dict(user_config)
+
+        # Ensure model is set in config (required; overrides user if they passed it)
```
dbutenhof, line 238: This should be documented; there's a lot about discovering vLLM config options and how to pass them as backend args, so we should be careful to document options that *can't* be passed that way.

I get the necessity, although it seems fairly artificial (but probably too messy to fix now) ... `target` and `model` aren't actually required by `Backend.create`, even though `resolve_backend` always passes them ... arguably they should be demoted to `backend_kwargs` especially as we now have concrete examples which conflict in their requirements for those two parameters!

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1049 @@
...
+    async def process_shutdown(self):
+        """
+        Clean up VLLM AsyncLLMEngine instance and resources.
+
+        :raises RuntimeError: If backend was not properly initialized
+        """
+        if not self._in_process:
+            raise RuntimeError("Backend not started up for process.")
+
+        # Shutdown the async engine if it has a shutdown method
```
dbutenhof: Misleading comment ... there's no conditional here for "if" it has a shutdown method -- if there's an `_engine` we just call it. If it's really an "if" then the call should be conditional or in a captured `try` block... if not, can we remove that comment phrase?

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1049 @@
...
+        self, request_info: RequestInfo, iter_time: float
+    ) -> None:
+        """
+        Update request iteration timing information.
+
+        :param request_info: Request tracking info to update
+        :param iter_time: Current iteration time
+        """
+        if request_info.timings.first_request_iteration is None:
+            request_info.timings.first_request_iteration = iter_time
```
dbutenhof, line 715: Without checking other references ... below in token timing, the `is None` dynamic initialization case sets `token_iterations = 0` along with storing the first timestamp. Should this logic be following that same pattern?

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1049 @@
...
+        if not self._in_process:
+            raise RuntimeError("Backend not started up for process.")
+
+        # Shutdown the async engine if it has a shutdown method
+        if self._engine is not None:
+            self._engine.shutdown()
+            self._engine = None
+        self._in_process = False
+
+    async def validate(self):
```
dbutenhof, line 291: Apparently the HTTP backend used to do a real request to `validate` but was "downgraded" to do a `/health` check. So the question here is: do we want to depart from that convention by doing a real request here, or should we use the `check_health()` API to be equivalent to the HTTP behavior?

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1049 @@
...
+        if isinstance(content, str):
+            return content
+        if isinstance(content, list):
+            # Extract text from content blocks with type "text"
+            text_parts = []
+            for block in content:
+                if isinstance(block, dict):
+                    block_type = block.get("type", "")
+                    if block_type == "text":
+                        text = block.get("text", "")
```
dbutenhof: In both cases, it seems unnecessarily "wasteful" to add the `get` fallback to `""` since the default `None` is perfectly (and more straightforwardly) falsey already. The fallback suggests that the result will be used as a `str` in a context where `None` would present a problem ... that represents an unnecessary cognitive load (i.e., it makes me scan the code to try to find such reference ... which doesn't exist. 😁 )

This obviously isn't a big deal, and probably isn't significantly slower -- it just feels like it makes the code slightly less easy to read, and that always bugs me.

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1049 @@
...
+        delta_text = generated_text[len(accumulated_text) :]
+        delta_data = self._openai_payload_for_mode(
+            ctx.mode, delta_text, is_delta=True
+        )
+        if hasattr(request_output, "request_id") and request_output.request_id:
+            delta_data["id"] = request_output.request_id
+        json_str = json.dumps(delta_data)
+        json_str_decoded = (
+            json_str.decode("utf-8") if isinstance(json_str, bytes) else json_str
+        )
```
dbutenhof: Ouch ... this is entirely nonsensical unless you take the long journey to figure out that we try to import `orjson` as `json`. Not thrilled by that obfuscation, which makes this weird phrase necessary; but there you go. 😦

src/guidellm/backends/vllm_python/vllm.py
dbutenhof: RE: "less than 0" -- the test is `> 0`, and I could argue the test probably isn't even worthwhile for 0 if there's no possibility of it being less than 0. (Though, if it *can* be, `> 0` is more efficient than checking `>= 0` and adding the 0. 🙂)

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1049 @@
...
+    ) -> AsyncIterator[tuple[GenerationResponse, RequestInfo]]:
+        """Run engine.generate loop and yield final (response, request_info)."""
+        state["final_output"] = None
+        state["accumulated_text"] = ""
+        state["total_output_tokens"] = 0
+        previous_token_count = 0
+        stream = ctx.stream
+        engine = self._engine
+        if engine is None:
+            raise RuntimeError("Backend not started up.")
```
dbutenhof: There are several similar checks, some with slightly different exception text. Maybe one of the different patterns is meaningful, but I suspect it'd be better to give a single consistent message if the engine isn't started, probably by consistently calling `_validate_backend_initialized` rather than checking locally.

src/guidellm/backends/vllm_python/vllm_response.py
```diff
@@ -0,0 +1,213 @@
...
+            audio_seconds=output_details.get("seconds"),
+        )
+        return input_metrics, output_metrics
+
+    def _parse_line(self, line: str) -> dict[str, Any] | None:
+        """Parse SSE line to JSON dict; return None if done or invalid."""
+        if line == "data: [DONE]":
+            return None
+        line = (line or "").strip()
+        if not line or not line.startswith("data:"):
```
dbutenhof: If `line` is `""` (which is the least it can be after the previous line), then `not line.startswith(<anything>)`. Yeah, checking for an empty line is probably slightly more efficient, but I'd be more tempted to simplify the statement here by dropping the first phrase.

And if we're uninterested in a line that doesn't start with "data:", I'm a bit curious why we're special-casing that by returning `{}` instead of `None`?

Containerfile.vllm
```diff
@@ -0,0 +1,53 @@
...
+ARG BASE_IMAGE=vllm/vllm-openai
+
+FROM $BASE_IMAGE
+
+# release: take the last version and add a post if build iteration
+# candidate: increment to next minor, add 'rc' with build iteration
+# nightly: increment to next minor, add 'a' with build iteration
+# alpha: increment to next minor, add 'a' with build iteration
+# dev: increment to next minor, add 'dev' with build iteration
+ARG GUIDELLM_BUILD_TYPE=dev
```
dbutenhof, line 11: Do we really want `dev` default instead of `release`??

Containerfile.vllm
```diff
@@ -0,0 +1,53 @@
...
+# Switch to root for installing system deps and pip install
+USER root
+
+# Install git for setuptools-git-versioning (version discovery)
+RUN apt-get update && apt-get install -y --no-install-recommends git \
+    && rm -rf /var/lib/apt/lists/*
+
+ENV GUIDELLM_BUILD_TYPE=$GUIDELLM_BUILD_TYPE
+
+# Copy repository and install GuideLLM from source with pip (no uv, to avoid
```
dbutenhof: Hmm. "no uv" sounds stilted ... "not [with] uv" would sound a bit better.

Containerfile.vllm
```diff
@@ -0,0 +1,53 @@
...
+      org.opencontainers.image.documentation="https://blog.vllm.ai/guidellm/stable" \
+      org.opencontainers.image.license="Apache-2.0"
+
+ENV HOME="/home/guidellm" \
+    GUIDELLM_OUTPUT_DIR="/results"
+
+WORKDIR $HOME
+
+# Ensure that the user home dir can be used by any user
+# (OpenShift Pods can't use the cache otherwise)
```
dbutenhof, line 44: Similar comment that I made on Kevin's PR ... granted, he responded that this is essentially a quote from the OpenShift documentation, it still bugs me. It can be used by any user *in gid 0". That includes both the random uid in OpenShift and the root (0) uid in standalone podman, both in gid 0; but it's still not "**any** user".

Containerfile.vllm
```diff
@@ -0,0 +1,53 @@
...
+RUN chgrp -R 0 "$HOME" && chmod -R g=u "$HOME"
+
+VOLUME /results
+
+# Root group for k8s
+USER 1001:0
```
dbutenhof, line 50: A constant uid? Is that ... wise?

dbutenhof (commented):
More comments / questions / rambling

docs/guides/vllm-python-backend.md
jaredoconnell: Good points. I re-worded this paragraph.

docs/guides/vllm-python-backend.md
jaredoconnell: I agree. I've removed the extra note.

docs/getting-started/benchmark.md
jaredoconnell: I added it with this list format. Let me know what you think.
I also added some extra notes.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: Sure. Done.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: It's possible to install it that way if your machine is one natively supported. But I guess this isn't the place for that, so I just made the message more generic.

docs/getting-started/benchmark.md
```diff
@@ -4,7 +4,20 @@ weight: -6
 
 # Run a Benchmark
 
-After [installing GuideLLM](install.md) and [starting a server](server.md), you're ready to run benchmarks to evaluate your LLM deployment's performance. Alternatively, you can run benchmarks with the vLLM Python backend (`--backend vllm_python`) without a separate server; see [vLLM Python backend](../guides/vllm-python-backend.md).
+1. [Install GuideLLM](install.md)
+2. You can run GuideLLM two ways:
+   1. Targeting a running OpenAI-compatible LLM server
+      1. The typical workflow
```
dbutenhof: I think maybe the second level items should just be bullets, not numbered items, and I'm not sure that "The typical workflow" is particularly useful. Maybe drop the second level bullets and just go with something like:

1. Targeting a running OpenAI-compatible LLM server. This is the default operation.

Operation? Workflow? We're specifying (or defaulting) a "backend", but that may be a bit technically specific for this level of documentation. Mode? ❓

docs/getting-started/benchmark.md
```diff
@@ -4,7 +4,20 @@ weight: -6
...
+   2. Using the vLLM Python backend, with vLLM running in the same process
+      1. Advanced, but all-in-one.
```
dbutenhof: "Advanced" doesn't really tell us a lot. Can you expand on that briefly? And "all-in-one" is a looser restatement of "with vLLM running in the same process", and doesn't really add anything.

docs/getting-started/benchmark.md
```diff
@@ -4,7 +4,20 @@ weight: -6
...
+      2. For more information, see [vLLM Python backend](../guides/vllm-python-backend.md)
+
+> [!NOTE]\
+> Everything in this guide applies to both backends except the backend-specific inputs.
+>
+> This guide assumes you're using the OpenAI HTTP backend with an OpenAI-compatible LLM server.
```
dbutenhof: The "For more information on the vLLM Python backend, see ..." sentence might be more appropriate as a second sentence in this paragraph so it's part of the main flow rather than isolated as a second level bullet item.

docs/getting-started/benchmark.md
```diff
@@ -40,16 +53,16 @@ GuideLLM offers a wide range of configuration options to customize your benchmar
...
-| `--data`        | Data configuration for benchmarking            | `--data "prompt_tokens=256,output_tokens=128"` |
-| `--profile`     | Type of benchmark profile to run               | `--profile sweep`                              |
-| `--rate`        | Request rate or number of benchmarks for sweep | `--rate 10`                                    |
-| `--max-seconds` | Duration for each benchmark in seconds         | `--max-seconds 30`                             |
-| `--output-dir`  | Directory path to save output files            | `--output-dir results/`                        |
-| `--outputs`     | Output formats to generate                     | `--outputs json csv html`                      |
+| Parameter       | Description                                          | Example                                        |
+| --------------- | ---------------------------------------------------- | ---------------------------------------------- |
+| `--target`      | URL of the OpenAI-compatible server                  | `--target "http://localhost:8000"`             |
```
dbutenhof: Should we add "ignored for vLLM Python backend"? (That's a bit more than I'd really want to squeeze into the cell, unfortunately, but it corresponds with your "with http backend" below.)

docs/getting-started/benchmark.md
```diff
@@ -40,16 +53,16 @@ GuideLLM offers a wide range of configuration options to customize your benchmar
...
+| `--model`       | Model name to benchmark (optional with http backend) | `--model "Meta-Llama-3.1-8B-Instruct"`         |
+| `--data`        | Data configuration for benchmarking                  | `--data "prompt_tokens=256,output_tokens=128"` |
```
dbutenhof: Markdown doesn't care about the alignment of the pipe characters, and I'm not that concerned about "pretty" raw Markdown code ... do we really need to change the alignment on each row to keep them lined up?? The lines get long and, without Markdown's wrapping, I think the alignment makes the raw text *harder* to read.

docs/guides/vllm-python-backend.md
```diff
@@ -1,6 +1,6 @@
 # vLLM Python Backend
 
-The **vLLM Python backend** (`vllm_python`) runs inference in the **same process** as GuideLLM using vLLM's [AsyncLLMEngine](https://docs.vllm.ai/). No HTTP server is involved, reducing overheat and variables. This is useful for isolating performance bottlenecks or simplifying your benchmark setup. You do **not** pass `--target`; you **must** pass `--model`.
+The **vLLM Python backend** (`vllm_python`) runs inference in the same process as GuideLLM using vLLM's [AsyncLLMEngine](https://docs.vllm.ai/). No HTTP server is involved, eliminating network and IPC overhead. This is useful for isolating performance bottlenecks or simplifying your benchmark setup. You do **not** pass `--target`; you **must** pass `--model`, which will then be downloaded and hosted in vLLM.
```
dbutenhof: IPC overhead?? Where does the http backend introduce that??? 😕

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: Especially with the new input validation, I think I can just remove that line altogether.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: I've documented the limitation and added a warning when the value is specified in the backend options.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: I removed the comment. I think that's leftover from when I was using a different type in vLLM for the prototype. The prior one wasn't shutting down correctly.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: Good call. I just added the call to `check_health` in addition to the inference.

I think it probably makes sense to keep the inference check, since it's the most thorough way to ensure that inference works. I checked and the quick token generation takes 70-90 milliseconds. 
I am open to removing it if you two think that's worth it.

docs/guides/vllm-python-backend.md
jaredoconnell: I made it just "eliminating network overhead"

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: Good point. I guess I can remove the redundant assignment of 0.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: Good point. Done.

src/guidellm/backends/__init__.py
jaredoconnell: Imported from the next level.

src/guidellm/backends/__init__.py
jaredoconnell: Switched to using HAS_VLLM.

docs/getting-started/benchmark.md
dbutenhof: And ... working on my documentation change, I realize that the precommit hooks are doing this automatically. I still think it's kinda ridiculous, but not worth worrying about...

src/guidellm/benchmark/entrypoints.py
jaredoconnell: I ended up making is even more generic.

Containerfile.vllm
jaredoconnell: I wonder that as well, but it is following the example set by the non-vllm Containerfile. My guess is that most builds are dev, but the others are generated by CI, so the CI is responsible for passing in the appropriate build type value. Which makes sense.

Containerfile.vllm
jaredoconnell: I switched it to "without uv".

Containerfile.vllm
jaredoconnell: Leaving this out makes it basically impossible to use on OpenShift. It's picky. Is there a change you would want?

Containerfile.vllm
jaredoconnell: I don't think it ends up mattering. The main value of this line is that it's a non-privileged user.

docs/getting-started/benchmark.md
jaredoconnell: I'm not sure. The documentation has the HTTP backend as the tier-1 option. The vLLM python one is not tier-1 since it's more specialized. The backend will properly report to the user when they try to use it, so I'm not too concerned. But I'm also not opposed to adding that.

docs/getting-started/benchmark.md
jaredoconnell: As you mentioned, this is an automated linter. I personally like the formatting for readability.

docs/getting-started/benchmark.md
jaredoconnell: I moved it to the NOTE block. How's that?

docs/getting-started/benchmark.md
jaredoconnell: I re-worded this.

docs/getting-started/benchmark.md
jaredoconnell: I switched it to "The most common setup"

src/guidellm/scheduler/worker.py
jaredoconnell: I fixed this when I re-rebased it. Probably a merge/rebase error

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: Switched to simpler logic.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: This part was AI generated. It seems to just use hasattr to hallucinate a way of doing things.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: I removed it. That was from debugging when there was a problem that led to problems with token counts.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: I reformatted it. Hopefully I got them all.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: Fixed. Though it looks like an empty string concatenation is the right way, so that's how I implemented it.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: I still need to address this. I will do that tonight.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: Fixed.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: Done.

src/guidellm/benchmark/schemas/generative/entrypoints.py
jaredoconnell: I refactored it. The new design kinda still does this in a generic way, so hopefully that's good. I think the new design is good.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: Refactored to a generic pydantic based validation.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: I finished support for multimodal. I just don't have any references to ensure that the implementation is performing as expected.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: I will look into this tonight.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: I will look into this tonight.

jaredoconnell (commented):
I addressed almost all requested changes. It's much simpler now. Much less dead code.

src/guidellm/backends/vllm_python/vllm.py
dbutenhof: I think that if we're going to argue here that the inference check provides better "safety" than just the health check, we should consider doing that on the `http` backend as well to make it "as safe". I'm not strongly arguing one way or another, but I'd prefer that we're consistent. If it's really not working, we'll find out anyway. (Granted, we ought to be failing early on a 404 or equivalent rather than running out the full test and reporting empty data; but we need to fix that even with a "robust" initial check.)

Containerfile.vllm
dbutenhof: Yeah, I'm not complaining about the `chgrp` & `chmod`, just the incorrect *comment*. (It doesn't ensure it can be used by "any user", only by "users in gid 0; and I'm kinda surprised that more people, and OpenShift documentation, don't consider that an important distinction. I mean, it doesn't matter *here* because we know that in k8s/OpenShift/local-podman we *will* be in gid 0 regardless of the uid, but I object to making the over-generalized statement here. Oh well... clearly I'm in the minority.)

Containerfile.vllm
dbutenhof: Maybe. For podman, you're virtual root and can do anything, so setting a random non-existent uid isn't a big deal. I think it's weird this doesn't trip anything up in OpenShift where the container is launched with a real non-0 uid that's probably *not* 1001. But ... whatever.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: Agreed that we need a better way to handle errors during inference. We could remove this once that update is complete.
What are your thoughts, @sjmonson?

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: Yeah that doesn't add any value. Removed the explicit default.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: This section may have been removed in the refactor.

docs/guides/vllm-python-backend.md
```diff
@@ -0,0 +1,70 @@
...
+  Build and run the image that uses the vLLM base image (e.g. [Containerfile.vllm](https://github.com/vllm-project/guidellm/blob/main/Containerfile.vllm)). It is based on `vllm/vllm-openai` and installs GuideLLM on top, giving a known-good vLLM + GuideLLM stack with hardware support as provided by the base image.
+
+  **Note:** This method will result in the preference for vllm's requirements as opposed to GuideLLM's requirements. Since vLLM is the more complex project, this is the recommended configuration, but this may result in an older Python or dependency version, resulting in sub-optimal GuideLLM performance and behavior in some scenarios.
+
+- **Existing vLLM installation**\
+  Install vLLM first for your environment (GPU/CPU, CUDA, etc.), then install GuideLLM in the same environment (e.g. `pip install guidellm` or with extras). You avoid a duplicate vLLM install and reuse your existing acceleration setup.
+
+  **Note:** Using [uv](https://github.com/astral-sh/uv) is not recommended for the vLLM Python backend because of potentially incompatible requirements between the two projects. Prefer pip or the container / existing vLLM environment.
```
sjmonson: I think drop this note, its not entirely true. The way most people use `uv` to consume other peoples packages is `uv pip install` which will work just fine.

src/guidellm/backends/openai/http.py
```diff
@@ -28,9 +29,34 @@
 
 __all__ = [
     "OpenAIHTTPBackend",
+    "OpenAIHttpBackendArgs",
 ]
 
 
+class OpenAIHttpBackendArgs(BaseModel):
```
sjmonson: Yeah I like this design a lot more. For possible future extensibility add a empty `BaseModel` called `BackendArgs` to `backend.py` and have these classes inherent from it. Maybe also add `request_format` here too.

src/guidellm/backends/vllm_python/vllm.py
sjmonson: When we used real requests to validate the http backend people complained that `--max-requests` did not strictly enforce the total number of requests sent to the server. I think due to server-side metrics. So no, lets not revert to the old HTTP backend behavior. Here I am less concerned since the server-side is opaque. If someone complains we should find something akin to `/health` but for now leave it as is.

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1029 @@
...
+    ) -> str:
+        """Concatenate message content into a single raw prompt string.
+
+        Equivalent to the HTTP /v1/completions behaviour: prefix + text
+        with no role prefixes or trailing generation prompt.
+        """
+        return "".join(
+            msg.get("content", "") for msg in formatted_messages
+            if msg.get("content")
+        )
```
sjmonson: Add a space between joined messages and index `msg["content"]` directly since it is already checked by the if statement.

```suggestion
        return " ".join(
            msg["content"] for msg in formatted_messages
            if msg.get("content")
        )
```

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1029 @@
...
+        :param request: Column-based generation request
+        :return: Resolved request with formatted prompt and multimodal data
+        :raises ValueError: If request has no text or multimodal columns
+        """
+        columns = request.columns
+
+        messages: list[dict[str, Any]] = []
+
+        prefix = " ".join(
+            str(p) for p in (columns.get("prefix_column") or []) if p
```
sjmonson: Small optimization. Didn't verify so correct me if I am wrong and this does not work.

```suggestion
            str(p) for p in columns.get("prefix_column", []) if p
```

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1029 @@
...
+        )
+        if prefix:
+            messages.append({"role": "system", "content": prefix})
+
+        text_blocks = self._format_column_blocks(
+            columns.get("text_column") or [], "text_column"
```
sjmonson: Small optimization. Didn't verify so correct me if I am wrong and this does not work.

```suggestion
            columns.get("text_column", []), "text_column"
```

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -0,0 +1,1029 @@
...
+            multi_modal_data
+            and (text_blocks or prefix)
+            and self.request_format != "plain"
+        )
+
+        if use_content_blocks:
+            # Interleave text and media blocks into a single content list,
+            # matching the HTTP backend's roundrobin approach.
+            media_lists = [
+                self._format_column_blocks(columns.get(col) or [], col)
```
sjmonson: Small optimization. Didn't verify so correct me if I am wrong and this does not work.

```suggestion
                self._format_column_blocks(columns.get(col, []), col)
```

src/guidellm/benchmark/schemas/generative/entrypoints.py
```diff
@@ -39,7 +41,75 @@
...
+
+    :param backend: Backend type identifier or Backend instance
+    :return: The backend's backend_args model class
+    :raises ValueError: If backend type is not registered
+    """
+    if isinstance(backend, Backend):
+        return backend.__class__.backend_args()
+    backend_class = Backend.get_registered_object(backend)
+    if backend_class is None:
+        registry = getattr(Backend, "registry", None) or {}
```
sjmonson: Why would `Backend.registry` ever be none? Since this on the base class it should always exist.

src/guidellm/benchmark/schemas/generative/entrypoints.py
```diff
@@ -39,7 +41,75 @@
...
 
-__all__ = ["BenchmarkGenerativeTextArgs"]
+__all__ = [
+    "BenchmarkGenerativeTextArgs",
+    "format_backend_args_error",
+    "get_backend_args",
+]
+
+
+def get_backend_args(backend: BackendType | Backend) -> type[BaseModel]:
```
sjmonson: Make this a classmethod on `Backend` similar to `create`.

Containerfile.vllm
sjmonson: Podman will automatically boothstrap the container user/group to `/etc/passwd` / `/etc/group`. See `podman run --help | grep passwd`

OpenShift does the same but also sets uid to a namespace specific value. Running a container in OpenShift is equivalent to running rootful podman with `--user $NS_UID:0`.

pyproject.toml
```diff
@@ -11,6 +11,13 @@ include = ["*"]
 "guidellm.data" = ["*.gz"]
 "guidellm.benchmark.scenarios" = ["*.json", "**/*.json"]
 
+[tool.uv]
+# vllm hard-pins specific torch versions in its metadata, but is compatible
+# with newer versions at runtime. Override to allow a single torch resolution.
+override-dependencies = [
+    "torch>=2.10.0",
+]
```
sjmonson, line 20: Did you try leaving the version specifier off? E.g. 


```suggestion
override-dependencies = [
    "torch",
]
```

Based on some comments in https://github.com/astral-sh/uv/issues/9258 that might work. It would be nice to not have to update the version specifier in multiple places.

tox.ini
```diff
@@ -6,6 +6,7 @@ env_list = py3{10,11,12,13}
 [testenv:tests]
 description = Run all tests
 dependency_groups = dev
+deps = vllm
```
sjmonson:
```suggestion
extras = vllm
```

src/guidellm/backends/openai/http.py
sjmonson: Actually better idea: make `BackendArgs` a `RegistryMixin`. Then we can follow the same creation pattern as `Backend`.

src/guidellm/__main__.py
```diff
@@ -460,6 +469,21 @@ def run(**kwargs):  # noqa: C901
...
+        inputs = {k: kwargs.get(k) for k in args_model.model_fields}
+        args_model.model_validate(inputs)
+    except ValidationError as err:
+        param_hint, message = format_backend_args_error(args_model, backend_type, err)
+        raise click.BadParameter(
+            message,
+            ctx=click.get_current_context(),
+            param_hint=param_hint,
+        ) from err
+
```
sjmonson, line 485: Still don't like this here, though I have an idea to get rid of it. Low priority so I'll bring it up after this merges (Just resolve this comment).

src/guidellm/backends/vllm_python/vllm.py
sjmonson: Marking as resolved but we should definitely come up with a set of manual multimodal tests.

src/guidellm/backends/vllm_python/vllm.py
dbutenhof: > Here I am less concerned since the server-side is opaque. If someone complains we should find something akin to /health but for now leave it as is.

Meaning that nobody can scrape the equivalent of `/metrics` beyond what we already report, since vLLM doesn't persist. Fair enough. However, if we're "happy" with `/health` for HTTP, I don't see why we wouldn't be "happy enough" just calling `check_health` here, without the inference.

Not a big deal, I guess ... at this point.

docs/getting-started/benchmark.md
dbutenhof: I suppose maybe the "more work but potentially less confusing" alternative would be to break up sections like this by backend, so we can unambiguously document requirements for HTTP vs Python vs (soon) gRPC.

src/guidellm/backends/vllm_python/vllm.py
sjmonson: My selective vision somehow missed the note on `check_health`. Yeah that would be sufficient in my mind.

docs/guides/vllm-python-backend.md
jaredoconnell: I updated it to more clearly specify the issue.
>   **Note:** Installing from the lockfile for the vLLM Python backend may not install the correct dependencies for hardware acceleration.

src/guidellm/backends/openai/http.py
jaredoconnell: I don't see how a RegistryMixin makes sense for this situation. There is currently a 1:1 relationship between permitted arguments and their associated backends. RegistryMixin is more for multiple options within a backend.

I implemented the switch to a new base class, as well as adding `request_format`. The end result is the error message is better and you don't get a stacktrace for an invalid value for the http backend anymore.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: Implemented.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: I researched the difference, and the current design is more resilient in case the column has a falsey value. So if it's blank but still present in the column, the current code would address it, but the proposed code wouldn't.
So I originally implemented the suggestions, but based on these findings, I'll keep the code as-is in case the dataset isn't set up properly.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: See my other comment regarding falsey values.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: See my other comment regarding falsey values.

src/guidellm/benchmark/schemas/generative/entrypoints.py
jaredoconnell: I simplified this to remove getattr, but the None check is necessary because the registry is initialized to None:
```
registry: ClassVar[dict[str, RegistryObjT] | None] = None
```

pyproject.toml
jaredoconnell: The current requirement in vllm is `torch==2.10.0`
This override allows more lax version requirements, whereas without it, it must install 2.10.0. What do you think we should do, with this context?

src/guidellm/benchmark/schemas/generative/entrypoints.py
jaredoconnell: Done.

tox.ini
jaredoconnell: Done.

docs/guides/vllm-python-backend.md
```diff
@@ -16,7 +16,7 @@ For all engine options and supported models, see vLLM's [Engine Arguments](https
 - **Existing vLLM installation**\
   Install vLLM first for your environment (GPU/CPU, CUDA, etc.), then install GuideLLM in the same environment (e.g. `pip install guidellm` or with extras). You avoid a duplicate vLLM install and reuse your existing acceleration setup.
 
-  **Note:** Using [uv](https://github.com/astral-sh/uv) is not recommended for the vLLM Python backend because of potentially incompatible requirements between the two projects. Prefer pip or the container / existing vLLM environment.
+  **Note:** Installing from the lockfile for the vLLM Python backend may not install the correct dependencies for hardware acceleration.
```
dbutenhof, line 19: That sounds a bit "off". Maybe "GuideLLM uv lockfile"; and isn't the concern that a `uv sync` from looser (locked) GuideLLM requirements might downgrade vLLM dependencies? If that's the case, I think we should be more specific than "may not install"...

jaredoconnell: We specify the CPU versions of pytorch. I found trying to install this via the lockfile to be a nightmare to get CUDA working.
But if you two don't think this sounds right, let me know and I'll remove it.

src/guidellm/backends/vllm_python/vllm.py
sjmonson: The type for `GenerationRequest.columns` is `dict[str, list[Any]]` A column is **always** a list. If the dataset is not set up property it will fail much earlier.

pyproject.toml
sjmonson: I tried what I said and it worked. Setting `override-dependencies = ["torch"]` means that uv treats vLLM as having a requirement of `"torch"` rather than `"torch==2.10.0"` so it can use our normal locked version.

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -354,8 +354,8 @@ def _build_multi_modal_data_from_columns(  # noqa: C901, PLR0912
         multi_modal_data: dict[str, Any] = {}
         # We look specifically for "image_column" and "audio_column" which contain lists
         # of dicts
-        image_items = list(columns.get("image_column", []) or [])
-        audio_items = list(columns.get("audio_column", []) or [])
+        image_items = columns.get("image_column", [])
+        audio_items = columns.get("audio_column", [])
```
dbutenhof, line 358: I thought Jared's comment made sense here.

```python
columns = {"image_column": None}
columns.get("image_column", []) -> None (not [])
columns.get("image_column") or [] -> []
```

Is `None` what we want here, considering we're processing "arbitrary" input files?

The `for item in image_items:` will blow up if it's ever possible that the input `column` data has a `None` ...

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -625,12 +625,12 @@ def _resolve_request(self, request: GenerationRequest) -> _ResolvedRequest:
 
         messages: list[dict[str, Any]] = []
 
-        prefix = " ".join(str(p) for p in (columns.get("prefix_column") or []) if p)
+        prefix = " ".join(str(p) for p in columns.get("prefix_column", []) if p)
```
dbutenhof, line 628: Same here -- if `{"prefix_column": None}` then the `for` will blow up. If you're sure the input data can never have `None`, we're good ...

dbutenhof (approved):
You overrode Jared's protections on the column field accesses. I'm making comments, but also approving on the assumption that you did this out of certain knowledge that I (and apparently Jared) lack about about the input predicates -- particularly, that the column values really can never be `None`. So ... you're sure? 😁

src/guidellm/backends/vllm_python/vllm.py
sjmonson: Read my response to the original comment `"prefix_column": None` is impossible. It will fail in the data generator since the type is enforced to `list[Any]`.

src/guidellm/backends/vllm_python/vllm.py
sjmonson: Read my response to the original comment `{"image_column": None}` is impossible. It will fail in the data generator since the type is enforced to `dict[str, list[Any]]`.

sjmonson: [link](https://github.com/vllm-project/guidellm/pull/596#discussion_r2932864759) since it is getting very difficult to find specific commits on this PR.

dbutenhof: I saw that comment, and in the heat of the moment the implications got lost. Got you: in that case, your simplification is better.

src/guidellm/backends/openai/http.py
sjmonson: The idea was you would first call `args = BackendArgs.create("vllm_python", **kwargs)` to get a validated `VLLMPythonBackendArgs`. Then you could call `Backend.create("vllm_python", args)` to create the backend. That way we can avoid some of the extra validation code. Needs some more planning though so I think what you did is good for now and we can look at it later.

tox.ini
```diff
@@ -6,7 +6,6 @@ env_list = py3{10,11,12,13}
 [testenv:tests]
 description = Run all tests
 dependency_groups = dev
-extras = vllm
```
sjmonson, line 9: vLLM is not critical for any tests. Since its causing CI issues will just remove for now.

docs/guides/vllm-python-backend.md
sjmonson: I think this is done... Will resolve and can open an issue later if wrong.

src/guidellm/backends/vllm_python/vllm.py
sjmonson: I think this is done...

PR #605, opened 2026-02-23
Feat/cross rate early exit

### LINKED ISSUE #604: Stop multi-rate benchmarks after first failure threshold
### Problem Statement

When running multi-rate benchmark profiles (`constant`, `poisson`, `concurrent`, and the async phase of `sweep`), benchmarking continues at higher rates even after a lower rate has already hit a failure constraint (for example over-saturation or excessive errors).

This wastes benchmark time, adds low-signal data at clearly unsustainable rates, and can place unnecessary load on the target system.

### Proposed Solution

- Treat scheduler constraints with `request_processing=stop_all` as terminal for rate escalation.
- Sort configured rates/streams ascending for deterministic escalation.
- For multi-rate profiles, stop generating higher-rate strategies after the first terminal failure.
- Keep normal completion constraints (`stop_local`, e.g. max duration/max requests) non-terminal so progression continues.
- In `sweep`, always run `synchronous` and `throughput`, then apply early-exit logic only during the async-rate phase.

### Alternatives Considered

_No response_

### Usage Examples

```markdown

```

### Additional Context

_No response_

ushaket:
## Summary

This PR adds cross-rate early-exit behavior for multi-rate benchmark profiles so benchmarks stop escalating once a terminal failure condition is hit at a lower rate/stream. It also makes rate/stream ordering deterministic (ascending) and updates user-facing docs/help text to reflect multi-rate semantics and skip behavior.

## Details

- Added shared failure-check logic in profile flow to detect terminal scheduler constraints (`request_processing=stop_all`) from the previous benchmark state.
- Updated `AsyncProfile.next_strategy()` to stop scheduling higher rates after a terminal failure; continue normally on `stop_local`.
- Updated `ConcurrentProfile.next_strategy()` with the same early-exit behavior for stream escalation.
- Updated `SweepProfile.next_strategy()` so `synchronous` and `throughput` always run, with early-exit applied only during async-rate continuation.
- Sorted multi-value rates/streams ascending in argument resolution for deterministic progression; added warning logs when input order is changed.
- Updated CLI help and README to document per-profile `--rate` semantics, multi-value behavior, and failure-triggered skipping.
- Added unit tests covering:
  - sorting behavior
  - continuation on normal completion (`stop_local`)
  - early exit on terminal failures (`stop_all`)
  - sweep-specific phase behavior and edge cases

## Test Plan

- Run unit tests for benchmark profiles:
  - `pytest tests/unit/benchmark/test_profiles.py`
- Sanity-check existing benchmark profile tests still pass:
  - `pytest tests/unit/benchmark -k profile`
- Run a multi-rate benchmark and verify higher rates are skipped after a terminal failure:
  - `guidellm benchmark ... --profile constant --rate 1 --rate 5 --rate 10`

## Related Issues

- Resolves https://github.com/vllm-project/guidellm/issues/604

---

## Use of AI

---

# git log

commit 577e1b9aaedada851bce96931aef4aaa6896f999
Date:   Sun Feb 15 13:56:30 2026 +0200

    Add cross-rate early exit for multi-rate benchmark profiles
    
    When running multiple rates (constant, poisson, concurrent profiles) or
    sweeping, stop escalating to higher rates if a failure constraint
    (over-saturation, max errors, error rate) triggers at a lower rate.
    
    - Sort rates/streams ascending in AsyncProfile and ConcurrentProfile
    - Add _should_stop_escalating() on base Profile class using stop_all
      as the failure signal (vs stop_local for normal completions)
    - Skip failure check after throughput phase in SweepProfile since
      over-saturation is expected at maximum load
    - Log warning when rate order is changed by sorting
    - Update CLI help and README with multi-rate documentation
    - Add comprehensive unit tests for all profile types
    
commit db623461e0ca2a90037828eb8a9cf8672d452720
Date:   Mon Jun 29 16:40:30 2026 +0300

    Address review feedback: fix docs, comments, and docstring
    
    - Simplify _should_stop_escalating param docstring (not a precondition)
    - Consolidate sweep.py early-exit comments into one clear block
    - Revert README multi-rate sentence, move docs to getting-started/benchmark.md
    - Add early-exit behavior notes to Concurrent, Constant, Poisson, Sweep sections
    
commit 37895189fc92fa440a134d67957810d347ae0d80
Date:   Wed Jul 8 12:46:36 2026 +0300

    CR
    
commit 95b52fd5d707f80915005061dfd93f3ec0877bdc
Date:   Wed Jul 8 13:34:14 2026 +0300

    fix tests
    
commit d98ba5d99615f7671ffdce180bf899457cee999b
Date:   Thu Jul 9 20:49:15 2026 +0300

    CR
    
commit 1e429538e22d77b2ac94d1a86d40982a9987edf5
Date:   Sun Jul 12 10:37:56 2026 +0300

    CR
    
commit 3a3a8479785a4ebdb92257c4dc5fd73dcb71d340
Date:   Tue Jul 14 20:40:28 2026 +0300

    Update src/guidellm/scheduler/constraints/saturation.py
    
    Co-authored-by: SkiHatDuckie <63932363+SkiHatDuckie@users.noreply.github.com>

commit d058ff2d324dddc0c33a13d0aefdd2c12539d888
Date:   Tue Jul 14 21:00:02 2026 +0300

    fix lint
    
commit ce09cdad9864cb4b71141fbd813435f42443a6f2
Date:   Tue Jul 14 21:03:52 2026 +0300

    fix lint
    
---------

Co-authored-by: SkiHatDuckie <63932363+SkiHatDuckie@users.noreply.github.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ guidellm run \
 **Key parameters:**
 
 - `--profile kind=<type>`: Defines the traffic pattern — `synchronous`, `concurrent`, `throughput`, `constant`, `poisson`, or `sweep`
-- `--profile kind=constant,rate=10`: For `constant`/`poisson`, set requests per second in the profile config; for `concurrent`, use `streams=`; for `throughput`, use `max_concurrency=`. For `constant`, `poisson`, and `concurrent`, multiple rates can be specified as a JSON list (e.g., `rate=[1,5,10]`). Rates are sorted ascending, and if a failure constraint (over-saturation, errors) triggers at a given rate, remaining higher rates are skipped.
+- `--profile kind=constant,rate=10`: For `constant`/`poisson`, set requests per second in the profile config; for `concurrent`, use `streams=`; for `throughput`, use `max_concurrency=`
 - `--constraint kind=max_duration,seconds=<seconds>` or `--constraint kind=max_requests,count=<count>`: Limit each strategy by time or request count
 
 ### Dataset Sources
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ guidellm run \
 **Key parameters:**
 
 - `--profile kind=<type>`: Defines the traffic pattern — `synchronous`, `concurrent`, `throughput`, `constant`, `poisson`, or `sweep`
-- `--profile kind=constant,rate=10`: For `constant`/`poisson`, set requests per second in the profile config; for `concurrent`, use `streams=`; for `throughput`, use `max_concurrency=`
+- `--profile kind=constant,rate=10`: For `constant`/`poisson`, set requests per second in the profile config; for `concurrent`, use `streams=`; for `throughput`, use `max_concurrency=`. For `constant`, `poisson`, and `concurrent`, multiple rates can be specified as a JSON list (e.g., `rate=[1,5,10]`). Rates are sorted ascending, and if a failure constraint (over-saturation, errors) triggers at a given rate, remaining higher rates are skipped.
 - `--constraint kind=max_duration,seconds=<seconds>` or `--constraint kind=max_requests,count=<count>`: Limit each strategy by time or request count
 
 ### Dataset Sources
--- a/docs/getting-started/benchmark.md
+++ b/docs/getting-started/benchmark.md
@@ -178,7 +178,7 @@ guidellm run --profile kind=poisson,rate=16 --seed kind=static,value=42
 | `rate`            | Target rate(s) in requests per second   | `--profile kind=poisson,rate=10` or `--profile '{"kind":"poisson","rate":[10,20]}'` |
 | `max_concurrency` | Maximum concurrent requests to schedule | `--profile kind=poisson,rate=10,max_concurrency=32`                                 |
 
-When multiple rates are specified, they are sorted in ascending order. If a failure constraint (such as over-saturation or excessive errors) triggers at a given rate, all remaining higher rates are skipped.
+When multiple rates are specified and a constraint with `stopping_scope="all"` is configured (e.g., over-saturation or error rate limits), rates must be in ascending order. If such a constraint triggers at a given rate, remaining rates are skipped.
 
 Use `--seed kind=static,value=42` for reproducible Poisson scheduling.
 
--- a/docs/getting-started/benchmark.md
+++ b/docs/getting-started/benchmark.md
@@ -178,7 +178,7 @@ guidellm run --profile kind=poisson,rate=16 --seed kind=static,value=42
 | `rate`            | Target rate(s) in requests per second   | `--profile kind=poisson,rate=10` or `--profile '{"kind":"poisson","rate":[10,20]}'` |
 | `max_concurrency` | Maximum concurrent requests to schedule | `--profile kind=poisson,rate=10,max_concurrency=32`                                 |
 
-When multiple rates are specified and a constraint with `stopping_scope="all"` is configured (e.g., over-saturation or error rate limits), rates must be in ascending order. If such a constraint triggers at a given rate, remaining rates are skipped.
+When multiple rates are specified and a constraint with `stopping_scope="all"` is configured (e.g., over-saturation or error rate limits), triggering that constraint will skip all remaining rates in the list. For best results, specify rates in ascending order so that only higher (harder) rates are skipped. With non-ascending order, easier rates that follow a failed rate will also be skipped.
 
 Use `--seed kind=static,value=42` for reproducible Poisson scheduling.
 
--- a/docs/getting-started/benchmark.md
+++ b/docs/getting-started/benchmark.md
@@ -190,7 +190,7 @@ The sweep profile applies a sequence of benchmark strategies to find the optimal
 
 1. It runs a `synchronous` strategy to measure the baseline rate,
 2. then runs a `throughput` strategy to determine peak throughput,
-3. and finally runs a series of asynchronous strategies with rates interpolated between the baseline and maximum throughput. (The number of interpolated strategies is `sweep_size` minus 2.) The asynchronous strategy type is determined by the `strategy_type` profile parameter. The default strategy type is `constant`. During the async phase, if a failure constraint triggers at a given rate, all remaining higher rates are skipped.
+3. and finally runs a series of asynchronous strategies with rates interpolated between the baseline and maximum throughput. (The number of interpolated strategies is `sweep_size` minus 2.) The asynchronous strategy type is determined by the `strategy_type` profile parameter. The default strategy type is `constant`. During the async phase, if a constraint with `stopping_scope="all"` triggers at a given rate, remaining rates are skipped.
 
 For example, to run a sweep with 10 strategies, 10 seconds of rampup, and a strategy type of `poisson`:
 
--- a/docs/getting-started/benchmark.md
+++ b/docs/getting-started/benchmark.md
@@ -178,6 +178,8 @@ guidellm run --profile kind=poisson,rate=16 --seed kind=static,value=42
 | `rate`            | Target rate(s) in requests per second   | `--profile kind=poisson,rate=10` or `--profile '{"kind":"poisson","rate":[10,20]}'` |
 | `max_concurrency` | Maximum concurrent requests to schedule | `--profile kind=poisson,rate=10,max_concurrency=32`                                 |
 
+When multiple rates are specified, they are sorted in ascending order. If a failure constraint (such as over-saturation or excessive errors) triggers at a given rate, all remaining higher rates are skipped.
+
 Use `--seed kind=static,value=42` for reproducible Poisson scheduling.
 
 You can use the `--override` option to specify a list of rate values, to run a set of Poisson "strategies" (sub-benchmarks) with different rate values. For example, `--profile kind=poisson --override profile.rate 10,20,30` will run a Poisson strategy with 10 requests per second, a Poisson strategy with 20 requests per second, and a Poisson strategy with 30 requests per second.
@@ -188,7 +190,7 @@ The sweep profile applies a sequence of benchmark strategies to find the optimal
 
 1. It runs a `synchronous` strategy to measure the baseline rate,
 2. then runs a `throughput` strategy to determine peak throughput,
-3. and finally runs a series of asynchronous strategies with rates interpolated between the baseline and maximum throughput. (The number of interpolated strategies is `sweep_size` minus 2.) The asynchronous strategy type is determined by the `strategy_type` profile parameter. The default strategy type is `constant`.
+3. and finally runs a series of asynchronous strategies with rates interpolated between the baseline and maximum throughput. (The number of interpolated strategies is `sweep_size` minus 2.) The asynchronous strategy type is determined by the `strategy_type` profile parameter. The default strategy type is `constant`. During the async phase, if a failure constraint triggers at a given rate, all remaining higher rates are skipped.
 
 For example, to run a sweep with 10 strategies, 10 seconds of rampup, and a strategy type of `poisson`:
 
--- a/src/guidellm/benchmark/profiles/asynchronous.py
+++ b/src/guidellm/benchmark/profiles/asynchronous.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, Any, Literal
 from pydantic import Field, PositiveFloat, PositiveInt, field_validator
 
 from guidellm.benchmark.schemas import ProfileArgs
+from guidellm.logger import logger
 from guidellm.scheduler import (
     AsyncConstantStrategy,
     AsyncPoissonStrategy,
@@ -56,7 +57,12 @@ class AsyncProfileArgs(ProfileArgs):
         if not value:
             raise ValueError("rate requires at least one value")
         if isinstance(value, list | tuple):
-            return value
+            sorted_rates = sorted(value)
+            if list(sorted_rates) != list(value):
+                logger.warning(
+                    f"Rates reordered from {list(value)} to {sorted_rates} (ascending)"
+                )
+            return sorted_rates
         if isinstance(value, int | float):
             return [value]
         raise ValueError(
@@ -107,15 +113,22 @@ class AsyncProfile(Profile):
         """
         Generate async strategy for next configured rate.
 
-        :param prev_strategy: Previously completed strategy (unused)
-        :param prev_benchmark: Benchmark results from previous execution (unused)
+        Rates are sorted ascending, so if a previous rate was terminated by a
+        failure constraint (over-saturation, errors, etc.), all remaining higher
+        rates are skipped.
+
+        :param prev_strategy: Previously completed strategy
+        :param prev_benchmark: Benchmark results from previous execution
         :return: AsyncConstantStrategy or AsyncPoissonStrategy for next rate,
-            or None if all rates completed
+            or None if all rates completed or failure detected
         :raises ValueError: If strategy_type is neither 'constant' nor 'poisson'
         """
-        _ = (prev_strategy, prev_benchmark)  # unused
+        _ = prev_strategy
 
         if len(self.completed_strategies) >= len(self.args.rate):
+            return None
+
+        if prev_benchmark is not None and self._should_stop_escalating(prev_benchmark):
             return None
 
         current_rate = self.args.rate[len(self.completed_strategies)]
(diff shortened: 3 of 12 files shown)


src/guidellm/benchmark/profiles/profile.py
```diff
@@ -115,6 +116,33 @@ def strategy_types(self) -> list[str]:
...
+        """
+        Check if a benchmark was terminated by a failure constraint.
+
+        Inspects the scheduler state's end_queuing_constraints for any constraint
+        that used "stop_all" for request processing, which indicates the system
+        could not handle the load (over-saturation, excessive errors, etc.).
+        Constraints that use "stop_local" (max duration, max requests) are normal
+        completions and do not trigger escalation stops.
+
+        :param prev_benchmark: Benchmark instance with a scheduler_state attribute
```
dbutenhof: Technically, "with a scheduler_state attribute" is not an entry predicate since you check safely for the attribute and for a non-None value. The only real requirement is a non-None benchmark value. Just "Benchmark instance" is good enough here.

src/guidellm/benchmark/profiles/sweep.py
```diff
@@ -129,6 +131,18 @@ def next_strategy(
                     self.args.sweep_size - 1,
                 )
             )[1:]  # don't rerun synchronous
+            # After throughput, fall through to async rate logic below.
+            # Don't check escalation since throughput is designed to push
+            # beyond sustainable load (over-saturation is expected).
```
dbutenhof: This comment at the tail of the indented block looks weird, and I don't think it's a good place. I think this should be combined with the following comment, and clarified. E.g., "the throughput guard above" doesn't have a `!= "throughput"` check and doesn't "skip" anything (unless it raises an error), so at best this is a bit misleading.

README.md
```diff
@@ -180,7 +180,7 @@ guidellm run \
 **Key parameters:**
 
 - `--profile kind=<type>`: Defines the traffic pattern — `synchronous`, `concurrent`, `throughput`, `constant`, `poisson`, or `sweep`
-- `--profile kind=constant,rate=10`: For `constant`/`poisson`, set requests per second in the profile config; for `concurrent`, use `streams=`; for `throughput`, use `max_concurrency=`
+- `--profile kind=constant,rate=10`: For `constant`/`poisson`, set requests per second in the profile config; for `concurrent`, use `streams=`; for `throughput`, use `max_concurrency=`. For `constant`, `poisson`, and `concurrent`, multiple rates can be specified as a JSON list (e.g., `rate=[1,5,10]`). Rates are sorted ascending, and if a failure constraint (over-saturation, errors) triggers at a given rate, remaining higher rates are skipped.
```
dbutenhof: "sorted [in] ascending [order]".

I don't think this is the right context for the "multiple rates can be specified" sentence, which complicates the bullet item here. "JSON list" is both too specific and also misleading since with v0.7.0 we also have `--override profile.streams 10,20,30`-- which may become the "recommended" mechanism specifically because it *doesn't* require turning the whole profile specification into a full JSON/YAML block. (E.g., because our simplified CLI key=value parsing doesn't [easily] allow a list.)

This bit should probably be worked into the `getting-started/benchmark.md` section on "Benchmark Profiles" instead.

src/guidellm/benchmark/profiles/profile.py
ushaket: Good point, simplified to "Benchmark instance"

src/guidellm/benchmark/profiles/sweep.py
ushaket: Agreed, consolidated into a single comment that explains the full logic

README.md
ushaket: Reverted the README line and moved the multi-rate / early-exit documentation into docs/getting-started/benchmark.md under the Concurrent, Constant, Poisson, and Sweep profile sections.

src/guidellm/benchmark/profiles/asynchronous.py
```diff
@@ -56,7 +57,12 @@ def _coerce_rate_to_list(
         if not value:
             raise ValueError("rate requires at least one value")
         if isinstance(value, list | tuple):
-            return value
+            sorted_rates = sorted(value)
+            if list(sorted_rates) != list(value):
+                logger.warning(
+                    f"Rates reordered from {list(value)} to {sorted_rates} (ascending)"
+                )
+            return sorted_rates
```
sjmonson: You cannot reorder rates. We have use-cases for running benchmarks in different orders then ascending.

src/guidellm/benchmark/profiles/concurrent.py
```diff
@@ -48,7 +49,13 @@ def _coerce_streams_to_list(cls, value: Any) -> Any:
...
             raise ValueError("streams requires at least one value")
         if isinstance(value, list | tuple):
-            return [int(stream) for stream in value]
+            streams = [int(stream) for stream in value]
+            sorted_streams = sorted(streams)
+            if sorted_streams != streams:
+                logger.warning(
+                    f"Streams reordered from {streams} to {sorted_streams} (ascending)"
+                )
+            return sorted_streams
```
sjmonson: You cannot reorder streams. We have use-cases for running benchmarks in different orders then ascending.

src/guidellm/benchmark/profiles/profile.py
```diff
@@ -115,6 +116,33 @@ def strategy_types(self) -> list[str]:
...
+        scheduler_state = getattr(prev_benchmark, "scheduler_state", None)
+        if scheduler_state is None:
+            return False
+
+        for name, action in scheduler_state.end_queuing_constraints.items():
+            if action.request_processing == "stop_all":
+                logger.info(
+                    f"Stopping rate escalation: constraint '{name}' "
+                    f"triggered (request_processing=stop_all)"
+                )
```
sjmonson: stop_all is not for stopping all benchmarks, its for stopping all distributed processes in a multi-node benchmark. We still plan to support that so stop_all  cannot be hijacked for this.

src/guidellm/benchmark/profiles/sweep.py
```diff
@@ -130,6 +132,16 @@ def next_strategy(
                 )
             )[1:]  # don't rerun synchronous
 
+        # Stop escalation if a failure constraint was triggered during the
+        # async phase. Throughput is excluded (type_ != "throughput") because
+        # it intentionally pushes beyond sustainable load. Synchronous never
+        # reaches here — it returns a ThroughputStrategy above.
+        if (
+            prev_strategy.type_ != "throughput"
+            and self._should_stop_escalating(prev_benchmark)  # type: ignore[arg-type]
```
sjmonson, line 140: This behavior should be off by default. Probably configurable per-constraint is best.

sjmonson (changes requested):
Comments below are copies of the feedback I left over slack. Regarding `stop_all` I think we need a new field, maybe `stopping_scope: Literal["current", "all"]` ideally this is defined on `ConstraintArgs` and not the scheduler state but in order to guarantee you are mapping to the correct constraint it might have to be in both places. In which case each constraint should pass its value of `stopping_scope` from its `ConstraintArgs` into the scheduler state it maintains.

src/guidellm/benchmark/profiles/asynchronous.py
ushaket: Reverted sorting. User order preserved; ascending validated only when stopping_scope="all" is configured.

src/guidellm/benchmark/profiles/concurrent.py
ushaket: Reverted sorting. User order preserved; ascending validated only when stopping_scope="all" is configured.

src/guidellm/benchmark/profiles/profile.py
ushaket: Fixed — added stopping_scope field on ConstraintArgs (defaults to "current"), propagated to SchedulerUpdateAction. _should_stop_escalating now checks that instead of stop_all.

src/guidellm/benchmark/profiles/sweep.py
ushaket: Done — early-exit is opt-in per-constraint via stopping_scope="all".

ushaket:
Thanks @sjmonson

Implemented stopping_scope on ConstraintArgs as suggested, propagated to SchedulerUpdateAction. All constraint types pass it through, default is "current". Removed rate/stream sorting — user order is preserved. When stopping_scope="all" is configured, ascending order is validated at profile construction.

src/guidellm/benchmark/profiles/concurrent.py
```diff
@@ -78,6 +78,16 @@ def __init__(
...
+        if (
+            len(args.streams) > 1
+            and self._has_escalation_stopping()
+            and list(args.streams) != sorted(args.streams)
+        ):
+            raise ValueError(
+                "Early-exit (stopping_scope='all') requires streams in ascending "
+                f"order, got {list(args.streams)}"
+            )
+
```
sjmonson:
```suggestion
```

Let the user do what they want 😁

src/guidellm/benchmark/profiles/profile.py
```diff
@@ -115,6 +124,31 @@ def strategy_types(self) -> list[str]:
...
+        scheduler_state = getattr(prev_benchmark, "scheduler_state", None)
+        if scheduler_state is None:
+            return False
+
+        for name, action in scheduler_state.end_queuing_constraints.items():
+            if action.stopping_scope == "all":
+                logger.info(
+                    f"Stopping rate escalation: constraint '{name}' "
+                    f"triggered (stopping_scope=all)"
+                )
```
sjmonson: Small nit but no f-string log messages. See #902.

sjmonson (changes requested):

src/guidellm/benchmark/profiles/concurrent.py
jaredoconnell: Yeah. Instead, document the downside to doing it out of order, like how it would skip easier following benchmarks after aborting one that overloads the server.

docs/getting-started/benchmark.md
```diff
@@ -178,6 +178,8 @@ guidellm run --profile kind=poisson,rate=16 --seed kind=static,value=42
 | `rate`            | Target rate(s) in requests per second   | `--profile kind=poisson,rate=10` or `--profile '{"kind":"poisson","rate":[10,20]}'` |
 | `max_concurrency` | Maximum concurrent requests to schedule | `--profile kind=poisson,rate=10,max_concurrency=32`                                 |
 
+When multiple rates are specified and a constraint with `stopping_scope="all"` is configured (e.g., over-saturation or error rate limits), rates must be in ascending order. If such a constraint triggers at a given rate, remaining rates are skipped.
```
dbutenhof: "should be in ascending order", not "must" ... the following sentence more or less establishes *why*, but more detail might not hurt.

src/guidellm/benchmark/profiles/asynchronous.py
```diff
@@ -92,6 +92,16 @@ def __init__(
...
 
+        if (
+            len(args.rate) > 1
+            and self._has_escalation_stopping()
+            and list(args.rate) != sorted(args.rate)
+        ):
+            raise ValueError(
+                "Early-exit (stopping_scope='all') requires rates in ascending "
+                f"order, got {list(args.rate)}"
+            )
```
dbutenhof: There's no reason to *require* this, although it definitely makes sense to **recommend** it and explain why.

src/guidellm/benchmark/profiles/profile.py
```diff
@@ -101,6 +102,14 @@ def __init__(
         self.constraints = dict(constraints or {})
         self.completed_strategies: list[SchedulingStrategy] = []
 
+    def _has_escalation_stopping(self) -> bool:
+        """Check if any constraint is configured with stopping_scope='all'."""
+        for initializer in self.constraints.values():
+            args = getattr(initializer, "args", None)
+            if args is not None and getattr(args, "stopping_scope", "current") == "all":
+                return True
+        return False
```
dbutenhof: I think a cleaner way to do this would be to add something like a `should_terminate() -> bool` method to the constraint base class, (I think at this level we have constraints rather than constraint args, but it doesn't really matter), with a default `return False`, and override that for the error constraints... rather than all the `getattr` hackery here outside the framework of the O-O model. Then here you'd just iterate through the active constraints and ask each of them.

src/guidellm/benchmark/profiles/profile.py
dbutenhof: I'm not sure this really needs to be at `info` level, either. Maybe `debug`?

dbutenhof (dismissed):
This sounds like a good idea. Although there's a voice in the back of my head with a weird concern about arbitrarily sorting the rate/streams list ... I also can't convince myself there's any reasonable motivation for requiring an out-of-order list.

src/guidellm/benchmark/profiles/concurrent.py
ushaket: Done — removed validation, added docs explaining the trade-off (non-ascending + early-exit skips subsequent rates regardless).

src/guidellm/benchmark/profiles/profile.py
ushaket: Fixed — switched to deferred {} formatting.

docs/getting-started/benchmark.md
ushaket: Already addressed — current text says "For best results, specify rates in ascending order" (recommendation, not requirement).

src/guidellm/benchmark/profiles/asynchronous.py
ushaket: Agreed — validation removed. Users can run any order; docs explain the trade-off.

src/guidellm/benchmark/profiles/profile.py
ushaket: Removed — that _has_escalation_stopping() method with the getattr hackery was deleted. The remaining _should_stop_escalating reads the typed stopping_scope field from SchedulerUpdateAction in the post-execution scheduler state (no getattr on constraints)

ushaket:
Thanks @sjmonson @jaredoconnell @dbutenhof, I addressed the comments, please review again 🙏

docs/getting-started/benchmark.md
```diff
@@ -188,7 +190,7 @@ The sweep profile applies a sequence of benchmark strategies to find the optimal
 
 1. It runs a `synchronous` strategy to measure the baseline rate,
 2. then runs a `throughput` strategy to determine peak throughput,
-3. and finally runs a series of asynchronous strategies with rates interpolated between the baseline and maximum throughput. (The number of interpolated strategies is `sweep_size` minus 2.) The asynchronous strategy type is determined by the `strategy_type` profile parameter. The default strategy type is `constant`.
+3. and finally runs a series of asynchronous strategies with rates interpolated between the baseline and maximum throughput. (The number of interpolated strategies is `sweep_size` minus 2.) The asynchronous strategy type is determined by the `strategy_type` profile parameter. The default strategy type is `constant`. During the async phase, if a failure constraint triggers at a given rate, all remaining higher rates are skipped.
```
jaredoconnell: Is this still true given that stopping scope's default is now `current`?

sjmonson (dismissed):
Tested and works for me

docs/getting-started/benchmark.md
ushaket: Good catch — updated to clarify it requires stopping_scope="all".

dbutenhof (dismissed):

src/guidellm/scheduler/constraints/saturation.py
SkiHatDuckie, line 397:
```suggestion
        :param stopping_scope: Whether to halt only the current benchmark or also prevent escalation to subsequent rates/streams.
            (default: "current")
       """
```
Not that important since it seems like `stopping_scope` is just here to be passed to a `SchedulerUpdateAction`.

jaredoconnell (dismissed):

SkiHatDuckie (approved):
Just a minor nitpick in constraints/saturation.py. Otherwise, all seems to be working for me.

SkiHatDuckie:
Linter is complaining. That's on me, the trailing docstring `"""` is missing a space it seems.

dbutenhof (approved):

dbutenhof:
@mergifyio queue

jaredoconnell (approved):

PR #607, opened 2026-02-23
Instant ttft oversaturation

### LINKED ISSUE #606: Over-saturation detection stops prematurely with long contexts
### Problem Statement

When using `--detect-saturation` with large models and long input contexts (e.g., 10,000 tokens), the benchmark stops prematurely before any request completes.

### Root Cause

The over-saturation constraint uses both concurrent request slope and TTFT slope to detect saturation. TTFT data is only communicated to the constraint when a request fully completes.

With large models and long contexts, no request completes within the `minimum_duration` window (default: 30s), so the constraint has zero TTFT data and falls back to concurrent slope alone — which naturally rises during ramp-up.

The TTFT value is available much earlier — the backend knows it as soon as the first token arrives during streaming — but it is not communicated back to the main process until the request completes.

### Proposed Solution

When over-saturation detection is enabled, send a bounded number of early TTFT notifications from the worker process when the first token arrives during streaming, before the request completes. This gives the constraint real TTFT data to make a two-signal decision instead of falling back to concurrent slope alone.

The count of early notifications should be configurable and small (e.g., 5 per worker) to limit multi-process queue overhead. When over-saturation detection is not enabled, no extra notifications are sent and behavior is unchanged.

### Alternatives Considered

_No response_

### Usage Examples

```markdown

```

### Additional Context

_No response_

ushaket:
## Summary

When over-saturation detection is enabled (`--detect-saturation`), the constraint can only receive TTFT data after a request fully completes. With large models and long contexts, no request completes within the `minimum_duration` window (default 30s), so the constraint falls back to concurrent slope alone and stops prematurely.

This PR adds time-bounded instant TTFT notifications: when over-saturation detection is enabled, worker processes monitor for first-token arrival during streaming and send a `"first_token_arrived"` status update before the request completes. This gives the constraint real TTFT data for a two-signal decision. Notifications are sent only during the first `minimum_duration` seconds of the benchmark to limit IPC overhead.

## Details

- [x] Add `"first_token_arrived"` to `RequestInfo.status` literal (`schemas/info.py`)
- [x] Add TTFT polling monitor to `WorkerProcess` — spawns an async task per request that detects `first_token_iteration` and sends a `"first_token_arrived"` update (`scheduler/worker.py`)
- [x] Time-bound the monitor: notifications stop after `minimum_duration` seconds via `instant_ttft_duration` (`scheduler/worker.py`)
- [x] Handle `"first_token_arrived"` in `WorkerGroupState` — no request count changes, passes through to constraints (`scheduler/worker_group.py`)
- [x] Extract `minimum_duration` from `OverSaturationConstraint` to configure worker TTFT duration (`scheduler/worker_group.py`)
- [x] Accept TTFT from both `"first_token_arrived"` and `"completed"` in the constraint, deduplicated by request ID (`scheduler/constraints/saturation.py`)
- [x] Add 8 tests covering happy path, dedup, missing timings, backward compatibility, concurrent isolation, disabled mode, reset, and multi-request slope building

## Test Plan

- Run `pytest tests/unit/scheduler/ tests/unit/schemas/ tests/unit/backends/` — 1077 passed
- Run `pre-commit run --files` on changed files — all checks pass
- Verify with `--detect-saturation` on a large model with long context (>10k tokens) that the benchmark no longer stops prematurely

## Related Issues

- Resolves #606

---

## Use of AI

### DIFF
--- a/src/guidellm/backends/openai/http.py
+++ b/src/guidellm/backends/openai/http.py
@@ -279,7 +279,7 @@ class OpenAIHTTPBackend(Backend):
         request: GenerationRequest,
         request_info: RequestInfo,
         history: list[tuple[GenerationRequest, GenerationResponse]] | None = None,
-    ) -> AsyncIterator[tuple[GenerationResponse, RequestInfo]]:
+    ) -> AsyncIterator[tuple[GenerationResponse | None, RequestInfo]]:
         """
         Process generation request and yield progressive responses.
 
--- a/src/guidellm/scheduler/schemas.py
+++ b/src/guidellm/scheduler/schemas.py
@@ -119,14 +119,15 @@ class BackendInterface(Protocol, Generic[RequestT, ResponseT]):
         request: RequestT,
         request_info: RequestInfo,
         history: list[tuple[RequestT, ResponseT]] | None = None,
-    ) -> AsyncIterator[tuple[ResponseT, RequestInfo]]:
+    ) -> AsyncIterator[tuple[ResponseT | None, RequestInfo]]:
         """
         Process a request and yield incremental response updates.
 
         :param request: The request object to process
         :param request_info: Scheduling metadata and timing information
         :param history: Conversation history for multi-turn requests
-        :yield: Tuples of (response, updated_request_info) for each response chunk
+        :yield: Tuples of (response, updated_request_info) for each response chunk.
+            Response may be None for intermediate updates (e.g., first token arrival).
         :raises Exception: Implementation-specific exceptions for processing failures
         """
 


src/guidellm/scheduler/worker_group.py
```diff
@@ -126,6 +127,14 @@ def __init__(
...
         ) = None
 
+    def _resolve_instant_ttft_duration(self) -> float:
+        """Extract instant TTFT duration from over-saturation constraints."""
+        duration = 0.0
+        for c in self.constraints.values():
+            if isinstance(c, OverSaturationConstraint):
+                duration = max(duration, c.minimum_duration)
+        return duration
+
```
sjmonson: Delete this. Optional constraints should not be reaching into the scheduler logic and the ttft return condition is unnecessary.

src/guidellm/scheduler/worker_group.py
```diff
@@ -210,6 +219,8 @@ async def create_processes(self):
                 poll_interval=settings.mp_poll_interval,
             )
 
+        instant_ttft_duration = self._resolve_instant_ttft_duration()
+
```
sjmonson: Delete this. Optional constraints should not be reaching into the scheduler logic and the ttft return condition is unnecessary.

src/guidellm/scheduler/worker_group.py
```diff
@@ -237,6 +248,7 @@ async def create_processes(self):
                 constraint_reached_event=self.constraint_reached_event,
                 shutdown_event=self.shutdown_event,
                 error_event=self.error_event,
+                instant_ttft_duration=instant_ttft_duration,
```
sjmonson: Delete this. Optional constraints should not be reaching into the scheduler logic and the ttft return condition is unnecessary.

src/guidellm/scheduler/worker.py
```diff
@@ -396,6 +418,43 @@ async def _process_next_request(self, target_start: float):
...
+        constraint receives TTFT data before the request fully completes.
+
+        :param request: The request being processed
+        :param request_info: Shared request info mutated by the backend
+        """
+        while request_info.timings.first_token_iteration is None:  # noqa: ASYNC110
+            await asyncio.sleep(1.0)
+
+        self._send_update("first_token_arrived", None, request, request_info)
+
```
sjmonson: I don't understand this. Modify the backend to return an event on ttft and Just check in the status of `request_info.timings.first_token_iteration` in the resolve async loop. No need to create a weird side-car wait.

src/guidellm/scheduler/worker.py
```diff
@@ -428,7 +487,12 @@ async def _schedule_request(
     def _send_update(
         self,
         new_status: Literal[
-            "pending", "in_progress", "completed", "errored", "cancelled"
+            "pending",
+            "in_progress",
+            "first_token_arrived",
```
sjmonson: Too verbose. Rename to `"first_token"`

src/guidellm/scheduler/worker.py
```diff
@@ -98,6 +99,7 @@ def __init__(
         constraint_reached_event: ProcessingEvent,
         shutdown_event: ProcessingEvent,
         error_event: ProcessingEvent,
+        instant_ttft_duration: float = 0.0,
```
sjmonson: Drop

src/guidellm/scheduler/worker.py
```diff
@@ -114,6 +116,11 @@ def __init__(
         :param constraint_reached_event: Event signaling processing constraint reached
         :param shutdown_event: Event signaling graceful shutdown request
         :param error_event: Event signaling error conditions across processes
+        :param instant_ttft_duration: Duration in seconds during which the worker
+            sends early "first_token_arrived" status updates when the first token
+            arrives during streaming, before the request completes. Set to 0.0 to
+            disable. Used to provide TTFT data to the over-saturation constraint
+            before any request finishes.
```
sjmonson: Drop

src/guidellm/scheduler/worker.py
```diff
@@ -126,11 +133,13 @@ def __init__(
         self.constraint_reached_event = constraint_reached_event
         self.shutdown_event = shutdown_event
         self.error_event = error_event
+        self.instant_ttft_duration = instant_ttft_duration
```
sjmonson: Drop

src/guidellm/scheduler/worker.py
```diff
@@ -126,11 +133,13 @@ def __init__(
...
 
         # Internal states
         self.startup_completed = False
         self.backend_started = False
         self.messaging_started = False
+        self._instant_ttft_start: float | None = None
```
sjmonson: Drop

src/guidellm/scheduler/worker.py
```diff
@@ -396,6 +418,43 @@ async def _process_next_request(self, target_start: float):
...
+        Monitors ``request_info.timings.first_token_iteration`` which is set
+        in-place by the backend during streaming. When detected, sends a
+        ``"first_token_arrived"`` status update so that the over-saturation
+        constraint receives TTFT data before the request fully completes.
+
+        :param request: The request being processed
+        :param request_info: Shared request info mutated by the backend
+        """
+        while request_info.timings.first_token_iteration is None:  # noqa: ASYNC110
+            await asyncio.sleep(1.0)
```
sjmonson: This is a hack. It relies on the backend reusing the same object for each status update. We do not make that guarantee. If you need this information the backend should be modified to send it as part of `resolve`.

src/guidellm/scheduler/worker.py
ushaket: NP, 
That was my initial approach, but I noticed the backend yields just once, it felt very intrusive to start yielding twice

src/guidellm/scheduler/worker_group.py
ushaket: We had some performance concerns and we agreed to have this behind a flag (oversaturation is on) and bounded, otherwise we always call _send_update("first_token", ...)

*I went with time bound since it makes most sense (when final responses starts flowing we don't need to keep reporting the ttft immediately, and if we don't get ttft in minimum_duration we can say it's oversaturated)

Do you think we can base our condition on something else? or you're saying we can safely always call _send_update("first_token", ...)?

One more option is to add the flag in the backend (that would also address my concern about always yielding twice, if that concern is valid), but that means the backend needs to be aware of the oversaturation constraint

sjmonson: My concern was on sending back every token event. I didn't consider just the first token. I still think we need to validate that there is no performance regressions but I am not overly worried about it.

ushaket: Perfect,
Done

src/guidellm/scheduler/worker_group.py
ushaket: Done

src/guidellm/scheduler/worker_group.py
ushaket: Done

src/guidellm/scheduler/worker.py
ushaket: Done

src/guidellm/scheduler/worker.py
ushaket: Done

src/guidellm/scheduler/worker.py
ushaket: Done

src/guidellm/scheduler/worker.py
ushaket: Done

src/guidellm/scheduler/worker.py
ushaket: Done

src/guidellm/scheduler/worker.py
ushaket: Done

src/guidellm/scheduler/worker.py
ushaket: Done

src/guidellm/scheduler/worker.py
```diff
@@ -365,11 +365,19 @@ async def _process_next_request(self, target_start: float):
...
                 request_info = info
                 if request_info is None:
                     raise RuntimeError("Received invalid request info from backend")
 
+                if (
+                    resp is None
+                    and request_info.timings.first_token_iteration is not None
+                ):
+                    self._send_update("first_token", None, request, request_info)
+                    continue
```
sjmonson: Is this `continue` necessary? I think `response` should always be the most recently processed packet. Or did you see any issues to warrant skipping `response = resp`?

ushaket: The response is empty when engine resolves for first_token, continue was there to make the intent explicit ("this isn't a response, skip the assignment")

But I guess continue isn't necessary for correctness, and removing it keeps the logic simpler and treats all yields the same

Removed it

sjmonson (approved):
After some stress testing, no noticeable impact to scheduling performance. Did not test oversaturation detection but code changes seem reasonable.

PR #610, opened 2026-02-25
Add support for TerraTorch Geospatial models served via the vLLM /pooling endpoint

### LINKED ISSUE #587: Add support for Geospatial models (served via vLLM)

Add support in guideLLM to benchmark geospatial image-to-image models such as Prithvi and TerraMind.
Unlike LLM-style workloads:

- These models do not consume or produce tokens
- They take images as input and return images as output
- Inference requests are sent to a vLLM server at the /pooling endpoint

From a quick look, guideLLM benchmarks seem to be token-centric, so existing metrics do not correctly represent performance for this type of workload. In the case of Geospatial models token metrics are not relevant leaving requests/second and request latency as the main metrics.

As a first step, I would like to: 
- Add support for benchmarking via the /pooling endpoint
- Handle cases where input/output data contains no tokens
- Report performance using:
  - requests_per_second (1 request = 1 image)
  - request_latency (end-to-end image processing time)

**Additional context**
We currently benchmark these models using vLLM’s built-in benchmarking scripts with a custom backend. A prototype implementation is available in this ([fork](https://github.com/mgazz/vllm/tree/vllm_benchmarks)). 

Example command used today:
```
python -m vllm.entrypoints.cli.main bench serve \ 
    --base-url ${VLLM_ENDPOINT} \ 
    --dataset-name=custom \ 
    --model ibm-nasa-geospatial/Prithvi-EO-2.0-300M-TL-Sen1Floods11 \ 
    --skip-tokenizer-init true \ 
    --endpoint /pooling \ 
    --backend io-processor-plugin \ 
    --metric-percentiles 25,75,99 \ 
    --percentile-metrics e2el \ 
    --dataset-path ./benchmarks/dataset_url_input_india.jsonl \ 
    --num-prompts 5000 \ 
    --request-rate 10 \ 
    --max-concurrency 10 
```

mgazz:
## Summary

This PR adds support for TerraTorch Geospatial models like Prithvi. Compared to traditional LLMs, Geospatial Models support image-to-image or tensor-to-tensor input/output, but they do not process text. Moreover they are served in vLLM at the /pooling endpoint. 

## Details

- Support for the /pooling endpoint
- A custom PoolingColumnMapper to map the `prompt` entry in datasets without creating conflict with the existing GenerativeColumnMapper
- A new request handler for the pooling endpoint: PoolingRequestHandler

## Test Plan

-

## Related Issues

- Resolves #587 

---

## Use of AI

### DIFF
--- a/docs/guides/geospatial.md
+++ b/docs/guides/geospatial.md
@@ -50,6 +50,7 @@ Geospatial models served with vLLM expect a specific nested JSON structure in th
       "out_data_format": "b64_json",
       "indices": [1, 2, 3, 8, 11, 12]
     },
+    "priority": 0
   }
 }
 ```
@@ -62,6 +63,7 @@ Geospatial models served with vLLM expect a specific nested JSON structure in th
     - **data_format**: Format of input data (`"url"` or `"base64"`)
     - **out_data_format**: Desired output format (e.g., `"b64_json"`)
     - **indices**: Array of band/channel indices to process (specific to the satellite sensor)
+  - **priority**: (Optional) Request priority for scheduling (default: 0)
 
 ### Band Indices
 


docs/guides/geospatial.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Complete Example: Flood Detection with Prithvi
+
+Here's a complete example for benchmarking the Prithvi flood detection model:
+
+### 1. Prepare Dataset
+
+Create `flood_detection_dataset.jsonl` remembering to update the GeoTIFF url reference by `data` with one reachable over the network. You can find examples in the [Prithvi repository on HuggingFace](https://huggingface.co/ibm-nasa-geospatial/Prithvi-EO-2.0-300M-TL-Sen1Floods11/tree/main/examples).
+
+```json
```
dbutenhof: Since most of your example stream is very concrete, and your `guidellm` command references this data as `flood_detection_dataset.jsonl`, I think you should specify that name here with the example input. (Yeah, I don't blame you here for not using real URLs, which means the example won't run as shown anyway ... but we can at least avoid a few simple small confusions.)

Also, your code recognizes both `"data"` and `"priority"` fields inside the top level `"data"`, but your documentation doesn't mention that at all, and the examples don't include it.

dbutenhof (approved):
This looks fairly straightforward (even if the entire concept of geospatial inference is more than slightly mysterious to me).

I made a minor comment on your documentation, which might help readers to follow the sequence... but I think that's a minor issue and I'm approving anyway. (Not that it really matters as you need a rebase now.)

FYI, the unit test failure is a known issue, and will be resolved when you rebase.

src/guidellm/data/preprocessors/mappers.py
```diff
@@ -201,3 +201,8 @@ def setup_data(
         self.datasets_column_mappings = self.datasets_mappings(
             datasets, self.input_mappings
         )
+
+
+@PreprocessorRegistry.register("pooling_column_mapper")
+class PoolingColumnMapper(GenerativeColumnMapper):
```
jaredoconnell, line 207: I'm not sure this sub-type is the ideal design. It has an is-a relationship with the generative column mapper.
Would it make sense to add the pooling column to the generative column mapper class, or to create a new parent class, and add child classes for each column mapper?

src/guidellm/backends/openai/http.py
```diff
@@ -110,6 +111,7 @@ def validate_request_format(cls, v: str | None) -> str | None:
     "chat_completions": "/v1/chat/completions",
     "audio_transcriptions": "/v1/audio/transcriptions",
     "audio_translations": "/v1/audio/translations",
+    "pooling": "/pooling",
```
sjmonson: The `LEGACY_API_ALIASES` are only to avoid breaking pre-existing scripts. No new aliases should be added.

```suggestion
```

src/guidellm/data/preprocessors/mappers.py
sjmonson: I was originally thinking the same thing, but since they are only overriding one variable I think the current design is fine. If we made a parent class then in this case, all of the logic would live there and each class would just override `defaults` which is not any better.

src/guidellm/backends/openai/http.py
sjmonson: Will also have to update documentation since its used there.

docs/guides/geospatial.md
```diff
@@ -0,0 +1,151 @@
...
+
+To benchmark a geospatial model, use the `--request-format pooling` parameter:
+
+```bash
+guidellm benchmark \
+  --target http://localhost:8000 \
+  --model ibm-nasa-geospatial/Prithvi-EO-2.0-300M-TL-Sen1Floods11 \
+  --backend openai_http \
+  --data dataset.jsonl \
+  --request-format pooling \
```
dbutenhof: This needs to change, too, since the "pooling" alias is no longer valid.

docs/guides/geospatial.md
```diff
@@ -0,0 +1,151 @@
...
+
+### 3. Run Benchmark
+
+```bash
+guidellm benchmark \
+  --target http://localhost:8000 \
+  --model ibm-nasa-geospatial/Prithvi-EO-2.0-300M-TL-Sen1Floods11 \
+  --backend openai_http \
+  --data flood_detection_dataset.jsonl \
+  --request-format pooling \
```
dbutenhof: And here ...

docs/guides/geospatial.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Equivalent GuideLLM Command
+
+```bash
+guidellm benchmark \
+  --target http://localhost:8000 \
+  --model ibm-nasa-geospatial/Prithvi-EO-2.0-300M-TL-Sen1Floods11 \
+  --backend openai_http \
+  --data dataset.jsonl \
+  --request-format pooling \
```
dbutenhof: And here ...

docs/guides/geospatial.md
mgazz: With eeb9133a0c4e2fd1466eb6b53e6bed089f89cd90 I added a command to generate a valid dataset and referenced it in the rest of the documentation.

With 9bd337bca5e00285b78405ddb6a8568b136f65ca I added `priority` to the list of dataset entry fields. 

Please let me know if these changes fix the issues above.

docs/guides/geospatial.md
mgazz: Thank you for the references. With eeb9133a0c4e2fd1466eb6b53e6bed089f89cd90 I updated the commands with `--request-format /pooling`. The benchmark seem to run fine. Please let me know if I should specify the endpoint in a different way.

PR #615, opened 2026-02-26
Fix worker status for unusable terminal backend responses

### LINKED ISSUE #613: Worker marks completed/successful when final response has no usable data
### Problem Statement

- WorkerProcess._process_next_request always sends completed after backend.resolve(...) returns, without validating final response content.
- This can mark requests successful even when final response is missing (None) or empty (no choices/content/usage).
- Downstream this increments successful_requests and enters completed metrics.

### Proposed Solution

- If final resolve result has no usable output payload, request should be marked errored (or at minimum cancelled/incomplete, but not successful).
- Successful completion should require a valid terminal response contract.

### Alternatives Considered

_No response_

### Usage Examples

```markdown

```

### Additional Context

_No response_

ushaket:
## Summary

This PR fixes a scheduler correctness bug where requests could be marked as `completed` even when the backend resolved without a usable terminal response. It adds explicit terminal-response validation in the worker so malformed/empty terminal results are surfaced as `errored` with a clear diagnostic instead of being counted as successful requests.

## Details

- [x] Added terminal response validation in `WorkerProcess` to guard final status transitions.
- [x] Updated request finalization logic so unusable terminal responses set:
  - status: `errored`
  - error message: `[UNUSABLE_BACKEND_RESPONSE] backend resolved without a usable terminal response payload`
- [x] Implemented `GenerationResponse`-aware usability criteria:
  - usable if non-empty `text` **or** `output_metrics.total_tokens > 0`
  - `None` terminal response is always unusable
  - non-`GenerationResponse` fallback remains `bool(response)` for generic/test compatibility
- [x] Added regression tests in `tests/unit/scheduler/test_worker.py` for:
  - no terminal response object -> `errored`
  - empty `GenerationResponse` -> `errored`
  - token-bearing `GenerationResponse` with empty text -> `completed`
- [x] Preserved existing cancellation/error flow behavior outside terminal-response validation.

## Test Plan

- Run targeted worker regression tests:
  - `uv run pytest -q tests/unit/scheduler/test_worker.py -k "terminal_response or empty_generation_response or generation_response_with_tokens or invalid_initialization"`
- Verify expected outcomes:
  - missing terminal response is not marked `completed`
  - empty `GenerationResponse` is not marked `completed`
  - non-empty signal (`output_tokens > 0`) is accepted as `completed`
- Confirm no lint issues on touched files.

## Related Issues

- Resolves #613

---

## Use of AI

---

# git log

commit ffedece7bc093dcd18e4259c0d753f3cf03d7b27
Date:   Mon Mar 2 23:55:49 2026 +0200

    mvoe validation logic to RequestHandlers
    
commit 03cdec7a6954e3dcaa610a9c5b382966a8212098
Date:   Mon Mar 2 23:57:34 2026 +0200

    revert worker changes
    
commit e5fda3292f86e3f7892171a4106fb415504b66df
Date:   Wed Jul 15 11:25:12 2026 +0300

    fix tests
    
commit d459b2ef7eeda1bf1b0b8f0cf3655173163a2520
Date:   Wed Jul 15 13:23:05 2026 +0300

    self CR
    
commit 36036d0b0b7f67c111337f628e64173dfd5e69fc
Date:   Thu Jul 16 17:28:58 2026 +0300

    Clarify unusable backend response error message
    
commit 1789158ca2cedcf63d8ef55927ee77e77358c03a
Date:   Tue Jul 21 20:39:29 2026 +0300

    refactor: centralize post_validation hook for response validation
    
    Add post_validation to OpenAIRequestHandler and OpenAIWSRequestHandler
    protocols with a permissive no-op default. TextCompletionsRequestHandler
    and ResponsesRequestHandler override with a check for text, tool_calls,
    or output_tokens. PoolingRequestHandler and EmbeddingsRequestHandler
    keep the default no-op since they produce non-text output.
    
    Validation is called from the backend resolve methods (http.py,
    websocket.py) after compile, not inside each handler's compile methods.
    Cancellation paths skip validation for partial responses.
    
    Fixes tool-call-only false negative: responses with tool_calls but no
    text are now correctly treated as valid output.
    
    Assisted-by: Claude

commit 6de940a469e2bd55cbcacf96cef81d7a23c7047e
Date:   Thu Jul 23 09:44:58 2026 +0300

    refactor: deduplicate post_validation and add docstrings
    
    Extract _validate_text_response helper shared by
    TextCompletionsRequestHandler and ResponsesRequestHandler.
    Add docstrings to all post_validation overrides.
    
    Assisted-by: Claude

---------

Assisted-by: Claude

### DIFF
--- a/src/guidellm/backends/openai/request_handlers.py
+++ b/src/guidellm/backends/openai/request_handlers.py
@@ -341,6 +341,24 @@ def _apply_tool_call_metrics(
         output_metrics.mixed_content_tool_tokens = output_metrics.text_tokens
 
 
+def _validate_text_response(response: GenerationResponse) -> None:
+    """Reject a compiled response that has no usable output.
+
+    A response is considered usable if it has non-empty text, tool calls,
+    or output tokens. Used by text/chat completions and responses handlers.
+
+    :param response: The compiled generation response to validate.
+    :raises ValueError: If the response contains no usable output.
+    """
+    has_text = bool(response.text and response.text.strip())
+    has_tool_calls = bool(response.tool_calls)
+    output_tokens = response.output_metrics.total_tokens or 0
+    if not has_text and not has_tool_calls and output_tokens <= 0:
+        raise ValueError(
+            "[UNUSABLE_BACKEND_RESPONSE] backend resolved with empty response payload"
+        )
+
+
 _DEFAULT_REASONING_TEMPLATE = "<think>{reasoning}</think>"
 
 
@@ -596,14 +614,8 @@ class TextCompletionsRequestHandler(OpenAIRequestHandler):
         )
 
     def post_validation(self, response: GenerationResponse) -> None:
-        has_text = bool(response.text and response.text.strip())
-        has_tool_calls = bool(response.tool_calls)
-        output_tokens = response.output_metrics.total_tokens or 0
-        if not has_text and not has_tool_calls and output_tokens <= 0:
-            raise ValueError(
-                "[UNUSABLE_BACKEND_RESPONSE] backend resolved with empty "
-                "response payload"
-            )
+        """Reject responses with no text, tool calls, or output tokens."""
+        _validate_text_response(response)
 
     def extract_line_data(self, line: str) -> dict[str, Any] | None:
         """
@@ -1883,14 +1895,8 @@ class ResponsesRequestHandler(OpenAIRequestHandler):
         )
 
     def post_validation(self, response: GenerationResponse) -> None:
-        has_text = bool(response.text and response.text.strip())
-        has_tool_calls = bool(response.tool_calls)
-        output_tokens = response.output_metrics.total_tokens or 0
-        if not has_text and not has_tool_calls and output_tokens <= 0:
-            raise ValueError(
-                "[UNUSABLE_BACKEND_RESPONSE] backend resolved with empty "
-                "response payload"
-            )
+        """Reject responses with no text, tool calls, or output tokens."""
+        _validate_text_response(response)
 
     def extract_line_data(self, line: str) -> dict[str, Any] | None:
         """Parse a Responses API SSE line.
@@ -2192,7 +2198,7 @@ class PoolingRequestHandler(ChatCompletionsRequestHandler):
     """
 
     def post_validation(self, response: GenerationResponse) -> None:  # noqa: ARG002
-        pass
+        """Pooling responses produce non-text output; skip validation."""
 
     def format(
         self,
--- a/src/guidellm/backends/openai/request_handlers.py
+++ b/src/guidellm/backends/openai/request_handlers.py
@@ -581,8 +581,8 @@ class TextCompletionsRequestHandler(OpenAIRequestHandler):
         output_tokens = response.output_metrics.total_tokens or 0
         if not has_text and output_tokens <= 0:
             raise ValueError(
-                "[UNUSABLE_BACKEND_RESPONSE] backend resolved without a usable "
-                "terminal response payload"
+                "[UNUSABLE_BACKEND_RESPONSE] backend resolved with empty "
+                "response payload"
             )
 
     def extract_line_data(self, line: str) -> dict[str, Any] | None:
--- a/src/guidellm/backends/openai/request_handlers.py
+++ b/src/guidellm/backends/openai/request_handlers.py
@@ -512,7 +512,7 @@ class TextCompletionsRequestHandler(OpenAIRequestHandler):
         text = choice.get("text", "")
         input_metrics, output_metrics = self.extract_metrics(usage, text)
 
-        return GenerationResponse(
+        compiled = GenerationResponse(
             request_id=request.request_id,
             request_args=arguments.model_dump_json(),
             response_id=response.get("id"),  # use vLLM ID if available
@@ -520,6 +520,8 @@ class TextCompletionsRequestHandler(OpenAIRequestHandler):
             input_metrics=input_metrics,
             output_metrics=output_metrics,
         )
+        self._validate_compiled_response(compiled)
+        return compiled
 
     def add_streaming_line(self, line: str) -> int | None:
         """
@@ -562,7 +564,7 @@ class TextCompletionsRequestHandler(OpenAIRequestHandler):
         text = "".join(self.streaming_texts)
         input_metrics, output_metrics = self.extract_metrics(self.streaming_usage, text)
 
-        return GenerationResponse(
+        compiled = GenerationResponse(
             request_id=request.request_id,
             request_args=arguments.model_dump_json(),
             response_id=self.streaming_response_id,  # use vLLM ID if available
@@ -570,6 +572,18 @@ class TextCompletionsRequestHandler(OpenAIRequestHandler):
             input_metrics=input_metrics,
             output_metrics=output_metrics,
         )
+        self._validate_compiled_response(compiled)
+        return compiled
+
+    def _validate_compiled_response(self, response: GenerationResponse) -> None:
+        """Raise when endpoint produced a terminal payload with no usable output."""
+        has_text = bool(response.text and response.text.strip())
+        output_tokens = response.output_metrics.total_tokens or 0
+        if not has_text and output_tokens <= 0:
+            raise ValueError(
+                "[UNUSABLE_BACKEND_RESPONSE] backend resolved without a usable "
+                "terminal response payload"
+            )
 
     def extract_line_data(self, line: str) -> dict[str, Any] | None:
         """
@@ -1074,7 +1088,7 @@ class ChatCompletionsRequestHandler(TextCompletionsRequestHandler):
             output_metrics, len(tool_calls) if tool_calls else 0, text
         )
 
-        return GenerationResponse(
+        compiled = GenerationResponse(
             request_id=request.request_id,
             request_args=arguments.model_dump_json(),
             response_id=response.get("id"),  # use vLLM ID if available
@@ -1084,6 +1098,8 @@ class ChatCompletionsRequestHandler(TextCompletionsRequestHandler):
             input_metrics=input_metrics,
             output_metrics=output_metrics,
         )
+        self._validate_compiled_response(compiled)
+        return compiled
 
     def add_streaming_line(self, line: str) -> int | None:
         """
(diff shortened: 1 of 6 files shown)


sjmonson:
Hmm, yeah I am not a fan of depending on `GenerationResponse` in the worker. Can you maybe just check inside the `OpenAIRequestHandler.compile_streaming` and `OpenAIRequestHandler.compile_non_streaming` methods? It does not generalize as nicely but what qualifies as an unusable response may differ between backends/endpoints. If you throw an exception in those methods it should propagate up to the worker.

sjmonson (changes requested):
Can you rebase and get rid of the merge commits?

dbutenhof (commented):
I have really mixed feelings about this. I don't like the idea of hacking this into each request handler separately -- and you've only handled a subset of the http request handlers in this PR. Maybe that's OK, but it feels weird.

I like your original idea of triggering this through the worker, perhaps by delegating to a backend `validate_response` method (which could, if necessary, delegate to individual request handlers where more specialized knowledge is required). Your check, for example, is pretty much generic and might even be a default implementation on the base `Backend` class that everyone would get by default.

(The risk, of course, is that such a default might break some weird special handlers that don't have "normal" text/token outputs; which means a generalization would require extensive testing through all the validations.)

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -562,14 +564,26 @@ def compile_streaming(
...
+
+    def _validate_compiled_response(self, response: GenerationResponse) -> None:
+        """Raise when endpoint produced a terminal payload with no usable output."""
+        has_text = bool(response.text and response.text.strip())
+        output_tokens = response.output_metrics.total_tokens or 0
+        if not has_text and output_tokens <= 0:
+            raise ValueError(
+                "[UNUSABLE_BACKEND_RESPONSE] backend resolved without a usable "
+                "terminal response payload"
+            )
```
jaredoconnell: Instead of "without a usable terminal response payload", we should be more specific. Something like "empty response payload" would help.

tests/unit/backends/openai/test_request_handlers.py
```diff
@@ -1214,6 +1233,46 @@ def test_streaming(
...
 
+    @pytest.mark.regression
+    @pytest.mark.parametrize(
+        "response",
+        [
+            {"choices": [{"message": {"content": ""}}], "usage": {}},
+            {"choices": [], "usage": {}},
+        ],
+    )
+    def test_non_streaming_raises_for_unusable_terminal_payload(
```
SkiHatDuckie, line 1249: These tests (and the ones starting at 671) feel like duplicates of each other. But that's an issue for a different PR.

Tests are working for me, and I don't really have any additional complaints to add onto what the others have said.

SkiHatDuckie (approved):

tests/unit/backends/openai/test_request_handlers.py
ushaket: Thanks!

src/guidellm/backends/openai/request_handlers.py
ushaket: Done 🙏

ushaket:
@dbutenhof Thanks for the feedback, I agree the per-handler approach is a bit awkward, and it currently only covers text/chat.

I prototyped extending this to Responses (same text / tool_calls / tokens check) and Embeddings (raw data[].embedding check, since those responses never have text). Happy to include either or both in this PR if that’s useful.

How would you like to proceed?

Keep this PR scoped to text/chat and take broader coverage in a follow-up
Add Responses here
Add Responses + Embeddings here
Pull @sjmonson in and decide together (especially vs. a Backend.validate_response hook)

I’m fine with any of these, just want to align before I push more surface area into this PR.

dbutenhof:
> Keep this PR scoped to text/chat and take broader coverage in a follow-up Add Responses here Add Responses + Embeddings here Pull @sjmonson in and decide together (especially vs. a Backend.validate_response hook)
> 
> I’m fine with any of these, just want to align before I push more surface area into this PR.

Yeah, the risk is the "special cases" that don't have text, like embeddings, Geospatial pooling requests (I think), and maybe others.

While I like the idea of a meaningful `Backend` implementation that "works for most", the downside is a lot of awkward testing to be sure it doesn't break anything.

A safer alternative is probably a base implementation that does nothing, overridden by the various backends & request format handlers as necessary... I do prefer centralizing it from the worker loop, delegating through the backend to the request format... but as most of the implementation probably falls to the request formatter, maybe it doesn't matter much.

jaredoconnell:
> > Keep this PR scoped to text/chat and take broader coverage in a follow-up Add Responses here Add Responses + Embeddings here Pull @sjmonson in and decide together (especially vs. a Backend.validate_response hook)
> > I’m fine with any of these, just want to align before I push more surface area into this PR.
> 
> Yeah, the risk is the "special cases" that don't have text, like embeddings, Geospatial pooling requests (I think), and maybe others.
> 
> While I like the idea of a meaningful `Backend` implementation that "works for most", the downside is a lot of awkward testing to be sure it doesn't break anything.
> 
> A safer alternative is probably a base implementation that does nothing, overridden by the various backends & request format handlers as necessary... I do prefer centralizing it from the worker loop, delegating through the backend to the request format... but as most of the implementation probably falls to the request formatter, maybe it doesn't matter much.

Dave's proposal sounds good. It can have a default implementation that is permissive, and can be overridden whenever you need to validate this. The function could maybe be called "post_validation".

One thing that needs to be kept in mind is that for the chat completions endpoints and the responses API, no text is valid due to tool calls, and for other endpoints it's valid due to non-token content.

ushaket:
Thanks @dbutenhof, @jaredoconnell , pushed the refactor:

- post_validation added to the handler protocol with a no-op default, overridden in TextCompletionsRequestHandler and ResponsesRequestHandler
- Call site is in the backend resolve methods after compile, so new handlers get the no-op automatically
- Fixed the tool-call blind spot @jaredoconnell flagged: check now passes when any of text, tool_calls, or output_tokens is present
Pooling and embeddings keep the no-op
- Cancellation paths skip validation

Kept it at the backend resolve level rather than the worker loop since the implementation lives on the request handlers anyway as @dbutenhof pointed out

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -571,6 +595,16 @@ def compile_streaming(
...
+    def post_validation(self, response: GenerationResponse) -> None:
+        has_text = bool(response.text and response.text.strip())
+        has_tool_calls = bool(response.tool_calls)
+        output_tokens = response.output_metrics.total_tokens or 0
+        if not has_text and not has_tool_calls and output_tokens <= 0:
+            raise ValueError(
+                "[UNUSABLE_BACKEND_RESPONSE] backend resolved with empty "
+                "response payload"
+            )
+
```
jaredoconnell, line 619: It might be worth de-duplicating this and the other identical post_validation methods by making them both call the same helper function.

jaredoconnell (commented):
This seems like an appropriate design. One thing is that `OpenAIRequestHandler` is a protocol, so there is no default implementation. So no-op calls need to be added to `RealtimeTranscriptionWSRequestHandler` and `EmbeddingsRequestHandler`.

And I added a comment regarding de-duplicating the function you added.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -272,6 +284,18 @@ def compile_streaming(
         """
         ...
 
+    def post_validation(self, response: GenerationResponse) -> None:
```
dbutenhof, line 287: Duplicating this is ugly -- but I guess I also have to agree with not constructing a proper class hierarchy *in this PR* that would allow sharing it. That was why I was hoping to put it on `Backend`, because the backend classes *do* have a working hierarchy while the request handlers are a lot more chaotic. But this is probably safer since it minimizes the blast radius.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -571,6 +595,16 @@ def compile_streaming(
             output_metrics=output_metrics,
         )
 
+    def post_validation(self, response: GenerationResponse) -> None:
+        has_text = bool(response.text and response.text.strip())
```
dbutenhof: You wrote a big docstring for the null implementations, but nothing for the ones that do something? 😁

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -2147,6 +2191,9 @@ class PoolingRequestHandler(ChatCompletionsRequestHandler):
     pooling-specific request structure with nested data fields.
     """
 
+    def post_validation(self, response: GenerationResponse) -> None:  # noqa: ARG002
```
dbutenhof, line 2200: And ... completely different style here, without a docstring?

dbutenhof (changes requested):
I think there are a lot of compromises in this approach, but it's probably the lowest touch and safest. I'd like to see more consistent style in your validation methods, including docstrings; but aside from that, it's probably as good as it gets.

src/guidellm/backends/openai/request_handlers.py
ushaket: Done, extracted _validate_text_response helper, both handlers call it now.

src/guidellm/backends/openai/request_handlers.py
ushaket: Agreed, extracted a shared _validate_text_response helper to avoid the copy-paste.

src/guidellm/backends/openai/request_handlers.py
ushaket: added docstrings to all three overrides (Text, Responses, Pooling).

src/guidellm/backends/openai/request_handlers.py
ushaket: added a docstring explaining why pooling skips validation.

dbutenhof (approved):

jaredoconnell (approved):

ushaket:
@sjmonson can you re-review the requested changes?

sjmonson (approved):
My bad, forgot I blocked this.

sjmonson:
@mergifyio queue

PR #616, opened 2026-03-02
Add data parameter to benchmark command in README

S1ro1:
A very tiny PR, example in the README would fail with `ValueError: Data must be a non-empty list, got []` and was also missing a backslash.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,8 +84,8 @@ audio = [
     # Lowest version with full torchcodec support
     "datasets[audio]>=4.1.0",
     # Torchcodec needs specific torch version
-    "torch==2.9.*",
-    "torchcodec==0.8.*",
+    "torch==2.10.*",
+    "torchcodec==0.10.*",
 ]
 vision = [
     "datasets[vision]",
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -235,7 +235,8 @@ guidellm benchmark \
   --rate 16 \
   --warmup 0.1 \
   --cooldown 0.1 \
-  --max-errors 5
+  --max-errors 5 \
+  --data "prompt_tokens=1024,output_tokens=1024" \
```
dbutenhof: For the most part, the change looks good. However, in other places the documentation uses much smaller synthetic data parameters, like `prompt_tokens=256,output_tokens=128`, and I'd suggest following that precedent rather than leading readers to wonder whether there's a reason *this one* is different.

README.md
```diff
@@ -246,6 +247,7 @@ guidellm benchmark \
 - `--max-seconds`: Maximum duration in seconds for each benchmark before automatic termination
 - `--max-requests`: Maximum number of requests per benchmark before automatic termination
 - `--max-errors`: Maximum number of individual errors before stopping the benchmark entirely
+- `--data`: Data to use for benchmarking - random with 1024 input/output ratio in this case
```
dbutenhof: The "random" is true, but in other contexts the description refers to "synthetic data" with the randomness implied. Again, consistency is good.

jaredoconnell: I think the wording of "in this case" at the end is a little less clear. I would re-word this sentence.

Issue #602, opened 2026-02-19
Throughput statistics for min/max/stdev/percentiles do not give a good representation of the benchmark


DenDiv:
### Description
I am benchmarking Llama 3.1 8B on an NVIDIA H100 GPU. When running in concurrent mode, I observed highly inconsistent metrics: 

- A significant discrepancy between Mean and Median RPS and TPUT metrics.
- The Standard Deviation (std) for the RPS metric reached ~900, even though the actual maximum RPS was around 50. 

### Observations

- Console output: 
<img width="955" height="867" alt="Image" src="https://github.com/user-attachments/assets/06e18f4a-225f-48bc-82a2-c20d4618ba0f" />

- Data file:  [concurrenct50.json](https://github.com/user-attachments/files/25416006/concurrenct50.json)

### Debug
I suppose the problem is caused by a very low [threshold](https://github.com/vllm-project/guidellm/blob/4786a12d17f8c9c9a6de87e429c1d926085a0ceb/src/guidellm/schemas/statistics.py#L318) for merging. In concurrent mode, many requests can finish almost simultaneously, which leads to _durations_ being as low as 1e-4 in [this line](https://github.com/vllm-project/guidellm/blob/4786a12d17f8c9c9a6de87e429c1d926085a0ceb/src/guidellm/schemas/statistics.py#L380), resulting in an inflated rate. I set the threshold to 1.0 and received more stable results.

### Expected Behavior

Stable quantile metrics and low variance 

### Steps to Reproduce

Below code for reproducing problem: start model in sglang and benchmark scenario

model start:
```
python3 -m sglang.launch_server \
  --model-path /models/Llama/Llama3.1-8B-Instruct/ \
  --tp=1 \
  --dp=1 \
  --enable-metrics \
  --disable-radix-cache \
```

scenario.json:
```
{
  "profile": "concurrent",
  "rate": 50,
  "max_seconds": 10,
  "target": "http://localhost:30000",
  "data": "prompt_tokens=128,output_tokens=128",
  "processor": "/models/Llama/Llama3.1-8B-Instruct/",
}
```

### Operating System

Ubunty 22.04

### Python Version

Python 3.12.12

### GuideLLM Version

guidellm version: 0.6.0.dev75

### Installation Method

pip install guidellm

### Error Messages or Stack Traces

```shell

```

sjmonson:
[...]

PR #617, opened 2026-03-02
Drop median from throughput metrics in console

sjmonson:
## Summary

Only show mean for request/token based concurrences. See #602 for reasoning.

## Details

Result of change:

```
ℹ Server Throughput Statistics (All Requests)
|============|=======|======|=========|==============|===============|==============|
| Benchmark  | Requests             ||| Input Tokens | Output Tokens | Total Tokens |
| Strategy   | Concurrency || Per Sec | Per Sec      | Per Sec       | Per Sec      |
|            | Mdn   | Mean | Mean                                               ||||
|------------|-------|------|---------|--------------|---------------|--------------|
| concurrent | 25.0  | 25.0 | 2.3     | 2662.3       | 2319.4        | 4775.9       |
|============|=======|======|=========|==============|===============|==============|



✔ Benchmarking complete, generated 1 benchmark(s)
…   json    : /root/guidellm/benchmarks.json
…   csv     : /root/guidellm/benchmarks.csv
…   html    : /root/guidellm/benchmarks.html
```

## Related Issues

- Related to #602

---

## Use of AI

### DIFF
--- a/src/guidellm/benchmark/outputs/console.py
+++ b/src/guidellm/benchmark/outputs/console.py
@@ -467,18 +467,18 @@ class GenerativeBenchmarkerConsole(GenerativeBenchmarkerOutput):
                 type_="text",
             )
             columns.add_stats(
+                benchmark.metrics.request_concurrency,
+                status="total",
+                group="Requests",
+                name="Concurrency",
+                types=("median", "mean"),
+            )
+            columns.add_stats(
                 benchmark.metrics.requests_per_second,
                 status="total",
                 group="Requests",
                 name="Per Sec",
                 types=("mean",),
-            )
-            columns.add_stats(
-                benchmark.metrics.request_concurrency,
-                status="total",
-                group="Requests",
-                name="Concurrency",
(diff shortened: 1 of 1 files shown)


src/guidellm/benchmark/outputs/console.py
```diff
@@ -467,39 +467,39 @@ def print_server_throughput_table(self, report: GenerativeBenchmarksReport):
                 type_="text",
             )
             columns.add_stats(
-                benchmark.metrics.requests_per_second,
+                benchmark.metrics.request_concurrency,
```
dbutenhof, line 470: I kinda wonder if changing the order of columns in the console table is going to confuse anyone... although I do like having all the "Per Sec" columns together (which I would guess to be the motivation here...)

sjmonson: The tables have shifted a bit nearly every release so I don't think anyone is too dependent on the order. And yeah that was my motivation.

Issue #588, opened 2026-02-04
--max-requests with --rate-type=sweep causes first stage to run all requests synchronously, resulting in extremely long runtimes


cemigo114:
### Bug Description

When using --rate-type=sweep combined with --max-requests=N, the first synchronous stage of the sweep runs ALL N requests sequentially before moving to parallel stages. This results in unexpectedly long benchmark runtimes, especially with larger request counts.
Observed Behavior
A benchmark configured with --max-requests=1000 --rate-type=sweep took ~7 hours to complete on H100 GPUs, when users expected it to complete in minutes.
Root Cause
The sweep profile's first stage runs requests synchronously (one at a time). The --max-requests parameter applies to each stage, causing all 1000 requests to be processed sequentially in stage 1 alone before any parallel stages begin.

### Expected Behavior

Either:
--max-requests should be the total requests across all sweep stages, not per-stage
The synchronous stage should have a separate, smaller limit
Documentation should clearly warn against using high --max-requests values with sweep mode

### Steps to Reproduce

apiVersion: batch/v1
kind: Job
metadata:
  name: guidellm-benchmark
spec:
  template:
    spec:
      containers:
      - name: guidellm
        image: quay.io/jhurlocker/guidellm:latest
        args:
          - benchmark
          - --target=http://my-llm-service:8080/v1
          - --model=granite-33-8b-instruct
          - --rate-type=sweep
          - --max-requests=1000    # <-- This causes the issue
          - --data=/mnt/prompts/prefix-prompts.csv
          - --output-path=/results/output.json

### Operating System

OpenShift (RHOAI 3.0)

### Python Version

OpenShift (RHOAI 3.0)

### GuideLLM Version

latest

### Installation Method

pip install guidellm

### Error Messages or Stack Traces

```shell

```

### Additional Context

Suggested Improvements
Documentation: Add a warning to the README/docs that --max-requests with sweep mode runs synchronously in the first stage
CLI Warning: Emit a warning when --max-requests > 100 is used with --rate-type=sweep
Design Change: Consider making the synchronous stage use a smaller, fixed request count (e.g., 50-100) regardless of --max-requests
Environment
GuideLLM version: latest
Hardware: NVIDIA H100 GPUs
Platform: OpenShift (RHOAI 3.0)

dbutenhof:
[...]

PR #619, opened 2026-03-03
Add detail in benchmark profile documentation

dbutenhof:
## Summary

Attempt to clarify the interactions of `--rate` and the count and time-based constraints with the various benchmarking profiles; especially sweep, which presents effects that aren't obvious at first glance.

## Details

The simple list of profiles is broken up into sub-sections with more detail on behavior and interactions to, hopefully, aid in understanding.

## Test Plan

N/A

## Related Issues

This is a first attempt to mitigate the implications of #588 with a clear warning that the global constraints are implemented separately for each strategy, and that sweep especially includes multiple strategies. We expect that future work will allow specifying separate constraints for each strategy.

---

## Use of AI

### DIFF
--- a/docs/getting-started/benchmark.md
+++ b/docs/getting-started/benchmark.md
@@ -161,7 +161,6 @@ For synthetic data, parameters for random data generation are passed as key=valu
 
 - `prompt_tokens`: Average number of tokens for prompts
 - `output_tokens`: Average number of tokens for outputs
-- `samples`: Number of samples to generate (default: 1000)
 
 For example, to generate 1000 samples with a prompt length of 100 tokens and an output length of 50 tokens:
 


docs/getting-started/benchmark.md
```diff
@@ -40,27 +40,111 @@ GuideLLM offers a wide range of configuration options to customize your benchmar
...
+```
+
+| Strategy parameter | Description                                                                                       | Example                     |
+| ------------------ | ------------------------------------------------------------------------------------------------- | --------------------------- |
+| `--rate`           | Maximum asynchronous requests to run. May be a list of values to run successive constant streams. | `--rate 10`, `--rate 16,32` |
+| `--random-seed`    | Not used.                                                                                         |                             |
+
+#### Poisson Profile
+
+Sends asynchronous requests at varying rates using a Poisson distribution around the specified target rate(s).
```
jaredoconnell: It may be useful to add a statement about why someone would want this profile. I would say that the advantage of this profile is it can more accurately mimic a real-world scenario.

docs/getting-started/benchmark.md
```diff
@@ -40,27 +40,111 @@ GuideLLM offers a wide range of configuration options to customize your benchmar
...
+```
+
+| Strategy parameter | Description  | Example |
+| ------------------ | ------------ | ------- |
+| `--rate`           | Not allowed. |         |
+| `--random-seed`    | Not used.    |         |
+
+#### Throughput Profile
+
+Tests maximum throughput by running requests in parallel.
```
jaredoconnell: I'm not sure on the perfect wording, but I would make it clear that this is designed to overwhelm the servers with more parallel requests than they can easily handle, which will increase latency readings, but help ensure it finds the maximum throughput of the server.

docs/getting-started/benchmark.md
```diff
@@ -40,27 +40,111 @@ GuideLLM offers a wide range of configuration options to customize your benchmar
...
+
+```bash
+guidellm benchmark --profile throughput
+```
+
+| Strategy parameter | Description                         | Example     |
+| ------------------ | ----------------------------------- | ----------- |
+| `--rate`           | Maximum concurrent requests to run. | `--rate 10` |
```
jaredoconnell: Rampup can also apply to this one.

docs/getting-started/benchmark.md
```diff
@@ -40,27 +40,111 @@ GuideLLM offers a wide range of configuration options to customize your benchmar
...
+
+(The profile name `async` is an alias for the `constant` profile.)
+
+```bash
+guidellm benchmark --profile constant
+```
+
+| Strategy parameter | Description                                                                                       | Example                     |
+| ------------------ | ------------------------------------------------------------------------------------------------- | --------------------------- |
+| `--rate`           | Maximum asynchronous requests to run. May be a list of values to run successive constant streams. | `--rate 10`, `--rate 16,32` |
```
jaredoconnell: Rampup can also apply to this one, creating a linear buildup to the specified rate.

docs/getting-started/benchmark.md
```diff
@@ -40,27 +40,111 @@ GuideLLM offers a wide range of configuration options to customize your benchmar
...
+
+Runs a fixed number of parallel request streams.
+
+```bash
+guidellm benchmark --profile concurrent
+```
+
+| Strategy parameter | Description                                                                                                          | Example                     |
+| ------------------ | -------------------------------------------------------------------------------------------------------------------- | --------------------------- |
+| `--rate`           | Number of concurrent streams to run. May be a single value or a list of values to run successive concurrent streams. | `--rate 10`, `--rate 16,32` |
```
jaredoconnell: Rampup can also apply to this one.

docs/getting-started/benchmark.md
dbutenhof: That's a good point. I went back and forth a lot on what to include in these tables. I decided that it makes sense to include the global options (like `--rate`) which have distinct behaviors for each profile -- I missed `--rampup`, but since it's not used for all profiles that makes sense. Any others you can think of?

jaredoconnell: I checked and I think that's it.

docs/getting-started/benchmark.md
```diff
@@ -40,27 +40,117 @@ GuideLLM offers a wide range of configuration options to customize your benchmar
...
+Attempts to discover the server's maximum throughput by continually making requests in parallel.
+
+```bash
+guidellm benchmark --profile throughput
+```
+
+| Strategy parameter | Description                                              | Example       |
+| ------------------ | -------------------------------------------------------- | ------------- |
+| `--rate`           | Number of concurrent request streams.                    | `--rate 10`   |
+| `--random-seed`    | Not used.                                                |               |
```
sjmonson: Main concern is that `--random-seed` *could* be used by synthetic data even in synchronous, throughput, concurrent, etc. Putting "not used" here seems like it will lead to user confusion. Perhaps we should have a section at the top that explains random-seed and a few of its uses in the codebase rather than this.

Something like (don't need to use this exact wording):

> `--random-seed` is used for any operation in GuideLLM that involves randomness such as synthetic data generation or poisson distribution calculations. By default it is a fixed value; meaning that rerunning GuideLLM with the exact same arguments should produce the same results.

jaredoconnell: In that case, I would only include it in the table of strategy parameters of the profiles that use it.

dbutenhof: Yeah, I thought about that at the time and was hoping the "profile" context would be clear enough. I could say "not used by the scheduling profile or strategy" -- but I suppose you're correct that might not really help a lot.

I did write the "Strategy Constraints" note above the profile subsections, and I could add a "note about random seed" as well.

I think I like Sam's suggestion ... which also suggests pulling it up out of "Benchmark profiles", maybe into a separate H3 section. I'll give that a try...

docs/getting-started/benchmark.md
```diff
@@ -34,49 +34,148 @@ During the benchmark, you'll see a progress display similar to this:
...
 
-### Data Options
+### Synthetic Data Options
 
-For synthetic data, some key options include, among others:
+For synthetic data, parameters for random data generation are passed as key=value pairs to the `--data` parameter. Some key options include:
 
 - `prompt_tokens`: Average number of tokens for prompts
 - `output_tokens`: Average number of tokens for outputs
 - `samples`: Number of samples to generate (default: 1000)
```
sjmonson: Samples no longer exists, can you remove this?

```suggestion
```

dbutenhof:
> Looks good except you need to fix the signoff.

Ouch ... I keep just pushing "Commit" in Cursor, which, despite following all the instructions I'd found, still fails to sign by default. But on a whim I searched through the Cursor settings and found a checkbox ... hopefully it'll work next time. (And at least I aspire to remember to confirm with `git log` before pushing. 😆)

Issue #597, opened 2026-02-12
Feature Request: Add Mooncake Trace Data Support


cemigo114:
### Problem Statement

GuideLLM currently lacks support for the Mooncake trace format, which provides real datasets with synthetic tokens and time-based request rates. This gap prevents users from performing proper customer workload simulation using Mooncake traces.

Mooncake traces are becoming an industry standard for KV cache testing and LLM inference benchmarking. Without native support for this trace format, GuideLLM users cannot:

- Replay production-representative workloads that use Mooncake's real dataset with synthetic token distributions
- Leverage Mooncake's time-based request rate patterns for realistic load generation
- Benchmark and demonstrate performance against industry-standard KV cache testing methodologies

This gap directly affects the ability to evaluate and compare LLM deployment performance against widely adopted industry benchmarks.

### Proposed Solution

Add native Mooncake trace format support to GuideLLM's data ingestion pipeline. This would involve:

1. **Trace Parser**: Implement a Mooncake trace parser that can read and interpret Mooncake's trace file format, extracting request timestamps, token counts (input/output), and request metadata.

2. **Data Source Integration**: Add a new `--data` source type (e.g., `--data mooncake:<path_to_trace>`) that allows users to load Mooncake trace files directly as workload definitions.

3. **Time-Based Rate Replay**: Support Mooncake's time-based request rate patterns so that GuideLLM can replay requests at the exact inter-arrival times specified in the trace, enabling faithful production workload reproduction.

4. **Synthetic Token Mapping**: Map Mooncake's synthetic token specifications to GuideLLM's internal token representation, ensuring that input/output token distributions from the trace are accurately reflected in the benchmark requests.

5. **Documentation**: Add documentation and examples showing how to use Mooncake traces with GuideLLM for KV cache performance testing.

### Alternatives Considered

**Manual trace conversion**: Users could manually convert Mooncake trace files into GuideLLM's existing supported formats (e.g., custom JSON or HuggingFace datasets). However, this is error-prone, loses the time-based request rate information that is central to Mooncake's value, and creates a significant barrier for users who want to quickly benchmark with industry-standard traces.

**External preprocessing scripts**: A standalone script could preprocess Mooncake traces into a compatible format. This adds toolchain complexity and maintenance burden, and still loses fidelity in the time-based scheduling that Mooncake traces provide.

[...]

PR #620, opened 2026-03-04
[FEAT] Add replay from trace strategy

VincentG1234:
### Summary

- Add a new `replay` benchmarking strategy that reproduces real-world request patterns from trace log files (.jsonl)
- Enable time-based request rate replay with precise timestamp scheduling
- Support synthetic prompt generation that matches token counts from trace files
- use `max_requests` and `max_seconds` cli options to limit the number of requests processed from a trace

### Motivation

This change addresses issue #597 by enabling users to benchmark their vLLM servers using real production traces. Instead of synthetic load patterns, users can now replay exact request arrival times and token distributions from their actual workloads for more realistic performance testing.

### Changes

- Add `TraceReplayStrategy` scheduler strategy for timestamp-based request dispatching
- Add `ReplayProfile` class for configuring trace-based benchmarking parameters
- Add `TraceSyntheticDatasetDeserializer` to generate prompts matching trace input/output lengths
- Add `TraceReader` utility for reading .jsonl trace files with timestamp, input_length, output_length fields
- Update `Entrypoint` to handle replay profile and dataset configuration
- use `max_requests` and `max_seconds` truncation support to limit trace replay length

### Testing

- `pytest tests/unit/scheduler/test_trace_replay.py` (pass)
- `pytest tests/unit/benchmark/test_replay_profile.py` (pass)
- `pytest tests/unit/data/deserializers/test_trace_synthetic.py` (pass)
- Added tests: scheduling accuracy, boundary conditions, malformed trace handling, empty trace cases, max_requests truncation

- test it in practice quickly with [NB COLAB](https://colab.research.google.com/drive/1hOY9Kg5BVYz4BZzJLrcUKWiMDR2_lU7V?usp=sharing)

### Next Steps (this PR)

1. Apply reviewer feedback
2. Add E2E tests verifying end-to-end trace replay flow ✅ 
3. Add integrations tests (if needed)
4. Add CLI usage examples in PR description and docs ✅ 

### Out of Scope (future PRs or not)

- Mooncake trace format support (token-level traces)
- Helper utilities for timestamp format conversions (Unix epoch, ISO8601, relative timestamps)
- Support for request payload traces (not just token counts)
- Trace file validation and schema verification tools
- Performance optimizations for large trace files (streaming, chunked processing)
- Metrics export formatted for trace analysis comparison
- Support for trace file compression formats (.gz, .bz2)

## Use of AI


src/guidellm/data/trace_io.py
sjmonson, line 1: Since this file is used by data, scheduler, and benchmark move it to `utils`.

src/guidellm/scheduler/__init__.py
```diff
@@ -50,6 +50,8 @@
     StrategyType,
     SynchronousStrategy,
     ThroughputStrategy,
+    TraceReplayStrategy,
+    load_relative_timestamps,
```
sjmonson, line 54: Not a method of this submodule.

```suggestion
```

src/guidellm/scheduler/__init__.py
```diff
@@ -90,7 +92,9 @@
     "StrategyType",
     "SynchronousStrategy",
     "ThroughputStrategy",
+    "TraceReplayStrategy",
     "UnserializableConstraintInitializer",
     "WorkerProcess",
     "WorkerProcessGroup",
+    "load_relative_timestamps",
```
sjmonson, line 99: Not a method of this submodule.

```suggestion
```

src/guidellm/scheduler/strategies.py
```diff
@@ -37,11 +38,13 @@
     "StrategyType",
     "SynchronousStrategy",
     "ThroughputStrategy",
+    "TraceReplayStrategy",
+    "load_relative_timestamps",
```
sjmonson, line 42: Not a method of this submodule.

```suggestion
```

src/guidellm/scheduler/strategies.py
```diff
@@ -25,6 +25,7 @@
 
 from pydantic import Field, NonNegativeFloat, NonNegativeInt, PositiveInt, PrivateAttr
 
+from guidellm.data.trace_io import load_relative_timestamps
```
sjmonson, line 28: See comment on `trace_io.py`

```suggestion
from guidellm.utils.trace_io import load_relative_timestamps
```

src/guidellm/benchmark/entrypoints.py
```diff
@@ -314,6 +316,17 @@ async def resolve_request_loader(
...
+    if max_requests is not None and data:
+        if max_requests < 1:
+            raise ValueError(
+                "max_requests must be >= 1 when set for data truncation, "
+                f"got {max_requests}"
+            )
+        data_args = list(data_args) if data_args else [{} for _ in data]
+        if len(data_args) >= 1:
+            data_args[0] = {**data_args[0], "max_rows": max_requests}
+
```
sjmonson, line 329: Drop this, `max_requests` is a constraint on the number of requests that are allowed to complete. To limit the data source use `--data-samples`.


```suggestion
```

VincentG1234:
Yeah that’s true, most frameworks won’t produce this exact JSONL directly.

That’s kind of intentional. The idea here is to define a minimal, framework-agnostic canonical replay format, not something tied to a specific tracing stack.

In practice, the required fields already exist almost everywhere (timestamp, input token count, output token count), just under slightly different names, so a small mapping step is usually enough.

I agree it’s not the best UX on its own, but it felt like the right minimal base for the feature. Then we can iterate on top of it with helpers / converters for common sources like LiteLLM or Langfuse. And we can extend it later (e.g. optional prompt field, multiple timestamp formats, richer metadata) without breaking the core idea.

But happy to adjust the direction if maintainers prefer something more opinionated or integrated from the start.

src/guidellm/benchmark/entrypoints.py
```diff
@@ -496,24 +512,60 @@ async def benchmark_generative_text(
...
+        "data_sampler": args.data_sampler,
+        "data_num_workers": args.data_num_workers,
+        "random_seed": args.random_seed,
+        "console": console,
+    }
+
+    # For replay profile: resolve profile first to apply max_seconds filtering,
+    # then use the filtered count for the data loader. This ensures the data
+    # loader and scheduler both work with the same filtered request count.
+    if args.profile == "replay":
```
sjmonson, line 552: Unless I am missing something, this conditional should be unnecessary. There is no reason to do loader then profile other than its the way things were done before.

src/guidellm/benchmark/entrypoints.py
```diff
@@ -496,24 +512,60 @@ async def benchmark_generative_text(
...
+        profile = await resolve_profile(**profile_kwargs, data=args.data)  # type: ignore[arg-type]
+        effective_max_requests = (
+            profile.constraints.get("max_requests")
+            if profile.constraints
+            else args.max_requests
+        )
```
sjmonson, line 558: Not a huge fan of this and I also think its unnecessary. The profile can trigger a benchmark end based on the number of requests. Its fine if the request loader reads too many requests ahead.

src/guidellm/data/trace_io.py
```diff
@@ -0,0 +1,92 @@
...
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from typing import Any
+
+__all__ = ["load_relative_timestamps", "load_trace_rows"]
+
+
+def load_trace_rows(
```
sjmonson, line 18: Replace this function with a call to `datasets.load_dataset`. Can basically be a copy of `JSONFileDatasetDeserializer`. I.e.

```python
return load_dataset("json", data_files=str(path), **data_kwargs)
```

src/guidellm/data/trace_io.py
```diff
@@ -0,0 +1,92 @@
...
+    :param path: Path to the trace file.
+    :param timestamp_column: Name of the column/field containing the timestamp.
+    :return: List of relative timestamps in seconds (first is 0.0, always sorted).
+    :raises ValueError: If the trace file is empty or has no valid rows.
+    """
+    raw = load_trace_rows(
+        path,
+        required_columns=[timestamp_column],
+    )
+    timestamps = sorted([float(row[timestamp_column]) for row in raw])
```
sjmonson, line 88: if using `datasets` can do `raw.sort(timestamp_column)`.

sjmonson (changes requested):
Sorry for the silence on this. There are a few things with this PR that break other use-cases. I am still working on a more complete review but here are a few low hanging problems.

VincentG1234:
Thanks a lot for the detailed review, I really appreciate your time.

I’m fully aligned with your feedback, especially on the replay handling in the entrypoint, which is a key part of the PR. I agree that introducing a special case here is not ideal and should be avoided.

I’ll refactor this to make it cleaner and better aligned with the existing design.

VincentG1234:
Hi @sjmonson,

Thanks again for the feedback.

I’ve completed the refactor and addressed the main review points.

Key updates:
- replay now uses the standard request loader flow (removed special-case path)
- `data_samples` handles dataset truncation, `max_requests` stays runtime-only
- shared trace loading moved to `guidellm.utils.trace_io`
- simplified replay constraints
- tests/docs cleaned up accordingly

Would appreciate another look when you have time.

Optional: I also put together a small Colab notebook to try the feature quickly if useful:
https://colab.research.google.com/drive/1hOY9Kg5BVYz4BZzJLrcUKWiMDR2_lU7V?usp=sharing

docs/getting-started/benchmark.md
```diff
@@ -187,6 +188,32 @@ guidellm benchmark \
 
 You can customize synthetic data generation with additional parameters such as standard deviation, minimum, and maximum values. See the [Datasets Synthetic data documentation](../guides/datasets.md#synthetic-data) for more details.
 
+### Trace Replay Benchmarking (beta)
+
+For realistic load testing, replay trace events using each row's timestamp and token lengths. Trace files must be JSONL and are loaded with the `trace_synthetic` data type. By default, each row uses `timestamp`, `input_length`, and `output_length` fields:
+
+```json
+{"timestamp": 0, "input_length": 256, "output_length": 128}
```
dbutenhof: There's some inconsistencies in the documentation of timestamps. You're expecting absolute timestamps, which are converted during import to a sequence of relative timestamps; and the description of `--rate` behavior says you're "multiply[ing] timestamps" by the rate factor when you're actually scaling the intervals.

The use of small values here obscures the semantics by using a base time of 0, which legitimately supplies a series of intervals: but your code is still subtracting timestamp[0] from each value, and I don't think this is the best way to describe the behavior.

docs/guides/datasets.md
```diff
@@ -116,22 +120,51 @@ GuideLLM supports various file formats for datasets, including text, CSV, JSON,
...
+
+  Use `--data-args '{"type_": "trace_synthetic"}'` to enable trace loading. If your trace uses different column names, configure them with `timestamp_column`, `prompt_tokens_column`, and `output_tokens_column`:
+
+  ```bash
+  guidellm benchmark \
+      --target "http://localhost:8000" \
+      --profile replay \
+      --rate 1.0 \
+      --data "path/to/trace.jsonl" \
+      --data-args '{"type_": "trace_synthetic", "timestamp_column": "ts", "prompt_tokens_column": "input_tokens", "output_tokens_column": "output_tokens"}'
```
dbutenhof: I'm not crazy about the `"type_": "trace_synthetic"` here, although there isn't a great alternative right now since there's no good way for the current deserializer mechanism to distinguish that the JSONL file represents a trace. (Note we're working on a CLI revamp that will make the data "type" unambiguous as part of `--data`, something like `--data trace_synthetic <file>`, which will take care of that problem, so luckily we don't have to worry about that in this PR.)

Since you've shown a sample JSONL above, I think it's a bit odd that you're showing column mappings here that don't match the sample. You might consider renaming your sample JSON timestamp keys match your specification here so the spec is compatible with your example.

dbutenhof (commented):
A couple of comments mostly around documentation consistency...

docs/guides/datasets.md
VincentG1234: Thanks for the review.

I’ve addressed those points and updated the docs for clarity and consistency. Let me know if anything still needs adjustment.

docs/guides/datasets.md
```diff
@@ -116,22 +120,53 @@ GuideLLM supports various file formats for datasets, including text, CSV, JSON,
...
   ```
+
 - **JSON Lines files (`.jsonl`)**: Where each line is a separate JSON object. The objects should include `prompt` or other common names for the prompt which will be used as the prompt column. Additional fields can be included based on the previously mentioned aliases for the `--data-args` argument.
+
   ```json
   {"prompt": "Hello, how are you?", "output_tokens_count": 5, "additional_column": "foo", "additional_column2": "bar"}
   {"prompt": "What is your name?", "output_tokens_count": 3, "additional_column": "baz", "additional_column2": "qux"}
   ```
+
+- **Trace files (`.jsonl` with `trace_synthetic` type)**: Specialized JSONL files for replay benchmarking with `timestamp`, `input_length`, and `output_length` fields. Used with `--profile replay` to replay trace events using each row's timestamp and token lengths. Timestamps may be absolute or monotonic values; GuideLLM sorts them and converts them to offsets from the first event before scheduling. See [Trace Replay Benchmarking](../getting-started/benchmark.md#trace-replay-benchmarking).
```
dbutenhof: "Monotonic" is an odd choice considering that you explicitly sort the values. Also, the timestamps always need to be "absolute" (since you're sorting them!); I guess that by "absolute" you mean "offset from the UNIX epoch", which is just a particular absolute zero point. Your algorithm will work with *any* zero point, but your sequence is always interpreted as "absolute" offsets from that arbitrary zero point.

(And do you need to sort? That is, are real traces recorded out of order? I can certainly see that happening, so I'm not necessarily objecting: but, e.g., if Mooncake & other common trace standards require that traces are recorded in order, we should skip that since sorting a long list of traces can be an expensive startup cost.)

docs/guides/datasets.md
```diff
@@ -141,27 +141,29 @@ GuideLLM supports various file formats for datasets, including text, CSV, JSON,
...
       --target "http://localhost:8000" \
       --profile replay \
       --rate 1.0 \
       --data "path/to/trace.jsonl" \
-      --data-args '{"type_": "trace_synthetic", "timestamp_column": "ts", "prompt_tokens_column": "input_tokens", "output_tokens_column": "output_tokens"}'
+      --data-args '{"type_": "trace_synthetic"}'
   ```
 
-  For replay, `--rate` is a time scale rather than requests per second. Use `--data-samples` to limit how many trace rows are loaded and replayed. Use `--max-requests` only as a runtime completion constraint; it does not limit the trace rows loaded from the file.
+  If your trace uses different column names, configure them with `timestamp_column`, `prompt_tokens_column`, and `output_tokens_column`.
```
dbutenhof: I like the cli sample change to conform to the key names in your sample JSONL, *but* I did like that you had an example of how to specify the custom column headers, and this paragraph no longer connects those mapping keywords into the CLI. I'm not sure that'll be obvious to all users. A second sample `guidellm benchmark` command showing column mapping might help.

```console
guidellm benchmark \
    --profile replay \
    --rate 1.0 \
    --data replay.jsonl \
    --data-args type_=trace_synthetic,timestamp_column=timestamp
```

Also, note that GuideLLM supports the above slimmer `k=v` style in place of the full inline serialized JSON here, and it's a bit easier to read and generally preferred in documentation:

```console
✔ Request loader initialized with inf unique requests
  {'data': "['replay.jsonl']", 'data_args': "[{'type_': 'trace_synthetic', 'timestamp_column': 'timestamp'}]", 'data_samples': -1, 'preprocessors': ['GenerativeColumnMapper', 'MediaEncoder'], 'finalizer':                           
  'GenerativeRequestFinalizer', 'collator': 'GenerativeRequestCollator', 'sampler': 'None', 'num_workers': 1, 'random_seed': 42}                                                                                                       
```

docs/guides/datasets.md
VincentG1234: Good point, “monotonic” was not the right wording there. I changed it

On sorting: yes, I think it is still useful. In many traces the logs are ordered by completion time rather than arrival time (e.g in litellm on my side), and in distributed setups records can also be emitted or collected slightly out of timestamp order. Moreover, these traces are usually already close to sorted, so this should be near the best case for sorting in practice, while preserving correct arrival ordering for replay.

src/guidellm/utils/trace_io.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def load_relative_timestamps(
+    path: Path | str,
+    timestamp_column: str = "timestamp",
+) -> list[float]:
+    """
+    Load timestamps from a trace file and return times relative to the first event.
+
+    Trace file must be JSONL (one JSON object per line). Timestamps are sorted
```
dbutenhof: Generally, a docstring should describe what *this* function does; the sorting is not done here, and that description belongs in `load_trace_rows`, above. Although *this* function always includes `load_trace_rows`, it's also called directly, so the behavior shouldn't be described here.

Which actually raises an interesting issue, though I don't think it's feasible to resolve in this PR so this is just commentary. It's unfortunate that `load_trace_rows` is called multiple times, re-reading the trace JSONL file, converting and sorting the timestamps -- a potentially expensive redundancy.

In the longer run, we probably need to make the constructed dataset available to the profile with an interface to access scheduling information.

VincentG1234: Just pushed the doc updates. Let me know if anything else is needed on this.

sjmonson:
I tried running a test with [this dataset](https://huggingface.co/datasets/valeriol29/mooncake-traces/viewer/mooncake) and got a deadlock at the start of the benchmark. Here is a jsonl version for testing: [data.jsonl.gz](https://github.com/user-attachments/files/27366205/data.jsonl.gz).

VincentG1234:
> I tried running a test with [this dataset](https://huggingface.co/datasets/valeriol29/mooncake-traces/viewer/mooncake) and got a deadlock at the start of the benchmark. Here is a jsonl version for testing: [data.jsonl.gz](https://github.com/user-attachments/files/27366205/data.jsonl.gz).

Thanks for testing this and for sharing the dataset.

I can reproduce the issue on my side as well, including with a smaller subset of the trace. I’ll investigate it and work on a fix as soon as possible.

At first glance, this seems related to how replay handles large/bursty traces and high-token-count requests.

I’ll follow up once I have a clearer diagnosis and a fix.

VincentG1234:
Hi @sjmonson 
This commit fixes two issues found while replaying the shared JSONL trace.

First, synthetic prompt generation now builds one reusable base prompt and creates each request prompt by adding a unique prefix before slicing to the requested input length. This keeps prompts cache-resistant while avoiding the previous expensive per-request generation path.

Second, trace replay is temporarily limited to one process. With multiple processes, there is currently a race condition where some scheduled requests can be consumed out of order or never sent, which leaves the benchmark waiting forever. Capping replay to one process is a workaround, but it makes the benchmark complete reliably; the only expected limitation is for extreme traces where one scheduling process may become a bottleneck.

I tested this with the shared JSONL trace: the benchmark no longer hangs and starts correctly after roughly 50 seconds. On a representative subset, the previous prompt generation path was at least 10x slower...

sjmonson:
> First, synthetic prompt generation now builds one reusable base prompt and creates each request prompt by adding a unique prefix before slicing to the requested input length. This keeps prompts cache-resistant while avoiding the previous expensive per-request generation path.

Yeah... I don't love this idea, but fine for now. I see the dataset generation as temporary/fallback anyways since what we want longer term is to base the tokens off of the mooncake token ids.

> Second, trace replay is temporarily limited to one process. With multiple processes, there is currently a race condition where some scheduled requests can be consumed out of order or never sent, which leaves the benchmark waiting forever. Capping replay to one process is a workaround, but it makes the benchmark complete reliably; the only expected limitation is for extreme traces where one scheduling process may become a bottleneck.

Also works for now. Fixing this requires a way for the dataset to inform on request scheduling so we'll scope something out.

VincentG1234:
For multiprocessing, I think I may have found a fairly minimal approach that avoids the replay deadlock while keeping the implementation relatively clean, but I agree it’s probably better scoped for a follow-up PR.

For the Mooncake token-id direction, unless I miss something, I think the current prefix invalidation approach can remain compatible with a more structure-aware strategy later on.

Roughly, unrelated prompts would still receive different invalidating prefixes, while prompts sharing a common prefix could intentionally reuse the same initial invalidating block and only diverge later with unique suffix markers.

Example:

base prompt:
[A, B, C, D, E]

prompt 1:
[_x_, A, B, C]

prompt 2 (totally unrelated to prompt 1):
[_y_, A, B, C, D]

prompt 3 (shares the same prefix structure as prompt 2 up to block D):
[_y_, A, B, C, _z_, D, E]

This keeps prompts cache-resistant globally while still allowing controlled shared-prefix behavior between related requests. The same idea could likely be extended recursively for deeper shared-prefix structures.

sjmonson:
augment review

src/guidellm/benchmark/profiles.py
```diff
@@ -328,6 +332,104 @@ def next_strategy(
...
+        rate_type: str,
+        rate: list[float] | None,
+        random_seed: int,
+        **kwargs: Any,
+    ) -> dict[str, Any]:
+        _ = (rate_type, random_seed)  # unused
+        data = kwargs.get("data")
+        if not data or not data[0]:
+            raise ValueError("Replay profile requires data (path to trace file)")
+        path = Path(data[0]) if isinstance(data[0], str) else data[0]
```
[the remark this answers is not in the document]
sjmonson: Good idea. Add another `ValueError`.

src/guidellm/benchmark/profiles.py
```diff
@@ -328,6 +332,104 @@ def next_strategy(
...
+        # replay profile and trace_synthetic deserializer use the same field.
+        data_args = kwargs.get("data_args") or []
+        first_args = data_args[0] if data_args else {}
+        timestamp_column = "timestamp"
+        if isinstance(first_args, dict):
+            raw_timestamp_column = first_args.get("timestamp_column")
+            if isinstance(raw_timestamp_column, str) and raw_timestamp_column.strip():
+                timestamp_column = raw_timestamp_column
+
+        relative_timestamps = load_relative_timestamps(
```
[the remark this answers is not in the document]
sjmonson: Good point, but ignore for now.

src/guidellm/data/deserializers/trace_synthetic.py
```diff
@@ -0,0 +1,207 @@
...
+    base_prompt_token_ids: list[int],
+    request_index: int,
+) -> str:
+    """Build a prompt from unique prefix tokens and reusable base prompt tokens."""
+    if prompt_tokens_count <= 0:
+        return ""
+
+    unique_prefix = f"guidellm-trace-request-{request_index}: "
+    prefix_token_ids = _encode_prompt(processor, unique_prefix)
+    prompt_token_ids = (prefix_token_ids + base_prompt_token_ids)[:prompt_tokens_count]
```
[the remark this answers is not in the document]
sjmonson: Valid but a request of less then 50 tokens is unlikely.

src/guidellm/scheduler/strategies.py
```diff
@@ -671,3 +672,55 @@ def request_completed(self, request_info: RequestInfo):
...
+        start_time = await self.get_processes_start_time()
+        if not self.relative_timestamps:
+            return start_time
+
+        idx = self.next_request_index()
+        if idx > len(self.relative_timestamps):
+            # Trace exhausted: park this worker slot until the scheduler cancels
+            # the processing loop via constraint_reached_event. CancelledError
+            # propagates up cleanly, matching the exit path of all other strategies.
+            await asyncio.Event().wait()
```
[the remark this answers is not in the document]
sjmonson: Yeah this is pretty hacky/inefficient but since the trace file is the same for both profile and data this shouldn't be hit.

sjmonson:
^ Trialling some new AI review tools. I resolved the comments that I believe don't need to be fixed but check the remaining one.

> base prompt: [A, B, C, D, E]
> 
> prompt 1: [_x_, A, B, C]
> 
> prompt 2 (totally unrelated to prompt 1): [_y_, A, B, C, D]
> 
> prompt 3 (shares the same prefix structure as prompt 2 up to block D): [_y_, A, B, C, _z_, D, E]

There are two main problems with this strategy:

When you prepend tokens to a sequence there is no guarantee that you will end up with the same stream of tokens once you detokenize -> retokenize. Your prefix can merge with the first token of the prompt to form a different token which will cascade though the rest of the prompt. Usually this just results slight token count inaccuracies though occasionally you get a wider gap. In the past versions of GuideLLM we did something similar and could see as high as 25 tokens difference between requests token count and actual token count on a very long prompt. The severity of this issue depends on the tokenizer; many tokenizers are mostly unaffected as long as there is a space between the prefix and the prompt, others have many tokens that cross word boundaries. 

The other issue is that by using a single static block the bytes to tokens (and words to token) ratio is static; which can matter for all of the code around the LLM (templating, HTTP pathways, etc). Its not realistic for this ratio to be static so the test could produce biased results. 

I don't think these are deal-breakers to getting this code in but it is something we will probably address down the line.

VincentG1234:
I fixed the small issue around the trace source

> The other issue is that by using a single static block the bytes to tokens (and words to token) ratio is static; which can matter for all of the code around the LLM (templating, HTTP pathways, etc). Its not realistic for this ratio to be static so the test could produce biased results.

Thanks, this is really interesting, I had never really considered the bytes/token ratio aspect and its impact. If needed here, using a small reusable pool of diverse base prompts instead of a single static one could already help.

src/guidellm/data/deserializers/trace_synthetic.py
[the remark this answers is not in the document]
jaredoconnell: Is this noted somewhere? I can imagine a portion of requests to be that size.

jaredoconnell (dismissed):
Looks like a good first step towards trace datasets. I tested it locally and it scheduled the requests when they were expected.

The AI's note about token counts may be worth considering or documenting, because I can imagine setups with low token counts.

VincentG1234:
Good point. If you think it’s worth documenting, I can add a short note about this limitation. Another simple option could also be to keep the original random generation path for very small prompts only.

jaredoconnell:
> Good point. If you think it’s worth documenting, I can add a short note about this limitation. Another simple option could also be to keep the original random generation path for very small prompts only.

In real world use cases, if we anticipate it not being too much of a problem, documentation alone should be enough. I think it's worth documenting.

jaredoconnell (approved):

sjmonson (approved):

dbutenhof (approved):
Oops ... lost track of this one until I saw Sam's review.

sjmonson (changes requested):
Sorry to go back on my approval, but I am still seeing issues with the `data.jsonl` above. After 15 requests it hangs. Can someone else validate they can run that dataset for at least 100 requests or so.

```sh
guidellm benchmark \
    --data ./data.jsonl \
    --data-args type_=trace_synthetic \
    --profile replay \
    --rate 1000.0 \
    --request-format /v1/completions
```

VincentG1234:
> Sorry to go back on my approval, but I am still seeing issues with the `data.jsonl` above. After 15 requests it hangs. Can someone else validate they can run that dataset for at least 100 requests or so.
> 
> ```shell
> guidellm benchmark \
>     --data ./data.jsonl \
>     --data-args type_=trace_synthetic \
>     --profile replay \
>     --rate 1000.0 \
>     --request-format /v1/completions
> ```

Hello @sjmonson 
The issue is likely the `--rate 1000.0`.
In replay mode, the rate scales the original intervals, so `1000.0` makes them 1000x longer.

To accelerate the replay and reduce the intervals, it should be something like:

```bash
--rate 0.001
```
The current behavior is a bit counter-intuitive at first glance though. I can also improve the CLI/docs wording to make it clearer that the rate acts as a multiplier on the original intervals. We could also consider inverting the behavior in the future, since people tend to interpret it as a speed-up factor.

sjmonson:
> To accelerate the replay and reduce the intervals, it should be something like:
> 
> ```shell
> --rate 0.001
> ```

Oh my bad, that's what I get for not rereading the documentation after coming back to this.

> The current behavior is a bit counter-intuitive at first glance though. I can also improve the CLI/docs wording to make it clearer that the rate acts as a multiplier on the original intervals. We could also consider inverting the behavior in the future, since people tend to interpret it as a speed-up factor.

Yeah its a little counterintuitive when the option is just `--rate`, we have some upcoming work with #724 which will let each profile use their own argument names so that should help.

sjmonson (approved):
LGTM, Thanks for being such an attentive contributor!

PR #625, opened 2026-03-05
Run format job in CI

sjmonson:
## Summary

Adds formatting check to CI jobs and runs formatting against all files to ensure CI passes.

## Related Issues

- Blocked by #596 
- Blocked by #590 

---

## Use of AI

### DIFF
--- a/src/guidellm/data/builders.py
+++ b/src/guidellm/data/builders.py
@@ -219,9 +219,7 @@ def process_dataset(
     Main method to process and save a dataset with sampled prompt/output token counts.
     """
     _validate_output_suffix(output_path)
-    logger.info(
-        f"Starting dataset conversion | Input: {data} | Output: {output_path}"
-    )
+    logger.info(f"Starting dataset conversion | Input: {data} | Output: {output_path}")
 
     # Parse config
     config_obj = parse_synthetic_config(config)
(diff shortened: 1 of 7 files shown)


.github/workflows/quality.yml
```diff
@@ -39,4 +39,4 @@ jobs:
       - name: Install dependencies
         run: pip install pre-commit
       - name: Run pre-commit checks
-        run: SKIP=ruff-format pre-commit run --all-files
+        run: pre-commit run --all-files
```
dbutenhof: So we have 7 files that this will want to reformat. Actually reformatting in the CI is probably a bad idea, though; I at least wouldn't be comfortable with the CI committing changes, and otherwise it's pointless. Can we run `ruff --check`, for example, to complain but not make changes? (Granted, it doesn't hurt much to reformat in the Action workspace but then ignore them ... it just "feels wrong"...)

dbutenhof (commented):
While re-using the precommit hook is convenient... since we're not trying to commit the Action workspace changes into the repo, can we just use `ruff format --check` here?

.github/workflows/quality.yml
sjmonson: I don't really understand this concern but sure. Each CI job is isolated and the pre-commit jobs are designed to build on each-other so if a formatting change fixes a linting issue you don't get errors for both. Its cleaner to handle this in tox anyways so switched to that.

PR #630, opened 2026-03-12
Add merge policies

sjmonson:
## Summary

vLLM uses [Mergify](https://docs.mergify.com/configuration/file-format/) to configure various policy checks. Since we already have access to the app through the project, this PR adopts a few of vLLM's checks.

## Details

Adds PR notifications for:

- DCO signoff
- Needs rebase
- Remove merge commits

Also configures automatic rebase for PRs that are far behind.

## Test Plan

Review enforcement across new PRs. Mergify is currently non-blocking in CI which can be changed after we finish evaluating it.

---

## Use of AI

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -2,7 +2,7 @@ pull_request_rules:
 - name: comment-pre-commit-failure
   description: Comment on PR when pre-commit check fails
   conditions:
-    - status-failure=pre-commit
+    - status-failure = precommit-checks
     - -closed
     - -draft
   actions:
@@ -20,22 +20,10 @@ pull_request_rules:
 
         For future commits, `pre-commit` will run automatically on changed files before each commit.
 
-        > [!TIP]
-        > <details>
-        > <summary>Is <code>mypy</code> failing?</summary>
-        > <br/>
-        > <code>mypy</code> is run differently in CI. If the failure is related to this check, please use the following command to run it locally:
-        >
-        > ```bash
-        > # For mypy (substitute "3.10" with the failing version if needed)
-        > pre-commit run --hook-stage manual mypy-3.10
-        > ```
-        > </details>
-
 - name: comment-dco-failure
   description: Comment on PR when DCO check fails
   conditions:
-    - status-failure=dco
+    - status-failure = dco
     - -closed
     - -draft
   actions:
@@ -43,21 +31,8 @@ pull_request_rules:
       message: |
         Hi @{{author}}, the DCO check has failed. Please click on DCO in the Checks section for instructions on how to resolve this.
 
-- name: auto-rebase if approved, ready, and 40 commits behind main
-  conditions:
-    - base = main
-    - label=ready
-    - "#approved-reviews-by >= 1"
-    - "#commits-behind >= 40"
-    - -closed
-    - -draft
-    - -conflict
-  actions:
-    rebase: {}
-
 - name: ping author on conflicts and add 'needs-rebase' label
   conditions:
-    - label != stale
     - conflict
     - -closed
   actions:


.github/mergify.yml
```diff
@@ -0,0 +1,99 @@
...
+    comment:
+      message: |
+        Hi @{{author}}, the DCO check has failed. Please click on DCO in the Checks section for instructions on how to resolve this.
+
+- name: auto-rebase if approved, ready, and 40 commits behind main
+  conditions:
+    - base = main
+    - label=ready
+    - "#approved-reviews-by >= 1"
+    - "#commits-behind >= 40"
```
dbutenhof: Yeah -- pulled directly from vLLM ... but *40* commits behind? That seems excessive. I'm also not quite sure of the logic behind the approved count: I think it'd make more sense to gate on "no requested changes", maybe, but this seems a bit arbitrary.

.github/mergify.yml
```diff
@@ -0,0 +1,99 @@
...
+    - -draft
+  actions:
+    comment:
+      message: |
+        Hi @{{author}}, the DCO check has failed. Please click on DCO in the Checks section for instructions on how to resolve this.
+
+- name: auto-rebase if approved, ready, and 40 commits behind main
+  conditions:
+    - base = main
+    - label=ready
```
dbutenhof: The difference in spacing is a bit weird: the base branch is the only place there's spacing around the `=`, although all the comparisons are spaced. Not a problem, I guess; but the different stands out and I don't like arbitrary artifacts that unnecessarily stand out during a review.

.github/mergify.yml
```diff
@@ -0,0 +1,99 @@
...
+        > <details>
+        > <summary>Is <code>mypy</code> failing?</summary>
+        > <br/>
+        > <code>mypy</code> is run differently in CI. If the failure is related to this check, please use the following command to run it locally:
+        >
+        > ```bash
+        > # For mypy (substitute "3.10" with the failing version if needed)
+        > pre-commit run --hook-stage manual mypy-3.10
+        > ```
+        > </details>
```
dbutenhof: Since we don't use `mypy` directly in our pre-commit, I suspect this complication isn't necessary for our CI.

.github/workflows/development.yml
```diff
@@ -124,7 +124,7 @@ jobs:
           edit-mode: replace
           body: |
             <!-- pr-preview-comment -->
-            🎉 **Live Preview:** [Click here to view the live version](${{ steps.deploy.outputs.url }})
+            🎉 **Live Preview:** [Click here to view the live version](${{ steps.deploy.outputs.url }}/index.html)
```
dbutenhof, line 127: Ah -- good catch (even if a bit out of scope for "merge policies"!); I'd briefly wondered why the link didn't work even though the files were clearly there. (Although I would have expected the browser to implicitly look for an `index.html` when following a link to a folder...)

.github/mergify.yml
```diff
@@ -0,0 +1,99 @@
...
+      add:
+        - needs-rebase
+    comment:
+      message: |
+       This pull request has merge conflicts that must be resolved before it can be
+       merged. Please rebase the PR, @{{author}}.
+
+       https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork
+
+- name: block PRs containing merge commits
```
dbutenhof, line 49: I think we could also globally disable merge commits in the first place in our repo settings; although noting it and requiring the rebase here may be a bit more friendly.

.github/mergify.yml
sjmonson: We can still manual rebase, this is just intended to auto if the PR gets really behind. I didn't notice the `label=ready` requirement so I think I am going to remove it from this PR since its a little half-baked for our use-case. We can add it back later.

.github/mergify.yml
sjmonson: Copy/paste from vLLM. Planning to remove this anyways; see prev comment.

.github/workflows/development.yml
sjmonson: My bad. Was originally going to include a grab bag of CI fixes in this PR.

.github/mergify.yml
sjmonson: You can't allow merge commits on main and disallow them on PRs. I tried and it has two major problems:

1. Rulesets only apply to the primary repo, not forks.
2. If you try to push a branch, GH will reject it due to the merge commits from the point before the branch forked off of main.

.github/mergify.yml
sjmonson: Fixed

PR #632, opened 2026-03-13
Drop RC jobs and nightly PyPi publish

sjmonson:
## Summary

Our current development workflow does not involve RC branches so drop the CI automation. The nightly PyPi builds have been failing for some time due to an issue with z-stream tags. Users can just pip install from git if they need latest or use the nightly container image.

## Related Issues

- Resolves #

---

## Use of AI

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -18,61 +18,6 @@ jobs:
   ui-tests:
     uses: ./.github/workflows/ui-testing.yml
 
-  build-and-publish:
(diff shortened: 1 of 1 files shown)


sjmonson:
@mergify rebase

dbutenhof (approved):
Yeah, OK: I'm good with that -- the idea of flooding PyPi with nightly builds actually seems a bit silly. RC is perhaps more debatable, if we used the concept consistently and meaningfully, but I don't have a problem with dropping it.

PR #635, opened 2026-03-13
Drop vLLM extras group

sjmonson:
## Summary

Drop vLLM extras group and instruct users to install vLLM separately. 

## Details

Had planned to put this in #596 but wanted to give @jaredoconnell a chance to respond. The main reasons to have an `extras` group is if multiple packages are needed to make an optional feature work or a specific version is needed. vLLM is neither of those things. Not having a vLLM extras group will simplify a lot of things on the locking side for the low cost of asking a user to do `pip install vllm guidellm` instead of `pip install guidellm[vllm]`. In fact, the vLLM backend container image does not even use the extras group since it is easier to just install GuideLLM into an existing vLLM image.

## Related Issues

- Closes #634 
- Closes #633 

---

## Use of AI

### DIFF
--- a/docs/guides/vllm-python-backend.md
+++ b/docs/guides/vllm-python-backend.md
@@ -18,8 +18,6 @@ For all engine options and supported models, see vLLM's [Engine Arguments](https
 
   **Note:** Installing from the lockfile for the vLLM Python backend may not install the correct dependencies for hardware acceleration.
 
-It is also possible to install GuideLLM and vLLM via pip using `pip install guidellm[vllm]`. This method may make **hardware acceleration** (e.g. CUDA) harder to get working. See [vLLM installation](https://docs.vllm.ai/en/latest/getting_started/installation) and GPU/hardware-specific docs there. For production or GPU use, the container or existing-install path is recommended.
-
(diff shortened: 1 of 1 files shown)


docs/guides/vllm-python-backend.md
```diff
@@ -18,8 +18,6 @@ For all engine options and supported models, see vLLM's [Engine Arguments](https
 
   **Note:** Installing from the lockfile for the vLLM Python backend may not install the correct dependencies for hardware acceleration.
 
-It is also possible to install GuideLLM and vLLM via pip using `pip install guidellm[vllm]`. This method may make **hardware acceleration** (e.g. CUDA) harder to get working. See [vLLM installation](https://docs.vllm.ai/en/latest/getting_started/installation) and GPU/hardware-specific docs there. For production or GPU use, the container or existing-install path is recommended.
```
jaredoconnell, line 21: We may want to keep a link to vLLM's installation docs, but that's not required.

sjmonson: Its still linked in the "See also" section.

PR #638, opened 2026-03-16
Fix HTML on main and disable by default

sjmonson:
## Summary

Something I failed to notice during the scramble to get v0.5.4 out is that the HTML CI only generates the new version template on release, so tagging it with the current dev version will not work. Also just remove it from the defaults in preparation for phasing it out entirely.

---

## Use of AI

### DIFF
--- a/src/guidellm/settings.py
+++ b/src/guidellm/settings.py
@@ -31,8 +31,8 @@ class Environment(str, Enum):
 
 
 ENV_REPORT_MAPPING = {
-    Environment.PROD: "https://raw.githubusercontent.com/vllm-project/guidellm/refs/heads/gh-pages/ui/v0.6.0/index.html",
-    Environment.STAGING: "https://raw.githubusercontent.com/vllm-project/guidellm/refs/heads/gh-pages/ui/release/v0.6.0/index.html",
+    Environment.PROD: "https://raw.githubusercontent.com/vllm-project/guidellm/refs/heads/gh-pages/ui/v0.5.4/index.html",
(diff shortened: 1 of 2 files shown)


src/guidellm/settings.py
```diff
@@ -31,8 +31,8 @@ class Environment(str, Enum):
 
 
 ENV_REPORT_MAPPING = {
-    Environment.PROD: "https://raw.githubusercontent.com/vllm-project/guidellm/refs/heads/gh-pages/ui/v0.6.0/index.html",
-    Environment.STAGING: "https://raw.githubusercontent.com/vllm-project/guidellm/refs/heads/gh-pages/ui/release/v0.6.0/index.html",
+    Environment.PROD: "https://raw.githubusercontent.com/vllm-project/guidellm/refs/heads/gh-pages/ui/v0.5.4/index.html",
+    Environment.STAGING: "https://raw.githubusercontent.com/vllm-project/guidellm/refs/heads/gh-pages/ui/release/v0.4.0/index.html",
```
dbutenhof, line 35: I get that we haven't tagged a v0.6.0 release yet; but why do we have v0.4.0 but not v0.5.0?

dbutenhof (approved):

src/guidellm/settings.py
sjmonson: Staging does not seem to be updated automatically so the latest version available is v0.4.0. I am not even sure when the staging template is used. See https://github.com/vllm-project/guidellm/tree/gh-pages/ui/release

PR #640, opened 2026-03-17
Fix and improve the mock server

jaredoconnell:
## Summary

This fixes a few issues with the mock server, each in their own commits:
- It didn't support the list request format. Now it does.
- ~~All logs were duplicated. Now they're not.~~
- It didn't respond to CTRL-C. Now it shuts down gracefully.

## Test Plan

You can use the mock server to verify that it works.
There is also now a test for the list format requests.

---

## Use of AI

### DIFF
--- a/src/guidellm/mock_server/server.py
+++ b/src/guidellm/mock_server/server.py
@@ -53,7 +53,7 @@ class MockServer:
             timing parameters
         """
         self.config = config
-        self.app = Sanic("guidellm-mock-server", configure_logging=False)
+        self.app = Sanic("guidellm-mock-server")
         self.chat_handler = ChatCompletionsHandler(config)
         self.completions_handler = CompletionsHandler(config)
         self.tokenizer_handler = TokenizerHandler(config)


src/guidellm/mock_server/server.py
```diff
@@ -234,5 +234,5 @@ def run(self) -> None:
             debug=False,
             single_process=True,
             access_log=True,
-            register_sys_signals=False,  # Disable signal handlers for threading
+            register_sys_signals=True,
```
dbutenhof, line 237: Not critical (and maybe the explicit documentation isn't bad), but `True` seems to be the default. (Then again, `debug=False` is also the default ...)

src/guidellm/mock_server/server.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, config: MockServerConfig) -> None:
             timing parameters
         """
         self.config = config
-        self.app = Sanic("guidellm-mock-server")
+        self.app = Sanic("guidellm-mock-server", configure_logging=False)
```
dbutenhof: "Duplicate logs"? I just tried again running `guidellm mock-server` and then doing a run against it, and I'm not seeing any duplicate logging from the mock server. When I run with this change I don't see any of the mock server startup or access log messages:

```console
(guidellm) dbutenho 10:02 grpc:feature/grpc uv run guidellm mock-server --host localhost --port 8000 --model Qwen/Qwen3Guard-Gen-0.6B
✔ GuideLLM mock server starting...
  Listening on http://localhost:8000 for model Qwen/Qwen3Guard-Gen-0.6B                                                                                                                                                                
Sanic is running in PRODUCTION mode. Consider using '--debug' or '--dev' while actively developing your application.
```

vs

```console
(guidellm) dbutenho 10:10 grpc:feature/grpc uv run guidellm mock-server --host localhost --port 8000 --model Qwen/Qwen3Guard-Gen-0.6B
(guidellm) dbutenho 10:10 grpc:feature/grpc ✔ GuideLLM mock server starting...
  Listening on http://localhost:8000 for model Qwen/Qwen3Guard-Gen-0.6B                                                                                                                                                                
Main  2026-03-18 10:10:29 -0400 INFO: 
  ┌───────────────────────────────────────────────────────────────────────────────────────┐
  │                                     Sanic v25.3.0                                     │
  │                           Goin' Fast @ http://localhost:8000                          │
  ├───────────────────────┬───────────────────────────────────────────────────────────────┤
  │                       │      app: guidellm-mock-server                                │
  │     ▄███ █████ ██     │     mode: production, single worker                           │
  │    ██                 │   server: sanic, HTTP/1.1                                     │
  │     ▀███████ ███▄     │   python: 3.12.12                                             │
  │                 ██    │ platform: Linux-6.18.13-200.fc43.x86_64-x86_64-with-glibc2.42 │
  │    ████ ████████▀     │ packages: sanic-routing==23.12.0                              │
  │                       │                                                               │
  │ Build Fast. Run Fast. │                                                               │
  └───────────────────────┴───────────────────────────────────────────────────────────────┘

Main  2026-03-18 10:10:29 -0400 WARNING:  Sanic is running in PRODUCTION mode. Consider using '--debug' or '--dev' while actively developing your application.
Main  2026-03-18 10:10:29 -0400 INFO:     Starting worker 
```

Now, this isn't necessarily bad, since they interleave with the GuideLLM output and that can be annoying. I'm more curious about the "duplicate log" motivation here. (Ideally, I think we should likely have a `guidellm mock-server --debug` or an environment variable to override this, but I doubt that's worthwhile for this PR.)

jaredoconnell: Thanks for testing this. It looks like env specific factors are causing different behavior on my env vs your env. So I reverted that change.

src/guidellm/mock_server/server.py
jaredoconnell: I'm leaning towards just leaving it explicit here.

src/guidellm/mock_server/server.py
dbutenhof: Interesting.

PR #641, opened 2026-03-17
Add lazy-loading for extras packages

sjmonson:
## TODO

- [ ] ~Pending upstream change: https://github.com/scientific-python/lazy-loader/pull/168~ upstream seems dead so the module has been pulled in-house.
- [ ] ~Support for soft extras (E.g. `perf` group)~ deferred
- [ ] ~Lazy load in `__main__.py` to improve CLI responsiveness~ deferred
- [ ] Docs cleanup (howto create a lazy module, do's and don'ts)
- [ ] Actually test vLLM python (loading of vLLM tested but not a full run)

## Summary

Lazy loads extras submodules in order to defer import errors to the time of use.

## Details

TODO

## Test Plan

Without vLLM:

1. Run `guidellm benchmark run --help` and observe no errors
2. Run `uv run guidellm benchmark run --backend vllm_python --model test` and observe error with helpful message

With vLLM:

1. Run `guidellm benchmark run --help` and observe no errors
2. Run `uv run guidellm benchmark run --backend vllm_python --model test ...` and observe successful benchmark
3. Run `tox -re test-e2e` and observe that tests pass (Previously they would fail when vLLM was installed due to load times)

## Related Issues

- Replaces #636 

---

- [x] "I certify that all code in this PR is my own, except as noted below."

## Use of AI

- [x] Includes AI-assisted code completion
- [ ] Includes code generated by an AI application
- [ ] Includes AI-generated tests (NOTE: AI written tests should have a docstring that includes `## WRITTEN BY AI ##`)

---

# git log

commit 19d8aadda0eecfb3717681174366719d13bf6d20
Author: Samuel Monson <smonson@redhat.com>
Date:   Tue May 19 17:49:25 2026 -0400

    Pull in custom lazy_load package as submodule
    
    Signed-off-by: Samuel Monson <smonson@redhat.com>

commit de50d319e77c39eeaa1344460c5d6cf41df2f4e3
Author: Samuel Monson <smonson@redhat.com>
Date:   Mon Jun 1 19:50:10 2026 -0400

    Minor type fix
    
    Signed-off-by: Samuel Monson <smonson@redhat.com>

commit 60663648996a974b414672a857c3abb0c8617bb7
Author: Samuel Monson <smonson@redhat.com>
Date:   Tue Mar 17 16:33:04 2026 -0400

    Switch to lazy-loading for extras packages
    
    Signed-off-by: Samuel Monson <smonson@redhat.com>

[...]

PR #647, opened 2026-03-19
Cleanup __init__

sjmonson:
## Summary

Clean up some `__init__` package code and move uvloop config to `__init__`.

## Details

This clean up was originally a part of #641 but as that PR is blocked I decided to split it out. Removing the transformers logging config does not seem to have any real affect; I do not get logs either way. Importing any huggingface libraries incurs a significant time cost so this is a prereq to improving CLI responsiveness. Additionally uvloop should be configured as early as possible so moved the setup to `__init__`.

---

## Use of AI

### DIFF
--- a/src/guidellm/settings.py
+++ b/src/guidellm/settings.py
@@ -2,40 +2,19 @@ from __future__ import annotations
 
 import json
 from collections.abc import Sequence
-from enum import Enum
 from typing import Literal
 
-from pydantic import BaseModel, Field, model_validator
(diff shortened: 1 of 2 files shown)


jaredoconnell (approved):
The new code looks a lot simpler. Are we losing any functionality with this change?
Approving because I don't see any potential issues with this code.

sjmonson:
> ⚠️ Configuration not compatible with a branch protection setting

😢 Maybe next time then.

PR #654, opened 2026-03-24
Improve environment variables warning and validation cleanup

sjmonson:
## Summary

1. Adds a CLI helper utility to bucket detected environment variables into "valid" and "invalid"
2. Clean up some validation logic for the new `BackendArgs`.

## Details

Adds a new module to utilities for detecting environment variables supported by click and global settings. Then uses this data to split the environment into variables recognized as valid options and those that are not. This should make it easier for users to determine what environment settings are actually having an affect on GuideLLM and if any current workflows are outdated.

Also reworks some of the new `BackendArgs` validation. Instead of storing backend arguments directly in `BenchmarkGenerativeTextArgs` instead have `BenchmarkGenerativeTextArgs` construct a `BackendArgs` for the given backend type and store that. This deduplicates some validation logic removes the need for the custom `format_backend_args_error` helper.

### 1. Env Helper

Run `GUIDELLM__LOGGING__DISABLED=true GUIDELLM__REQUEST_TIMEOUT=test GUIDELLM_RATE_TYPE=sweep GUIDELLM_TARGET=http://localhost:9 uv run guidellm benchmark run`.

Backend will fail with `RuntimeError: Backend validation request failed...` but will print env detection:

```
ℹ The following environment variables are set and will be used as GuideLLM defaults (if not overridden by CLI arguments/config)
  GUIDELLM_TARGET, GUIDELLM__LOGGING__DISABLED
⚠ Warning: the following environment variables are set but NOT RECOGNIZED by GuideLLM
  GUIDELLM_RATE_TYPE, GUIDELLM__REQUEST_TIMEOUT
```

### 2. `BackendArgs` Validation

Will vLLM run: `uv run --with vllm guidellm benchmark run --backend vllm_python --model test --target http://localhost:9`.

GuideLLM should reject it will a message about target:

```
Usage: guidellm benchmark run [OPTIONS]
Try 'guidellm benchmark run --help' for help.

Error: Invalid value for --target: Value error, Target is not supported; this backend runs locally.
```

---

## Use of AI

### DIFF
--- a/src/guidellm/__main__.py
+++ b/src/guidellm/__main__.py
@@ -426,8 +426,8 @@ def run(**kwargs):  # noqa: C901
 
     # Map top-level CLI options to backend_kwargs
     backend_kwargs = kwargs.pop("backend_kwargs", {})
-    with contextlib.suppress(KeyError):
-        for alias in ("target", "model", "request_format"):
+    for alias in ("target", "model", "request_format"):
+        with contextlib.suppress(KeyError):
             backend_kwargs[alias] = kwargs.pop(alias)
     kwargs["backend_kwargs"] = backend_kwargs
 
@@ -445,7 +445,7 @@ def run(**kwargs):  # noqa: C901
             console.print_update(
                 title=(
                     "The following environment variables are set and will be used "
-                    "as GuideLLM defaults (if not overridden by CLI arguments/config)"
+                    "by GuideLLM (if not overridden by CLI arguments/config)."
                 ),
                 details=", ".join(valid_set_envs),
                 status="info",
@@ -453,8 +453,9 @@ def run(**kwargs):  # noqa: C901
         if invalid_set_envs:
             console.print_update(
                 title=(
-                    "Warning: the following environment variables are set "
-                    "but NOT RECOGNIZED by GuideLLM"
+                    "The following environment variables are set "
+                    "but not recognized by GuideLLM. Please verify "
+                    "that the benchmark is configured correctly."
                 ),
                 details=", ".join(invalid_set_envs),
                 status="warning",


src/guidellm/utils/env_validator.py
sjmonson, line 1: The public interface methods in this module are written by me, the helper methods are AI written. I am sure they are overengineered but since this is a non-critical path that will probably completely change next release IMO just validate they work. We can clean up later.

src/guidellm/__main__.py
```diff
@@ -448,37 +424,41 @@ def run(**kwargs):  # noqa: C901
...
+                    "as GuideLLM defaults (if not overridden by CLI arguments/config)"
+                ),
+                details=", ".join(valid_set_envs),
+                status="info",
+            )
+        if invalid_set_envs:
+            console.print_update(
+                title=(
+                    "Warning: the following environment variables are set "
+                    "but NOT RECOGNIZED by GuideLLM"
```
dbutenhof: I'm not crazy about the all-caps "NOT RECOGNIZED". (It's also, strictly speaking, untrue, as we've just loudly "recognized" them.)

How about "but won't be used by GuideLLM"?

I'm debating suggesting being more specific in the first phrase, because the phrasing almost suggests we're just picking environment variables out of a hat, but these are environment variables with, I don't know, maybe "config or CLI prefixes known to GuideLLM" like `GUIDELLM_` and `GUIDELLM__`. The flip side of course is that the reported names *will* follow a regular prefix pattern that's more obvious in the final form than in the message template here, and it would make the sentence a bit unwieldy.

jaredoconnell: Maybe we could say "The following environment variables are set, but have been deprecated and are no longer used by GuideLLM"

sjmonson: > (It's also, strictly speaking, untrue, as we've just loudly "recognized" them.)

If I say something like "I don't recognize the person named David Butenhof" its strictly true that I now recognize the name but that is a meaningless fact because I still don't recognize David Butenhof.

> "The following environment variables are set, but have been deprecated and are no longer used by GuideLLM"

Can't make this claim since the env var may never have existed in the first place (I have seen this before due to AI generated commands or people making inferences on what vars should exist).

sjmonson: Reworded a bit to hopefully clarify.

jaredoconnell: The new text looks good.

Issue #459, opened 2025-11-13
Add support for OpenAI Responses API


tosokin:
**Is your feature request related to a problem? Please describe.**
I need to benchmark OpenAI's Responses API, which is integrated into [LlamaStack](https://github.com/llamastack/llama-stack) agentic workflows with tool calling (RAG, MCP, function calling, etc.). The full list of available tools can be found [here](https://platform.openai.com/docs/guides/tools#available-tools).
Currently, GuideLLM does not support OpenAI's Responses API and we can't execute performance tests for agentic systems built on the Responses API, which are becoming increasingly important as organizations move beyond simple chat to tool-enabled AI agents.
The Responses API has different request/response structures and multi-step execution patterns that aren't captured by existing completion endpoints.

**Describe the solution you'd like**
Send Responses API requests with proper format:
The tools array supports different tool types, each with its own structure. For example:
file_search (RAG):

```
`vector_store_id` is a unique identifier for a document collection in LlamaStack's vector database. You get it by calling the `/v1/vector_stores` endpoint which creates a new vector store and returns its ID (like vs_xyz789). After uploading documents to this vector store, you reference it by this ID in `file_search` tool requests - LlamaStack knows where to find the embedded documents because it manages the vector store internally.

curl -X POST http://llamastack:8321/v1/openai/v1/responses \
  -H "Content-Type: application/json" \
  -d '{
    "model": "vllm-inference/llama-32-3b-instruct",
    "input": "When did the Bering Land Bridge become a national preserve?",
    "tools": [{
      "type": "file_search",
      "vector_store_ids": ["vs_xyz789"]
    }],
    "stream": false
  }'
```
MCP (Model Context Protocol):
MCP server URL is the network address where your MCP server is running. You get it by deploying an MCP server (either as a standalone process, container, or Kubernetes pod/service), which exposes an HTTP endpoint for MCP protocol communication. 
```
curl -X POST http://llamastack:8321/v1/openai/v1/responses \
  -H "Content-Type: application/json" \
  -d '{
    "model": "vllm-inference/llama-32-3b-instruct",
    "input": "Search for parks in RI using state code RI",
    "tools": [{
      "type": "mcp",
      "server_url": "http://nps-mcp-server:3005/sse",
      "server_label": "National Parks"
    }],
    "stream": false
  }'
```

Parse Responses API responses with multi-item outputs.
Here's the full response we received when testing with the MCP tool:
```
{
  "created_at": 1762867562,
  "error": null,
[...]

PR #655, opened 2026-03-25
Basic Responses API Support

jaredoconnell:
## Summary

This PR adds basic support for the responses API.
It does not add support for features often associated with the responses API, like tool calling.

## Details

- Added support for the responses API.
- Included an AI-generated section to the mock-server for the responses API. That appears to work well with vLLM, but I haven't tested it with other tools.
- Does not support the websocket API.

## Test Plan

Start vLLM, and run GuideLLM with `--request-format /v1/responses`

Example command:
```
guidellm benchmark run   --backend openai_http  --data "prompt_tokens=128,output_tokens=256"   --max-seconds 60    --profile constant --rate 1 --max-requests 5 --target http://localhost:8000 --request-format /v1/responses
```

## Related Issues

- Related to #459. Resolves it in part. It does not handle tool calling.

---

## Use of AI

### DIFF
--- a/src/guidellm/backends/openai/http.py
+++ b/src/guidellm/backends/openai/http.py
@@ -61,7 +61,7 @@ class OpenAIHttpBackendArgs(BackendArgs):
             "Request format for OpenAI-compatible server. "
             "Valid values: /v1/completions, /v1/chat/completions, "
             "/v1/responses, /v1/audio/transcriptions, /v1/audio/translations, "
-            "or legacy aliases: text_completions, chat_completions, responses, "
+            "or legacy aliases: text_completions, chat_completions, "
             "audio_transcriptions, audio_translations."
         ),
         json_schema_extra={
@@ -69,7 +69,7 @@ class OpenAIHttpBackendArgs(BackendArgs):
                 "Backend '{backend_type}' received an invalid --request-format. "
                 "Valid values: /v1/completions, /v1/chat/completions, "
                 "/v1/responses, /v1/audio/transcriptions, /v1/audio/translations, "
-                "or legacy aliases: text_completions, chat_completions, responses, "
+                "or legacy aliases: text_completions, chat_completions, "
                 "audio_transcriptions, audio_translations."
             )
         },


src/guidellm/backends/openai/http.py
```diff
@@ -108,6 +109,7 @@ def validate_request_format(cls, v: str | None) -> str | None:
 LEGACY_API_ALIASES = {
     "text_completions": "/v1/completions",
     "chat_completions": "/v1/chat/completions",
+    "responses": "/v1/responses",
```
sjmonson: The `LEGACY_API_ALIASES` are only to avoid breaking pre-existing scripts. No new aliases should be added.

```suggestion
```

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -711,3 +712,229 @@ def extract_metrics(
...
+        arguments.body = {}
+
+        if kwargs.get("model") is not None:
+            arguments.body["model"] = kwargs["model"]
+
+        if kwargs.get("stream"):
+            arguments.stream = True
+            arguments.body["stream"] = True
+            # NOTE: The Responses API does not support stream_options; usage
+            # data is delivered in the response.completed SSE event instead.
```
sjmonson: This will be a problem because unless we can query usage metrics with every token incomplete request metrics will be inaccurate. Are there any pre-completion SSE we can look at? Or any way to enable that in vLLM?

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -711,3 +712,229 @@ def extract_metrics(
...
+                    input_items.extend(prev_input)
+
+        content_parts: list[dict[str, Any]] = []
+        for col in ("text_column", "image_column", "audio_column"):
+            content_parts.extend(self._format_prompts(data.columns.get(col, []), col))
+        if content_parts:
+            input_items.append({"role": "user", "content": content_parts})
+
+        if response and response.text:
+            input_items.append({"role": "assistant", "content": response.text})
```
sjmonson: These items should be organized in roundrobin order like the other endpoints. E.g. `[<text col 1>, <image col 1>, <audio col 1>, <text col 2>, <image col 2>, <audio col 2>]` rather than `[<text col 1>, <text col 2>, <image col 1>, <image col 2>, <audio col 1>, <audio col 2>]`

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -711,3 +712,229 @@ def extract_metrics(
...
+        line = line.strip()
+
+        if not line or line.startswith("event:"):
+            return {}
+
+        if line == "data: [DONE]":
+            return None
+
+        if not line.startswith("data:"):
+            return {}
```
sjmonson: Seems like this should accomplish the same thing:

```suggestion
        if not line or not line.startswith("data:"):
            return {}

        if line == "data: [DONE]":
            return None
```

src/guidellm/backends/openai/request_handlers.py
sjmonson: Ah ok, I read down a little lower and saw `response.incomplete` can we trigger that by interrupting the stream in some way?

src/guidellm/mock_server/handlers/responses.py
```diff
@@ -0,0 +1,367 @@
...
+                        "response": completed_response,
+                        "sequence_number": seq,
+                    },
+                )
+            )
+
+            # Real vLLM does NOT send "data: [DONE]" for the Responses API
+            # (unlike chat completions). We include it here so the mock server
+            # also works with clients that rely on it as a stream terminator.
+            await stream_response.write("data: [DONE]\n\n")
```
sjmonson: If vLLM does not send a `data: [DONE]` I think skip it unless its part of the OpenAI spec.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -711,3 +712,229 @@ def extract_metrics(
             text_words=len(text.split()) if text else 0,
             text_characters=len(text) if text else 0,
         )
+
+
+@OpenAIRequestHandlerFactory.register("/v1/responses")
+class ResponsesRequestHandler(TextCompletionsRequestHandler):
```
sjmonson: Should this inherit from `OpenAIRequestHandler` rather than `TextCompletionsRequestHandler`? Just asking, I might be missing somewhere where code is reused.

src/guidellm/mock_server/handlers/responses.py
jaredoconnell: Removed.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: https://github.com/vllm-project/vllm/pull/5742 Looks like this isn't possible on the responses API. I think we should get in a call on the implications of this.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: Done.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: I just removed the base class. Does that seem better? It didn't have a proper is-a relationship.

sjmonson: No, it should at least inherit from `OpenAIRequestHandler` otherwise you violate the return type of `OpenAIRequestHandlerFactory.create`.

jaredoconnell: Good point. Fixed.

tests/unit/backends/openai/test_request_handlers.py
```diff
@@ -1640,7 +1640,7 @@ def test_class_signatures(self):
         ## WRITTEN BY AI ##
         """
         handler = ResponsesRequestHandler()
-        assert not issubclass(ResponsesRequestHandler, TextCompletionsRequestHandler)
+        assert OpenAIRequestHandler in ResponsesRequestHandler.__mro__
```
dbutenhof, line 1643: I always hate descending into the dunderverse ... but since Protocol "inheritance" isn't runtime checkable, is this really the best we can do? 😕

sjmonson: I am guessing this is an AI change since:

```
assert issubclass(ResponsesRequestHandler, OpenAIRequestHandler)
```

should work just fine.

src/guidellm/backends/openai/request_handlers.py
sjmonson: Resolving since we don't have a solution atm and not a critical blocker.

tests/unit/backends/openai/test_request_handlers.py
jaredoconnell: This test is AI generated. Here's what it says about the approaches:

> Both assert the same thing — that ResponsesRequestHandler inherits from OpenAIRequestHandler. The difference is purely about runtime behavior with Protocol:
> 
> issubclass on a Protocol requires the protocol to be decorated with @runtime_checkable. Without that decorator, Python raises TypeError at runtime. That's what broke the test earlier.
> 
> __mro__ is just a tuple of the class's method resolution order. Checking membership in it always works regardless of @runtime_checkable, because it's not doing any protocol-specific logic — it's plain tuple lookup.
> 
> Since OpenAIRequestHandler is a Protocol without @runtime_checkable, only the __mro__ check works at runtime. If @runtime_checkable were added to the Protocol, both would be equivalent. But adding that decorator just to make a test assertion work isn't a good reason — __mro__ is fine here.

dbutenhof: Yeah -- I started writing that to say "this should be `issubclass`", but I stopped to actually test that manually and found the `TypeError`, and discovered the cause. It's weird. I left the comment just as a ... gee, that's odd. I don't like using the internal dunder-names; but I agree that for a unit test, it's likely better than annotating our base protocol unless the actual code needs to use `issubclass` or `isinstance` on this hierarchy in the future.

```console
>>> a = TextCompletionsRequestHandler()
>>> isinstance(a, OpenAIRequestHandler)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.12/typing.py", line 1917, in __instancecheck__
    raise TypeError("Instance and class checks can only be used with"
TypeError: Instance and class checks can only be used with @runtime_checkable protocols
```

dbutenhof (approved):

Issue #416, opened 2025-10-17
Support benchmarking of reasoning and tool calling Chat Completions requests


bbrowning:
**Is your feature request related to a problem? Please describe.**
I need to benchmark the performance of reasoning and tool call parsers in vLLM to understand the overall impact they have on time to time to first token, inter-token latency, inter-chunk latency (as in time between streaming chunks sent back to the client), and overall throughput of the vLLM server(s) so I can iterate and improve on the implementations of these in vLLM guided by quantitative performance numbers.

**Describe the solution you'd like**

I'd like us to ensure we can take inputs that express complex Chat Completions requests, that may include fields such as `tools` with a potentially large number of functions passed as well as fields like `tool_choice` to control auto/required/none/named function tool parsing. `tools` may need to be a new top-level input field we support while something like `tool_choice` may be fine to come in as an `extra_body`, since it's usually a simple string value. An example in our docs of how to do basic tool call benchmarking would help guide users in setting things up properly.

I'd like us to validate the metrics we emit today are accurate and not misleading when reasoning parser or tool call parsers are involved in the vLLM server-side.

For example, a tool call parser may buffer dozens of tokens server-side while waiting to determine if the response is generating a tool call or not. When does the clock start for time-to-first-token? When does it stop? Same for inter-token latency. The current OpenAI Chat Completions implementation has a bug here today in how it calculates ITL if tool call parsers are in use. We'll need to audit all our metrics carefully to ensure they don't make assumptions that are no longer valid in a world where the server is buffering, parsing, and changing what the client sees vs what the model generated.

What should we report when inter-token latency is say 20 ms, but inter-chunk latency (ie time between streaming chunks coming back to the client) is multiple seconds or even minutes? Do we need to start tracking a new metric that represents the time between streaming tokens back to the client, which is distinct and separate from the ITL? How would we optimize tool call parsers to lower inter-chunk latency objectively, backed by data?

**Describe alternatives you've considered**
I have not run across another project so far that can accurately measure inter-token latency when tool call parsing is in use. I've only really dug into the code of `vllm bench serve`, `llm-load-test`, and `guidellm`. But, am open to considering other projects if they're known to work for more advanced reasoning and tool calling cases.

I've also considered creating my own project, as GuideLLM today appears more focused on raw prompts/responses via the Completions API instead of newer APIs like Chat Completions or Responses. However, there are recent changes to make GuideLLM more pluggable so decided to start here first.

**Additional context**

[...]

PR #687, opened 2026-04-03
Process tool call requests for chat completions API

jaredoconnell:
## Summary

This is a first step towards getting tool calling working.

## Details

- Allows counting and reporting of tool calls and their associated tokens.
- Supports the chat completions API.
- Displays the tool metrics in the CSV output and the console output, in addition to the JSON/YAML output, which includes everything.

## Test Plan

- Start vLLM with tool calling enabled:
  ```
  --enable-auto-tool-choice --tool-call-parser hermes
  ```
- Run a benchmark with backend args set such that they specify a tool and request a tool call.
   Example:
   ```
   guidellm benchmark   --target http://localhost:8000   --model "Qwen/Qwen3-0.6B" --data mixed_tool_call_prompts.jsonl --profile constant  --rate 1   --max-requests 5 --backend-kwargs '{
    "extras": {
      "body": {
        "tools": [
          {
            "type": "function",
            "function": {
              "name": "get_weather",
              "description": "Get the current weather for a location",
              "parameters": {
                "type": "object",
                "properties": {
                  "location": {
                    "type": "string",
                    "description": "City and state, e.g. San Francisco, CA"
                  }
                },
                "required": ["location"]
              }
            }
          }
        ],
        "tool_choice": "required"
      }
    }
  }'
   ```

## Related Issues

- Works towards #416 

---

## Use of AI

### DIFF
--- a/src/guidellm/backends/openai/request_handlers.py
+++ b/src/guidellm/backends/openai/request_handlers.py
@@ -406,7 +406,7 @@ class ChatCompletionsRequestHandler(TextCompletionsRequestHandler):
 
     def __init__(self):
         super().__init__()
-        self.streaming_tool_calls: dict[int, dict] = {}
+        self.streaming_tool_call_indices: set[int] = set()
 
     def _format_prompts(
         self, column_data: list[dict[str, Any]], column_type: str
@@ -554,10 +554,11 @@ class ChatCompletionsRequestHandler(TextCompletionsRequestHandler):
         choices, usage = self.extract_choices_and_usage(response)
         choice: dict[str, dict] = choices[0] if choices else {}
         message = choice.get("message", {})
-        # content is null for tool-call-only responses (text=None means text
-        # metrics are excluded); empty string for rare edge cases where the
-        # model returns no text (text="" means text metrics are 0).
+        # Per spec, content is null when the model produces tool calls instead
+        # of text. text=None excludes text metrics; text="" means text metrics
+        # are 0 for the rare empty-content edge case.
         text = message.get("content")
+        # Per spec, tool_calls is present when content is null.
         tool_calls = message.get("tool_calls") if text is None else None
         if text is None and not tool_calls:
             text = ""
@@ -601,26 +602,11 @@ class ChatCompletionsRequestHandler(TextCompletionsRequestHandler):
             self.streaming_texts.append(content)
             updated = True
 
-        # Tool call streaming sends incremental chunks via delta.tool_calls.
-        # Each chunk (tc_delta) carries an "index" identifying which tool call
-        # it belongs to (for parallel tool calls), plus partial fragments of
-        # function.name and function.arguments that must be concatenated across
-        # multiple SSE events to reconstruct the complete call.
+        # tool_calls is absent from text-only deltas; empty default is
+        # spec-correct. Each entry carries a required "index" identifying the
+        # tool call; missing index raises KeyError → errored request.
         for tc_delta in delta.get("tool_calls", []):
-            idx = tc_delta.get("index", 0)
-            if idx not in self.streaming_tool_calls:
-                # First chunk for this tool call: initialize with id and type
-                self.streaming_tool_calls[idx] = {
-                    "id": tc_delta.get("id", ""),
-                    "type": tc_delta.get("type", "function"),
-                    "function": {"name": "", "arguments": ""},
-                }
-            tc = self.streaming_tool_calls[idx]
-            fn_delta = tc_delta.get("function", {})
-            if fn_name := fn_delta.get("name"):
-                tc["function"]["name"] += fn_name
-            if fn_args := fn_delta.get("arguments"):
-                tc["function"]["arguments"] += fn_args
+            self.streaming_tool_call_indices.add(tc_delta["index"])
             updated = True
 
         if usage:
@@ -638,13 +624,13 @@ class ChatCompletionsRequestHandler(TextCompletionsRequestHandler):
         :return: Standardized GenerationResponse with concatenated content and metrics
         """
         text = "".join(self.streaming_texts) or None
-        has_tool_calls = text is None and bool(self.streaming_tool_calls)
+        has_tool_calls = text is None and bool(self.streaming_tool_call_indices)
         if text is None and not has_tool_calls:
             text = ""
         input_metrics, output_metrics = self.extract_metrics(self.streaming_usage, text)
         if has_tool_calls:
             output_metrics.tool_call_tokens = output_metrics.text_tokens
-            output_metrics.tool_call_count = len(self.streaming_tool_calls)
+            output_metrics.tool_call_count = len(self.streaming_tool_call_indices)
 
         return GenerationResponse(
             request_id=request.request_id,
--- a/src/guidellm/backends/openai/request_handlers.py
+++ b/src/guidellm/backends/openai/request_handlers.py
@@ -335,19 +335,29 @@ class TextCompletionsRequestHandler(OpenAIRequestHandler):
         return response.get("choices", []), response.get("usage", {})
 
     def extract_metrics(
-        self, usage: dict[str, int | dict[str, int]] | None, text: str
+        self, usage: dict[str, int | dict[str, int]] | None, text: str | None
     ) -> tuple[UsageMetrics, UsageMetrics]:
         """
         Extract input and output usage metrics from API response usage data.
 
         :param usage: Usage data dictionary from API response
-        :param text: Generated text for calculating word and character counts
+        :param text: Generated text for calculating word and character counts.
+            None means text is not applicable (metrics will be None);
+            empty string means text was applicable but empty (metrics will be 0).
         :return: Tuple of input_metrics and output_metrics as UsageMetrics objects
         """
+        if text is None:
+            # text not applicable (e.g. tool-call-only) — exclude from aggregation
+            text_words = None
+            text_chars = None
+        else:
+            text_words = len(text.split())
+            text_chars = len(text)
+
         if not usage:
             return UsageMetrics(), UsageMetrics(
-                text_words=len(text.split()) if text else 0,
-                text_characters=len(text) if text else 0,
+                text_words=text_words,
+                text_characters=text_chars,
             )
 
         input_details: dict[str, int] = cast(
@@ -374,8 +384,8 @@ class TextCompletionsRequestHandler(OpenAIRequestHandler):
                 or usage_metrics.get("completion_tokens")
                 or 0
             ),
-            text_words=len(text.split()) if text else 0,
-            text_characters=len(text) if text else 0,
+            text_words=text_words,
+            text_characters=text_chars,
             image_tokens=output_details.get("image_tokens"),
             video_tokens=output_details.get("video_tokens"),
             audio_tokens=output_details.get("audio_tokens"),
@@ -525,27 +535,6 @@ class ChatCompletionsRequestHandler(TextCompletionsRequestHandler):
 
         return arguments
 
-    @staticmethod
-    def _tool_calls_to_text(tool_calls: list[dict]) -> str:
-        """Serialize a ``tool_calls`` array to a JSON string."""
-        # orjson.dumps returns bytes; stdlib json.dumps returns str
-        raw = json.dumps(tool_calls)
-        return raw.decode("utf-8") if isinstance(raw, bytes) else raw
-
-    @staticmethod
-    def _add_tool_call_metrics(
-        output_metrics: UsageMetrics, tool_call_count: int
-    ) -> None:
-        """Tag output metrics with tool call info (subset of text metrics).
-
-        Sets ``tool_call_tokens`` equal to ``text_tokens`` (since the server
-        reports all completion tokens together) and records the number of
-        individual tool calls. These fields are additive metadata -- they do
-        not affect ``total_tokens``.
-        """
-        output_metrics.tool_call_tokens = output_metrics.text_tokens
-        output_metrics.tool_call_count = tool_call_count
-
     def compile_non_streaming(
         self,
         request: GenerationRequest,
@@ -556,8 +545,7 @@ class ChatCompletionsRequestHandler(TextCompletionsRequestHandler):
         Process a complete chat completion response.
 
         Extracts content from the message object within choices, handling the nested
-        structure specific to chat completion endpoints. When the model returns tool
-        calls instead of text content, the tool calls are serialized as JSON text.
+        structure specific to chat completion endpoints.
 
         :param request: Original generation request
         :param response: Complete API response containing choices and usage data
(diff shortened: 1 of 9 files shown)


src/guidellm/backends/openai/request_handlers.py
```diff
@@ -520,6 +525,27 @@ def format(  # noqa: C901
...
+        """Tag output metrics with tool call info (subset of text metrics).
+
+        Sets ``tool_call_tokens`` equal to ``text_tokens`` (since the server
+        reports all completion tokens together) and records the number of
+        individual tool calls. These fields are additive metadata -- they do
+        not affect ``total_tokens``.
+        """
+        output_metrics.tool_call_tokens = output_metrics.text_tokens
+        output_metrics.tool_call_count = tool_call_count
+
```
sjmonson: This method is too simple. Just inline it.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -569,11 +604,34 @@ def add_streaming_line(self, line: str) -> int | None:
...
             self.streaming_texts.append(content)
             updated = True
 
+        # Tool call streaming sends incremental chunks via delta.tool_calls.
+        # Each chunk (tc_delta) carries an "index" identifying which tool call
+        # it belongs to (for parallel tool calls), plus partial fragments of
+        # function.name and function.arguments that must be concatenated across
+        # multiple SSE events to reconstruct the complete call.
+        for tc_delta in delta.get("tool_calls", []):
+            idx = tc_delta.get("index", 0)
```
sjmonson: Is `0` a safe default? If not having this field is invalid the request should error.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -569,11 +604,34 @@ def add_streaming_line(self, line: str) -> int | None:
...
+            if idx not in self.streaming_tool_calls:
+                # First chunk for this tool call: initialize with id and type
+                self.streaming_tool_calls[idx] = {
+                    "id": tc_delta.get("id", ""),
+                    "type": tc_delta.get("type", "function"),
```
sjmonson: Likewise to above. Never turn an invalid request into a valid one using defaults. Though I didn't check; these could be optional.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -530,16 +556,24 @@ def compile_non_streaming(
...
-        text = choice.get("message", {}).get("content", "")
+        message = choice.get("message", {})
+        text = message.get("content") or ""
+        # Tool call responses set content=null and put output in tool_calls
+        tool_calls = message.get("tool_calls") if not text else None
+        if tool_calls:
+            text = self._tool_calls_to_text(tool_calls)
         input_metrics, output_metrics = self.extract_metrics(usage, text)
+        if tool_calls:
+            self._add_tool_call_metrics(output_metrics, len(tool_calls))
```
sjmonson: Took me a while to figure this out. Don't try to make up metrics from the raw JSON dump of tool calls. That does not tell us anything. The logic of `extract_metrics` needs to be changed; perhaps to exclude text based metrics if `text` is null. Then you can get rid of `_tool_calls_to_text`.

sjmonson: Actually it already does exclude text metrics if text is empty. Maybe change it to `text=""` -> text metrics are `0` and `text=None` -> text metrics are `None`.

sjmonson (changes requested):

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -530,16 +556,24 @@ def compile_non_streaming(
...
         :param request: Original generation request
         :param response: Complete API response containing choices and usage data
         :return: Standardized GenerationResponse with extracted content and metrics
         """
         choices, usage = self.extract_choices_and_usage(response)
         choice: dict[str, dict] = choices[0] if choices else {}
-        text = choice.get("message", {}).get("content", "")
+        message = choice.get("message", {})
+        text = message.get("content") or ""
+        # Tool call responses set content=null and put output in tool_calls
```
dbutenhof: > Tool call responses set content=null

So is there a meaningful difference between `message["content"] == None` and `message["content"] == ""`? The comment implies there might be; but if so, the code deliberately blurs the line with the `or ""`. Can this be clarified in either the comment or the code?

src/guidellm/backends/openai/request_handlers.py
dbutenhof: > these could be optional

Sure -- but a direct consequence is that if `index` was missing above, we're creating `[0] = {"id": "", "type": "function"}` here ... is that odd? I suppose it makes sense if `"index" not in tc_delta` means that the tool calls aren't parallel? (In which case defaulting to `0` kinda makes sense, and a tweak to the comment could clarify.)

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -569,11 +604,34 @@ def add_streaming_line(self, line: str) -> int | None:
...
+        for tc_delta in delta.get("tool_calls", []):
+            idx = tc_delta.get("index", 0)
+            if idx not in self.streaming_tool_calls:
+                # First chunk for this tool call: initialize with id and type
+                self.streaming_tool_calls[idx] = {
+                    "id": tc_delta.get("id", ""),
+                    "type": tc_delta.get("type", "function"),
+                    "function": {"name": "", "arguments": ""},
+                }
+            tc = self.streaming_tool_calls[idx]
```
dbutenhof: I'm actually not sure how much difference this makes in Python, but in general I'd think it would be more efficient (and probably a bit more readable) to create `tc = { ... }` and *then* assign `self.streaming_tool_calls[idx] = tc` rather than creating the element and then fetching it.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: Done.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: It appears that an empty string response is a rare edge case where the model actually responds with text, but that response is empty. I added a comment that clarified this case and explicitly notes that it's treated identically to the None case.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: It would basically make it so that a server that doesn't index unchunked tool calls would work. I removed it, so now it will just error if it's missing the index.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: I removed this section.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: Done. Let me know if I misinterpreted anything.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: I simplified this section.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -569,11 +596,19 @@ def add_streaming_line(self, line: str) -> int | None:
...
+        delta = choice.get("delta", {}) if choices else {}
 
-        if choices and (content := choice.get("delta", {}).get("content")):
+        if content := delta.get("content"):
             self.streaming_texts.append(content)
             updated = True
 
+        # tool_calls is absent from text-only deltas; empty default is
+        # spec-correct. Each entry carries a required "index" identifying the
+        # tool call; missing index raises KeyError → errored request.
```
dbutenhof: No checking that `index` exists? While it's probably an unlikely event, `KeyError index` is going to be hard for a user to figure out through the layers of traceback.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -538,8 +553,19 @@ def compile_non_streaming(
         """
         choices, usage = self.extract_choices_and_usage(response)
         choice: dict[str, dict] = choices[0] if choices else {}
-        text = choice.get("message", {}).get("content", "")
+        message = choice.get("message", {})
+        # Per spec, content is null when the model produces tool calls instead
+        # of text. text=None excludes text metrics; text="" means text metrics
```
dbutenhof: If `text=None` "excludes text metrics", what does that really mean? The logic here is counting a return with 0 tokens, which seems to be correct: but should we be collaborating on some vast "excludes text metrics" conspiracy here instead?

Something about that repeated `text is None` phrase keeps bothering my brain, and I don't see a way around it ... but if we shouldn't be converting that to `text = ""`, then we could avoid that second test.

dbutenhof (dismissed):
A few questions / minor concerns to consider, but looks OK.

jaredoconnell:
Using the following command with the 7B parameter model, I was able to get it to use tool calls some of the time, which confirmed proper differentiation between all text metrics, and tool call only metrics. The smaller models were not able to automatically execute tool calls.

```bash
vllm serve Qwen/Qwen2.5-7B-Instruct --tool-call-parser hermes --enable-auto-tool-choice
```

```bash
guidellm benchmark run   --target "http://localhost:8000"   --request-format chat_completions   --data mixed_tool_call_prompts.jsonl   --data-column-mapper '{"text_column":"text"}'   --backend-kwargs '{"extras":{"body":{"tools":[{"type":"function","function":{"name":"get_weather","description":"Get current weather for a location","parameters":{"type":"object","properties":{"location":{"type":"string"}},"required":["location"]}}}],"tool_choice":"auto"}}}' --max-requests 5 --profile constant --rate 1
```

With this input file:
```jsonl
{"prompt": "Use the get_weather function to check the weather in San Francisco."}
{"prompt": "Explain how photosynthesis works."}
{"prompt": "Call get_weather for Tokyo to get the forecast."}
{"prompt": "Write a haiku about mountains."}
{"prompt": "Please call the get_weather tool for London."}
{"prompt": "Summarize the history of the Roman Empire."}
```

Here is a snippet of the output:
```
ℹ Text Metrics Statistics (Completed Requests)
|===========|=======|=======|======|======|======|=======|======|======|======|========|======|======|
| Benchmark | Input Tokens             |||| Input Words             |||| Input Characters         ||||
| Strategy  | Per Request  || Per Second || Per Request || Per Second || Per Request  || Per Second ||
|           | Mdn   | p95   | Mdn  | Mean | Mdn  | p95   | Mdn  | Mean | Mdn  | p95    | Mdn  | Mean |
|-----------|-------|-------|------|------|------|-------|------|------|------|--------|------|------|
| constant  | 172.0 | 176.0 | 27.8 | 42.5 | 7.0  | 11.0  | 1.2  | 1.7  | 44.0 | 67.0   | 7.3  | 10.9 |
|===========|=======|=======|======|======|======|=======|======|======|======|========|======|======|
| Benchmark | Output Tokens            |||| Output Words            |||| Output Characters        ||||
| Strategy  | Per Request  || Per Second || Per Request || Per Second || Per Request  || Per Second ||
|           | Mdn   | p95   | Mdn  | Mean | Mdn  | p95   | Mdn  | Mean | Mdn  | p95    | Mdn  | Mean |
|-----------|-------|-------|------|------|------|-------|------|------|------|--------|------|------|
| constant  | 21.0  | 393.0 | 5.6  | 23.4 | 0.0  | 269.0 | 0.0  | 13.9 | 0.0  | 1757.0 | 0.0  | 90.7 |
|===========|=======|=======|======|======|======|=======|======|======|======|========|======|======|


ℹ Tool Call Metrics Statistics (Completed Requests)
|===========|=======|======|======|======|=======|======|======|======|
| Benchmark | Output Tokens           |||| Output Count            ||||
| Strategy  | Per Request || Per Second || Per Request || Per Second ||
|           | Mdn   | p95  | Mdn  | Mean | Mdn   | p95  | Mdn  | Mean |
|-----------|-------|------|------|------|-------|------|------|------|
| constant  | 21.0  | 21.0 | 5.0  | 7.7  | 1.0   | 1.0  | 0.2  | 0.4  |
|===========|=======|======|======|======|=======|======|======|======|
```

dbutenhof:
Is it odd that Mdn & P95 tool call "output tokens per request" are both 21, the exact same value as the "text output tokens per request" Mdn?? Are all tool calls returning 21 tokens? Maybe it's just a coincidence...

jaredoconnell:
> Is it odd that Mdn & P95 tool call "output tokens per request" are both 21, the exact same value as the "text output tokens per request" Mdn?? Are all tool calls returning 21 tokens? Maybe it's just a coincidence...

Tool call calls are a subset of text, so the tool calls appear to be the ones representing the median, but a lot of the non-tool-call ones appear to be larger, bringing up the mean.

So I think this is a good sign that it's a subset, and that tool calls aren't being double-counted.

src/guidellm/backends/openai/request_handlers.py
sjmonson: If index not existing is a malformed request IMO we leave it as a key error for now. There are too many different errors that could occur; its better to just let the backend handler capture the whole traceback.

Also guessing this comment is AI written. Can probably just be removed (I remove/replace most AI comments since they are usually unhelpful streams of artificial conscience).

sjmonson (changes requested):
Addressed my main concerns though also see Dave's comments.

sjmonson (dismissed):
Srry, above was supposed to be approval.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: I agree that adding an error would make it a little easier to debug, but it wouldn't help much either. They likely wouldn't hit this with vLLM or any good server.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: `text=None` will result in the text metrics being set to `None`.

I re-wrote this section without AI. The checks seemed a little unnecessary.

dbutenhof (dismissed):

jaredoconnell:
I made some changes to ensure that the edge case of simultaneous content and tool calls result in tool calls properly getting counted. I did some research and that are models that can do that.

dbutenhof (dismissed):
OK, if we can really have `content` as well as `tool_calls` in the same response, it seems odd that we only have `completion_tokens`, which suggests that we're counting the `content` along with the tool call output as `tool_call_tokens`. Or is there just some subtlety I'm missing here?

dbutenhof (dismissed):
I think the situation kinda sucks; but we didn't make it and it seems we have limited visibility and no control. (The glitch is that we'll take the blame for "bad" counts.)

As I suggested in Slack, another option might be to break out a separate bucket of "mixed tool_call / text" response content that doesn't attribute it all or none to tool calling -- that's not very satisfying, but at least it provides visibility.

(Plus, if we make that display conditional on non-zero, and it's really rare, maybe nobody'll have to deal with it...)

jaredoconnell:
I've since learned that Qwen shows thinking tokens when tool calls are auto, but doesn't when tool calls are required. This added stat provides good visibility on what the model is doing without needing to inspect the outputs.

dbutenhof (approved):

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -554,18 +554,17 @@ def compile_non_streaming(
...
         input_metrics, output_metrics = self.extract_metrics(usage, text)
-        if tool_calls:
-            output_metrics.tool_call_tokens = output_metrics.text_tokens
-            output_metrics.tool_call_count = len(tool_calls)
+        if raw_tool_calls:
+            output_metrics.tool_call_count = len(raw_tool_calls)
+            if text is None:  # tool-only turn
+                output_metrics.tool_call_tokens = output_metrics.text_tokens
+            else:  # mixed content + tool call turn
+                output_metrics.mixed_content_tool_tokens = output_metrics.text_tokens
```
sjmonson, line 567: In the streaming case can one chunk contain a tool call and another chunk contain text or will every chunk contain the tool call when it occurs? I.e. Can this branch eval differently for each chunk? Mostly asking for down the line when the metrics accumulated we usually just use the most recent chunk.

sjmonson (approved):
LGTM with one clarifying question.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: Per-chunk it will just accumulate the tool indexes in a set, so duplicate items will be saved over time. And text will be accumulated as it is without this change.
But it just evaluates tokens at the end, and if you have a tool call in one chunk it will evaluate as `mixed_content_tool_tokens`. Does this address your question?

sjmonson: Yep! Thanks.

PR #689, opened 2026-04-10
Fix long import times due to processing all pydantic subclasses

### LINKED ISSUE #688: `import guidellm` can take a very long time in some environments
### Bug Description

Importing can take a very long time if other pydantic models have been imported first.

`reload_parent_schemas` causes a >1 min import slowdown when other pydantic libraries are loaded first.

### Expected Behavior

I expect that `import` time will be minimal and have no side effects.

### Steps to Reproduce

```bash
#!/usr/bin/env bash
set -e

VENV=$(mktemp -d)/venv
echo "Creating fresh venv at $VENV"
python3 -m venv "$VENV"
source "$VENV/bin/activate"

echo "Installing guidellm 0.6.0 and litellm..."
pip install -q guidellm==0.6.0 litellm 2>&1 | tail -1

echo ""
echo "=== Import litellm THEN guidellm (slow) ==="
time python -c "
import litellm
from guidellm.benchmark.schemas.generative.report import GenerativeBenchmarksReport
"

echo ""
echo "=== Import guidellm THEN litellm (fast) ==="
time python -c "
from guidellm.benchmark.schemas.generative.report import GenerativeBenchmarksReport
import litellm
"

echo ""
echo "=== Import litellm only (baseline) ==="
time python -c "
import litellm
"

echo ""
echo "=== Import guidellm only (baseline) ==="
time python -c "
from guidellm.benchmark.schemas.generative.report import GenerativeBenchmarksReport
"

rm -rf "$(dirname "$VENV")"
```

I got the following output:

```
=== Import litellm THEN guidellm (slow) ===

real	1m26.701s
user	1m21.577s
sys	0m6.405s

=== Import guidellm THEN litellm (fast) ===

real	0m9.420s
user	0m9.562s
sys	0m0.847s

=== Import litellm only (baseline) ===

real	0m2.576s
user	0m2.345s
sys	0m0.183s

=== Import guidellm only (baseline) ===

real	0m7.047s
user	0m7.367s
sys	0m0.867s
```

### Operating System

Linux 6.1.0 (Debian)

### Python Version

Python 3.12.11

### GuideLLM Version

0.6.0

### Installation Method

pip install guidellm

### Installation Details

_No response_

```shell

```

### Additional Context

_No response_

jaredoconnell:
## Summary

Other libraries that use pydantic could end up being processed by GuideLLM, significantly increasing the number of classes processed, noticeably increasing import times for GuideLLM.

Also, in my testing, it makes it so the startup times are less than half of what they were before.

## Details

- Rather than just searching every Pydantic subclass, it only searches sub-classes that we created. So no other libraries will add to the processing burden.
- On my mac, within the Python virtual environment, this change resulted in a change from around 27.5 seconds before the benchmark starts to 6.5 seconds. This is due to how long it spent processing the classes on my local env. It has been super slow since this was originally added in the refactor.
- The only classes that use this functionality are `BenchmarkConfig`, `BenchmarkGenerativeTextArgs`, and `Profile`,

## Test Plan

- Run benchmarks as normal, and confirm that the options that depend on registry mix in work.

## Related Issues

- Resolves #688 

---

## Use of AI

### DIFF
--- a/src/guidellm/schemas/base.py
+++ b/src/guidellm/schemas/base.py
@@ -87,7 +87,10 @@ class ReloadableBaseModel(BaseModel):
         subclasses, so third-party Pydantic models loaded in the same process
         are not visited.
         """
-        potential_parents: set[type[BaseModel]] = {ReloadableBaseModel, StandardBaseModel}
+        potential_parents: set[type[BaseModel]] = {
+            ReloadableBaseModel,
+            StandardBaseModel
+        }
         stack: list[type[BaseModel]] = list(potential_parents)
 
         while stack:
--- a/src/guidellm/schemas/base.py
+++ b/src/guidellm/schemas/base.py
@@ -89,7 +89,7 @@ class ReloadableBaseModel(BaseModel):
         """
         potential_parents: set[type[BaseModel]] = {
             ReloadableBaseModel,
-            StandardBaseModel
+            StandardBaseModel,
         }
         stack: list[type[BaseModel]] = list(potential_parents)
 
--- a/src/guidellm/schemas/base.py
+++ b/src/guidellm/schemas/base.py
@@ -79,14 +79,16 @@ class ReloadableBaseModel(BaseModel):
     @classmethod
     def reload_parent_schemas(cls):
         """
-        Recursively reload schemas for all parent Pydantic models.
+        Recursively reload schemas for all Pydantic models that reference this
+        class in their field annotations.
 
-        Traverses the inheritance hierarchy to find all parent classes that
-        are Pydantic models and triggers schema rebuilding on each to ensure
-        that any changes in child models are reflected in parent schemas.
+        Walks the subclass trees of guidellm's own model roots
+        (ReloadableBaseModel and StandardBaseModel) rather than all BaseModel
+        subclasses, so third-party Pydantic models loaded in the same process
+        are not visited.
         """
-        potential_parents: set[type[BaseModel]] = {BaseModel}
-        stack: list[type[BaseModel]] = [BaseModel]
+        potential_parents: set[type[BaseModel]] = {ReloadableBaseModel, StandardBaseModel}
+        stack: list[type[BaseModel]] = list(potential_parents)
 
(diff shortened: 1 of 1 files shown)


src/guidellm/schemas/base.py
dbutenhof, line 102: I shouldn't say this since it's not part of your change ... but `<set>.add(<x>)` is more efficient and generally "cleaner" than `if <x> not in <set>: <set>.add(<x>)`. You probably shouldn't change this since it's out of scope and irrelevant to the problem, but if you did I wouldn't mind. 😁

src/guidellm/schemas/base.py
```diff
@@ -79,14 +79,19 @@ def reload_schema(cls, parents: bool = True) -> None:
     @classmethod
     def reload_parent_schemas(cls):
         """
-        Recursively reload schemas for all parent Pydantic models.
+        Recursively reload schemas for all Pydantic models that reference this
```
dbutenhof, line 82: "for all GuideLLM Pydantic models ..." ?? Yeah, "that reference this" covers that base, but it's a bit less direct.

dbutenhof (approved):
Excellent! But, as Webb might say, just a few nits ...

dbutenhof:
Ha -- you might consider modifying the PR title: it's an incomplete sentence, and it displays as "Fix long import times due to #689", which just looks odd!

src/guidellm/schemas/base.py
jaredoconnell: What type of change would you suggest?

src/guidellm/schemas/base.py
jaredoconnell: I don't think this change would work due to the `stack` lines.

src/guidellm/schemas/base.py
dbutenhof: I thought a bit about how to clarify without over-complicating ... on second thought, it's probably not worth changing.

dbutenhof (approved):

PR #691, opened 2026-04-12
Adapt multi-turn documentation to new format

oparoz:
## Summary

The script in the documentation for running a multi-turn benchmark using TurnPivot had some issues

## Details

- Fixed a typo
- Remove comments from script as they prevented it from running properly in bash

## Test Plan

N/A

## Related Issues

N/A

---

## Use of AI

### DIFF
--- a/docs/guides/multiturn.md
+++ b/docs/guides/multiturn.md
@@ -349,9 +349,9 @@ guidellm benchmark run \
   --profile concurrent \
   --rate 10 \
   --max-requests 150 \
-  --data "prefix_tokens=512,prompt_tokens=128,output_tokens=256" \      # Turn 1
-  --data "prompt_tokens=256,prompt_token_stdev=32,output_tokens=128" \  # Turn 2
-  --data "prompt_tokens=64,output_tokens=128,output_tokens_stdev=16" \  # Turn 3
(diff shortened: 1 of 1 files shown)


sjmonson (changes requested):
Both formats are supported but I am more partial to using the arg-string format in our docs. Could you modify this PR to just fix the typo?

oparoz:
> Both formats are supported but I am more partial to using the arg-string format in our docs. Could you modify this PR to just fix the typo?

Yes, no problem (and good to know). I've just tested to be sure and it wouldn't work, but it was due to the inline comments.

sjmonson:
> Yes, no problem (and good to know). I've just tested to be sure and it wouldn't work, but it was due to the inline comments.

Oh yeah, you can drop the comments since a working example is preferable.

dbutenhof:
> > Both formats are supported but I am more partial to using the arg-string format in our docs. Could you modify this PR to just fix the typo?
> 
> Yes, no problem (and good to know). I've just tested to be sure and it wouldn't work, but it was due to the inline comments.

Yes, and that's an interesting point. Ideally, we should support copy-paste on the sample commands that are complete, and bash doesn't support comments (or anything) after the `\` "continuation" character (because it's actually quoting the newline that follows).

Those comments are informative, but I suspect we should remove them.

sjmonson:
Obsolete

PR #692, opened 2026-04-13
Apply tool calling stats to the responses API

jaredoconnell:
## Summary

This PR applies the same tool counting logic to the responses API.

## Details

- Counts tool-only or mixed responses.
- Is pretty simple.

## Test Plan

Start vLLM with tools enabled:
`--enable-auto-tool-choice --tool-call-parser hermes`

Run a benchmark:
  ```
  guidellm benchmark run   --target "http://localhost:8000"   --request-format /v1/responses   --data mixed_tool_call_prompts.jsonl   --data-column-mapper '{"text_column":"text"}'   --backend-kwargs '{"stream":false,"extras":{"body":{"tools":[{"type":"function","name":"get_weather","description":"Get current weather for a location","parameters":{"type":"object","properties":{"location":{"type":"string"}},"required":["location"]},"strict":true}],"tool_choice":"auto"}}}'   --max-requests 5 --profile constant --rate 0.2
  ```

Here is mixed_tool_call_prompts.jsonl
```jsonl
{"prompt": "Use the get_weather function to check the weather in San Francisco."}
{"prompt": "Explain how photosynthesis works."}
{"prompt": "Call get_weather for Tokyo to get the forecast."}
{"prompt": "Write a haiku about mountains."}
{"prompt": "Please call the get_weather tool for London."}
{"prompt": "Summarize the history of the Roman Empire."}
```

The logic is different with tool call auto vs required. It appears that required typically results in tool call only responses.

## Related Issues

- Works towards #416

---

## Use of AI

### DIFF
--- a/src/guidellm/backends/openai/request_handlers.py
+++ b/src/guidellm/backends/openai/request_handlers.py
@@ -898,7 +898,11 @@ class ResponsesRequestHandler(OpenAIRequestHandler):
         response: dict,
     ) -> GenerationResponse:
         text = self._extract_output_text(response)
-        tool_call_count = self._count_function_calls(response)
+        tool_call_count = sum(
+            1
+            for item in response.get("output", [])
+            if item.get("type") == "function_call"
+        )
         if text is None and not tool_call_count:
             text = ""
         usage = response.get("usage", {})
@@ -1039,7 +1043,7 @@ class ResponsesRequestHandler(OpenAIRequestHandler):
     def _extract_output_text(response: dict) -> str | None:
         """Extract generated text from a Responses API response object.
 
-        Returns ``None`` when no message/output_text items exist (e.g. tool-call-
+        :returns: ``None`` when no message/output_text items exist (e.g. tool-call-
         only responses), so callers can distinguish "no text" from "empty text".
         """
         texts: list[str] = []
@@ -1050,15 +1054,6 @@ class ResponsesRequestHandler(OpenAIRequestHandler):
                 if part.get("type") == "output_text":
                     texts.append(part.get("text", ""))
         return "".join(texts) if texts else None
-
-    @staticmethod
-    def _count_function_calls(response: dict) -> int:
-        """Count ``function_call`` output items in a Responses API response."""
-        return sum(
-            1
-            for item in response.get("output", [])
-            if item.get("type") == "function_call"
-        )
 
 
 @OpenAIRequestHandlerFactory.register("/pooling")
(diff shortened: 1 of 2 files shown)


src/guidellm/backends/openai/request_handlers.py
```diff
@@ -1011,16 +1036,29 @@ def extract_metrics(
...
+        return "".join(texts) if texts else None
+
+    @staticmethod
+    def _count_function_calls(response: dict) -> int:
+        """Count ``function_call`` output items in a Responses API response."""
+        return sum(
+            1
+            for item in response.get("output", [])
+            if item.get("type") == "function_call"
+        )
```
sjmonson: Unless linting complains just inline this. Also this is personal preference since I believe it performs similarly but maybe

```python
sum(map(
    lambda item: item.get("type") == "function_call",
    response.get("output", [])
))
```

is a little cleaner.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -1011,16 +1036,29 @@ def extract_metrics(
         )
 
     @staticmethod
-    def _extract_output_text(response: dict) -> str:
-        """Extract generated text from a Responses API response object."""
+    def _extract_output_text(response: dict) -> str | None:
+        """Extract generated text from a Responses API response object.
+
+        Returns ``None`` when no message/output_text items exist (e.g. tool-call-
+        only responses), so callers can distinguish "no text" from "empty text".
```
sjmonson: Since you are editing this anyways use the `:returns:` docstring format.

sjmonson (dismissed):
Minor nits but overall code looks good.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: I asked AI about the performance difference, and here it its analysis:
"The current generator expression has a slight theoretical edge because the if clause means sum only processes yielded 1 values (skipping non-matches entirely), whereas map produces a value for every item (True or False) and sum has to add them all. On the other hand, map avoids generator protocol overhead — but the lambda call overhead roughly cancels that out. Net effect: noise-level difference."

For code that I'm submitting, I think summing 1s makes more sense to me than using map. I'll inline it.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: Fixed.

sjmonson (approved):

PR #697, opened 2026-04-16
Support server-side conversation history on responses API

jaredoconnell:
## Summary

Add support for server-side conversation history via previous_response_id on the /v1/responses endpoint. This allows multi-turn benchmarks to offload conversation state to the server instead of replaying the full history in every request, which is critical for benchmarking servers that use response storage (e.g. vLLM with VLLM_ENABLE_RESPONSES_API_STORE=1).

## Details

 - Add `server_history` option to OpenAIHTTPBackend, validated to only work with /v1/responses
 - Update ResponsesRequestHandler.format() to set previous_response_id from the prior response instead of replaying history when `server_history` is enabled. 
- Add unit tests for validation and request formatting with server_history
- Document server-side history usage and requirements in docs/guides/multiturn.md

## Test Plan

- Run pytest tests/unit/backends/openai/test_http.py -k server_history
- Run pytest tests/unit/backends/openai/test_request_handlers.py -k server_history
- (Integration) Against a vLLM server with VLLM_ENABLE_RESPONSES_API_STORE=1, run a multiturn benchmark with --request-format /v1/responses --backend-kwargs '{"server_history": true}' and confirm turn 2+ requests use previous_response_id

---

## Use of AI

### DIFF
--- a/docs/guides/multiturn.md
+++ b/docs/guides/multiturn.md
@@ -40,13 +40,16 @@ When executing a multiturn benchmark, GuideLLM:
 4. **Sends the next turn** (prompt_i) along with the conversation history
 5. **Repeat from (2)** for the `n` given turns
 
-For `/v1/chat/completions`, the conversation history is passed as a messages array with alternating user and assistant roles. For `/v1/completions`, the history is concatenated as a single prompt string.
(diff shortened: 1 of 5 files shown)


sjmonson (approved):
LGTM

PR #700, opened 2026-04-20
Add a min-requests constraint

sjmonson:
## Summary

Adds a `--min-requests` constraint that acts like `--max-requests` but keeps scheduling requests until the last request under the threshold completes.

## Details

The normal `--max-requests` variant can have unexpectedly low per-request throughput / latency due to request trail-off at the end of the benchmark. The current solution to this problem is to set `--max-requests` so high that the proportion of trail-off time to total benchmark time is small. If we continue to schedule requests even after hitting the constraint we ensure that the requested rate is maintained for the entire duration of measurement.

Note that `--min-requests` is a bit of a misnomer when combined with other constraints, since any other active constraints can trigger the benchmark to end before `--min-requests`. Other name suggestions are welcome.

## Test Plan

Here is an example benchmark which should cause `--min-requests` to behave differently from `--max-requests`:

```sh
guidellm benchmark run \
    --target http://127.0.0.1:8000 \
    --request-format /v1/completions \
    --profile concurrent \
    --rate 30 \
    --data prompt_tokens=50,output_tokens=50 \
    --min-requests 50
```

---

## Use of AI


src/guidellm/scheduler/constraints/request.py
```diff
@@ -268,6 +268,117 @@ def _validate_max_duration(
...
 
 
+@ConstraintsInitializerFactory.register(  # type: ignore[arg-type]
+    ["min_number", "min_num", "min_requests", "min_req"]
+)
+class MinNumberConstraint(PydanticConstraintInitializer):
+    """
+    Constraint that limits execution based on minimum request counts.
+
+    Like MinNumberConstraint but instead of stopping request generation after reaching
```
jaredoconnell, line 278: Mistake: It should say "Like **Max**NumberConstraint"

I think this wording doesn't emphasize the nuances of this implementation enough. Maybe clarify generation and processing, and why this may be helpful. It's identical except that it doesn't stop queueing until the max processed quantity is reached.

src/guidellm/scheduler/constraints/request.py
```diff
@@ -268,6 +268,117 @@ def _validate_max_duration(
         return value[0] if isinstance(value, list) and len(value) == 1 else value
 
 
+@ConstraintsInitializerFactory.register(  # type: ignore[arg-type]
+    ["min_number", "min_num", "min_requests", "min_req"]
```
jaredoconnell, line 272: It may make sense to instead rename this to max-processed. I think this would be less confusing. But I can see the argument for `min`, since it's going to keep scheduling past that until the max-processed is reached. So I'm not sure what should be done.

jaredoconnell (commented):
I'm not 100% sure of the arg name. I can see arguments both ways.

But I think this would be a good opportunity to add a markdown file detailing all of the constraints.

I added two comments.

src/guidellm/scheduler/constraints/request.py
sjmonson: Yeah... `max-processed` is both a little too vague and also incorrect since we can end up processing more requests then set. I think min is fine actually. I'll just add some notes to the docs that clarify constraints are OR not AND. Maybe in the future we can support AND constraint combinations.

PR #707, opened 2026-04-24
Fix CSV column misalignment across multiple benchmarks

leehyeoklee:
## Summary:
Improve the CSV output formatter to dynamically construct a unified master header when running multiple benchmark profiles sequentially, preventing column misalignment by safely padding missing metrics.

## Details:

Currently, the CSV output formatter (`GenerativeBenchmarkerCSV`) locks the header structure of the entire result to the first executed benchmark when multiple benchmark profiles are run sequentially (sweep profile). 
As a result, if subsequent benchmarks measure new metrics or different modalities, it causes the data columns to misalign or break against the headers. 
To resolve this, the logic was improved to dynamically construct a superset (master header) of all benchmark result headers and safely pad any missing metrics with empty strings.

## Test Plan:
```
guidellm benchmark \
  --profile sweep \
  --max-seconds 30 \
  --data "prompt_tokens=256,output_tokens=128"
```

### Before
image

### After
image

## Related Issues

---

## Use of AI

### DIFF
--- a/src/guidellm/benchmark/outputs/csv.py
+++ b/src/guidellm/benchmark/outputs/csv.py
@@ -117,8 +117,9 @@ class GenerativeBenchmarkerCSV(GenerativeBenchmarkerOutput):
 
         with output_path.open("w", newline="") as file:
             writer = csv.writer(file)
-            headers: list[list[str]] = []
-            rows: list[list[str | int | float]] = []
+
+            row_maps: list[dict[tuple[str, ...], str | int | float]] = []
+            ordered_headers: dict[tuple[str, ...], None] = {}
 
             for benchmark in report.benchmarks:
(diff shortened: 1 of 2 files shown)


jaredoconnell (dismissed):
Looks good to me. If you have a chance, this does expose a gap in our unit tests, so if you could add tests that would be great, too.

leehyeoklee:
@jaredoconnell 
Thanks for the feedback! I've added the missing unit tests in the latest commit. Please take a look. 😊

jaredoconnell:
I pushed a commit to fix a missing case and follow project requirements.
Since I contributed, I'll have another maintainer give the final approval.

sjmonson (approved):

dbutenhof (approved):

PR #709, opened 2026-04-28
Checks for no valid requests in processed dataset

jaredoconnell:
## Summary

Adds warnings and errors so that problems in the datasets are visible, rather than silently failing.
In main, if the dataset isn't processed correctly, GuideLLM deadlocks while the datasets are pending.

## Details

- Adds failure condition in scheduler/worker_group.py to abort if there are no requests.
- Adds some other warnings when there are no matching columns

## Test Plan

Currently, due to the other bugs, this command triggers the problem:
```
guidellm benchmark run \
  --target http://localhost:8000/ \
  --data "abisee/cnn_dailymail" \
  --data-args '{"name": "3.0.0"}' \
  --data-column-mapper '{"text_column":"article"}'
```

## Use of AI

### DIFF
--- a/src/guidellm/data/loaders.py
+++ b/src/guidellm/data/loaders.py
@@ -85,8 +85,6 @@ class DatasetsIterator(TorchIterableDataset[DataT]):
     def set_epoch(self, epoch: int):
         self.epoch = epoch
 
-    _CONSECUTIVE_ERROR_WARN_THRESHOLD = 50
-
     def generator(  # noqa: C901
         self,
         max_items: int | None = None,
@@ -97,7 +95,6 @@ class DatasetsIterator(TorchIterableDataset[DataT]):
         gen_count = 0
         yield_count = 0
         error_count = 0
-        consecutive_errors = 0
 
         with contextlib.suppress(StopIteration):
             dataset_iters = []
@@ -127,27 +124,25 @@ class DatasetsIterator(TorchIterableDataset[DataT]):
                         row = preprocessor(row)
 
                     result = self.finalizer(row)
+                    # Filter empty results (e.g. column mapper matched
+                    # no columns, so finalizer returned an empty list)
+                    if not result:
+                        continue
                     yield result
                     yield_count += 1
-                    consecutive_errors = 0
                 except StopIteration:
                     raise  # Stop iteration when any dataset is exhausted
                 except Exception as err:  # noqa: BLE001 # Exception logged
                     error_count += 1
-                    consecutive_errors += 1
-                    logger.error(f"Skipping data row due to error: {err}")
-                    if consecutive_errors == self._CONSECUTIVE_ERROR_WARN_THRESHOLD:
-                        logger.warning(
-                            f"{self._CONSECUTIVE_ERROR_WARN_THRESHOLD} "
-                            f"consecutive row errors encountered "
-                            f"({error_count} total errors out of "
-                            f"{gen_count} rows). Check data format and "
-                            f"preprocessor configuration."
-                        )
+                    logger.error(
+                        "Skipping data row due to error: {}. "
+                        "Check data format and preprocessor configuration.",
+                        err,
+                    )
                     gen_count -= 1
 
         if gen_count > 0 and yield_count == 0:
-            logger.warning(
+            raise ValueError(
                 f"Dataset iterator processed {gen_count} rows but yielded "
                 f"zero results ({error_count} errors). This usually means "
                 f"the column mapper could not match any dataset columns. "
--- a/src/guidellm/data/loaders.py
+++ b/src/guidellm/data/loaders.py
@@ -95,7 +95,6 @@ class DatasetsIterator(TorchIterableDataset[DataT]):
         gen_count = 0
         yield_count = 0
         error_count = 0
-        empty_count = 0
 
         with contextlib.suppress(StopIteration):
             dataset_iters = []
--- a/src/guidellm/data/loaders.py
+++ b/src/guidellm/data/loaders.py
@@ -85,7 +85,9 @@ class DatasetsIterator(TorchIterableDataset[DataT]):
     def set_epoch(self, epoch: int):
         self.epoch = epoch
 
-    def generator(
+    _CONSECUTIVE_ERROR_WARN_THRESHOLD = 50
+
+    def generator(  # noqa: C901
         self,
         max_items: int | None = None,
         modulus: int | None = None,
@@ -93,6 +95,9 @@ class DatasetsIterator(TorchIterableDataset[DataT]):
         epoch: int = 0,
     ) -> Iterator[DataT]:
         gen_count = 0
+        yield_count = 0
+        error_count = 0
+        consecutive_errors = 0
 
         with contextlib.suppress(StopIteration):
             dataset_iters = []
@@ -121,12 +126,33 @@ class DatasetsIterator(TorchIterableDataset[DataT]):
                     for preprocessor in self.preprocessors:
                         row = preprocessor(row)
 
-                    yield self.finalizer(row)
+                    result = self.finalizer(row)
+                    yield result
+                    yield_count += 1
+                    consecutive_errors = 0
                 except StopIteration:
                     raise  # Stop iteration when any dataset is exhausted
                 except Exception as err:  # noqa: BLE001 # Exception logged
+                    error_count += 1
+                    consecutive_errors += 1
                     logger.error(f"Skipping data row due to error: {err}")
+                    if consecutive_errors == self._CONSECUTIVE_ERROR_WARN_THRESHOLD:
+                        logger.warning(
+                            f"{self._CONSECUTIVE_ERROR_WARN_THRESHOLD} "
+                            f"consecutive row errors encountered "
+                            f"({error_count} total errors out of "
+                            f"{gen_count} rows). Check data format and "
+                            f"preprocessor configuration."
+                        )
                     gen_count -= 1
+
+        if gen_count > 0 and yield_count == 0:
+            logger.warning(
+                f"Dataset iterator processed {gen_count} rows but yielded "
+                f"zero results ({error_count} errors). This usually means "
+                f"the column mapper could not match any dataset columns. "
+                f"Check --data-column-mapper and dataset column names."
+            )
 
         if max_items is not None and gen_count < max_items:
             raise ValueError(
--- a/src/guidellm/data/loaders.py
+++ b/src/guidellm/data/loaders.py
@@ -95,6 +95,7 @@ class DatasetsIterator(TorchIterableDataset[DataT]):
         gen_count = 0
         yield_count = 0
         error_count = 0
+        empty_count = 0
 
         with contextlib.suppress(StopIteration):
             dataset_iters = []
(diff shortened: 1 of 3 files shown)


src/guidellm/data/loaders.py
```diff
@@ -85,14 +85,19 @@ def __iter__(self) -> Iterator[DataT]:
     def set_epoch(self, epoch: int):
         self.epoch = epoch
 
-    def generator(
+    _CONSECUTIVE_ERROR_WARN_THRESHOLD = 50
```
sjmonson: Move this somewhere in the global `Settings` class.

src/guidellm/data/preprocessors/mappers.py
```diff
@@ -202,6 +203,27 @@ def setup_data(
...
+        elif not any(
+            ct == "text_column" for ct, _turn in self.datasets_column_mappings
+        ):
+            mapped_types = sorted({ct for ct, _ in self.datasets_column_mappings})
+            logger.warning(
+                "GenerativeColumnMapper mapped columns but none resolved to "
+                f"'text_column'. Mapped types: {mapped_types}. "
+                "Requests will have no text content."
+            )
+
```
sjmonson, line 229: No this will break Geospatial since it doesn't have a text_column and this class in inherited. Also it will break for anything not text like VL or audio.

src/guidellm/data/loaders.py
```diff
@@ -121,13 +126,34 @@ def generator(
...
+                    consecutive_errors += 1
                     logger.error(f"Skipping data row due to error: {err}")
+                    if consecutive_errors == self._CONSECUTIVE_ERROR_WARN_THRESHOLD:
+                        logger.warning(
+                            f"{self._CONSECUTIVE_ERROR_WARN_THRESHOLD} "
+                            f"consecutive row errors encountered "
+                            f"({error_count} total errors out of "
+                            f"{gen_count} rows). Check data format and "
+                            f"preprocessor configuration."
+                        )
```
sjmonson: Need to start nagging about using f-strings in logs. Replace with proper log formatting.

src/guidellm/data/loaders.py
```diff
@@ -121,13 +126,34 @@ def generator(
...
                     gen_count -= 1
 
+        if gen_count > 0 and yield_count == 0:
+            logger.warning(
+                f"Dataset iterator processed {gen_count} rows but yielded "
+                f"zero results ({error_count} errors). This usually means "
+                f"the column mapper could not match any dataset columns. "
+                f"Check --data-column-mapper and dataset column names."
+            )
```
sjmonson, line 149: f-strings

src/guidellm/scheduler/worker_group.py
```diff
@@ -589,13 +589,32 @@ def _turn_iter(
...
+                    if empty_count == 1:
+                        logger.warning(
+                            "requests_generator received an empty conversation "
+                            "from the data pipeline (no turns). This usually "
+                            "means the column mapper could not match dataset "
+                            "columns. Check --data-column-mapper settings."
+                        )
+                    continue
+
+                yield conversation
```
sjmonson: Handle this in `data` not here. The logic in this function should be as simple as we can make it because its a bottleneck.

src/guidellm/data/preprocessors/mappers.py
sjmonson: I think the only good way to do this right now is in `__call__` error if the list is empty then lean on the dataloader error count to handle if a dataset is completely empty.

sjmonson (changes requested):

src/guidellm/data/preprocessors/mappers.py
sjmonson: Actually that should probably happen in the finalizer since its the last thing to touch the request.

src/guidellm/scheduler/worker_group.py
sjmonson: Should clarify, if the handling in data is right this is entirely unnecessary since we ensure we don't emit an empty conversation.

jaredoconnell: Sounds good. I switched it to raise a ValueError in data if there are 0 valid rows after processing. Here's an example of the error:
```
26-04-28 18:36:45|ERROR            |guidellm.scheduler.worker_group:requests_generator:610 - Error generating requests: Caught ValueError in DataLoader worker process 0.
Original Traceback (most recent call last):
  File "/Users/joconnel/Documents/projects/ai/guidellm/.venv3.12/lib/python3.12/site-packages/torch/utils/data/_utils/worker.py", line 349, in _worker_loop
    data = fetcher.fetch(index)  # type: ignore[possibly-undefined]
           ^^^^^^^^^^^^^^^^^^^^
  File "/Users/joconnel/Documents/projects/ai/guidellm/.venv3.12/lib/python3.12/site-packages/torch/utils/data/_utils/fetch.py", line 33, in fetch
    data.append(next(self.dataset_iter))
                ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/joconnel/Documents/projects/ai/guidellm/src/guidellm/data/loaders.py", line 81, in __iter__
    yield from self.generator(
  File "/Users/joconnel/Documents/projects/ai/guidellm/src/guidellm/data/loaders.py", line 154, in generator
    raise ValueError(
ValueError: Dataset iterator processed 287113 rows but yielded zero results (0 errors). This usually means the column mapper could not match any dataset columns. Check --data-column-mapper and dataset column names.
```

src/guidellm/data/loaders.py
jaredoconnell: This one is now a ValueError, so f-strings remain unless you'd prefer another template format.

src/guidellm/data/loaders.py
jaredoconnell: ~~Fixed.~~ Actually, I removed this. The logger errors are probably enough.

src/guidellm/data/preprocessors/mappers.py
jaredoconnell: I switched that warning to debug, and clarified that it's okay if text isn't needed for that dataset. I can remove it, and let me know if you'd rather me do that, but it could be useful info for people.

What should maybe happen in the finalizer? All of these checks are now around where we first would notice the problems.

src/guidellm/data/loaders.py
jaredoconnell: I removed this.

src/guidellm/data/loaders.py
sjmonson: No that's fine. It was not really a big deal to begin with. I just want to start enforcing no f-strings in logs since its bad practice.

src/guidellm/data/loaders.py
sjmonson: Why? This solved a big problem we currently have.

src/guidellm/data/preprocessors/mappers.py
sjmonson: For `if not self.datasets_column_mappings:` if the user does not provide a custom column mapping this log message will be huge and meaningless. A user is not going to parse two large python objects embed in the middle of text. The `"text_column"` columns message has a similar problem. The whole `available_cols` is kinda hacky and doesn't add much to the error other then listing out every column of every dataset. Logging doesn't even work properly right now. I could name a bunch more reasons.

---

> I think the only good way to do this right now is in __call__ error if the list is empty then lean on the dataloader error count to handle if a dataset is completely empty.

This should happen in the finalizer because any of the preprocessors could result in an empty dataset. Though with your other changes maybe not needed.

src/guidellm/data/preprocessors/mappers.py
dbutenhof: Can't there be duplicate columns between the datasets? It might be better to make this a `set()` rather than a list.

src/guidellm/data/loaders.py
dbutenhof: I'm not sure that reporting an error count here really helps without some context. ("Something happened -- I'm not helping you figure out what it was".) This probably isn't a big issue one way or the other, but you could reference the <n> ERROR log entries previously generated... or (in some cases, though probably overkill here) I've gone to such ridiculous lengths as to keep a set of errors (either by class or full text) to report with the status.

sjmonson (changes requested):
See my new comments above.

src/guidellm/data/preprocessors/mappers.py
jaredoconnell: Yes, there can theoretically be duplicates. Leaving them in as a list may be useful for debugging purposes so that users can be aware of duplicates.

Another option is to use a counter, so it would show the count of duplicate columns.

src/guidellm/data/loaders.py
jaredoconnell: The current code logs `logger.error(f"Skipping data row due to error: {err}")`

And this warning was just on top of that. Is there another design that you'd like me to switch to?

src/guidellm/data/loaders.py
jaredoconnell: The context of how many rows there are is at least slightly helpful context because then you can know if it's every row, just a few rows, or something else.
The other errors are logged, so the user just needs to scroll up to those errors. But yes, logging isn't always guaranteed.

Anything more than what's proposed might be overkill.

src/guidellm/data/preprocessors/mappers.py
dbutenhof: Unless the dataset column sets are regular it might not even be obvious where the break between datasets is. Then again, it's likely rare to specify more than a few. But I think we're really saying "you're trying to map columns that don't exist" and I'm not sure the full list (or a counted list) is really likely to be all that helpful vs the simple set.

src/guidellm/data/loaders.py
dbutenhof: We're not talking about a "count of rows" so much as a count of whatever errors occurred across processing all the rows -- which I'm not convinced helps much without knowing the nature of the errors. (They may be all the same error, or a set of distinct errors.) Once they've gone through the log to find the `ERROR` lines, I'm not sure knowing the count here even helps.

dbutenhof: > `f"zero results ({error_count} errors; {gen_count - error_count} "`

This construct low-key bugs me, for several reasons, and I tried hard to ignore that ... without success.

1. because we show <gen_count> and we know the yield is 0, so showing both <error_count> and (<gen_count> - <error_count>) seems to just make the message longer without adding anything.
2. because the semicolon seems grammatically inappropriate here, and that kinda stuff bothers me.
3. because error rows are, by definition, also "empty", (since the exception handler has nothing to replace the intended operation of `finalizer`), so <error_count> is less a separate quantity than "the subset of empty rows that have an excuse for being empty (detailed elsewhere)" 😆 .

src/guidellm/data/loaders.py
sjmonson: I don't really want to go back and forth on this. If I find time I'll add what I was talking about in a follow-up.

sjmonson (approved):
My concerns are resolved. Leaving the comments open that @dbutenhof added to in-case any of those still need to be addressed.

src/guidellm/data/loaders.py
dbutenhof: Ha -- since Sam called my bluff by approving, I won't hold it up over awkward wording in an error message that shouldn't happen.

Not that I'd mind if you tweaked it ... but I'll resolve this and approve.

dbutenhof (approved):

PR #710, opened 2026-04-29
feat: Add embeddings endpoint support (MVP)

maryamtahhan:
Adds minimal embeddings support following the pooling pattern (~113 line handler).

Core Implementation:
- EmbeddingsRequestHandler for /v1/embeddings endpoint
- Formats requests with single text or list of texts
- Returns GenerationResponse (reuses existing infrastructure)
- Registered /v1/embeddings in DEFAULT_API_PATHS
- Added embeddings alias to LEGACY_API_ALIASES

Synthetic Data Support:
- Allow output_tokens=0 in SyntheticTextDatasetConfig
- Fix config parser to accept single key=value pairs (e.g., "prompt_tokens=64")
- Fix IntegerRangeSampler to support 0 tokens and prevent invalid ranges

Embeddings-Aware Console Output:
- Auto-detect embeddings mode (zero output tokens)
- Hide text metrics table (words/characters) for embeddings
- Hide streaming metrics (TTFT/ITL/TPOT) for embeddings
- Focus output on relevant metrics: latency, tokens/sec, throughput

Tests & Documentation:
- 13 comprehensive unit tests for EmbeddingsRequestHandler
- Complete embeddings benchmarking guide in docs/guides/embeddings.md
- Examples for all benchmark profiles and configuration options

## Details

- [ ]

## Test Plan

-

## Related Issues

- Resolves #

---

## Use of AI


src/guidellm/data/schemas.py
```diff
@@ -62,22 +62,23 @@ class PreprocessDatasetConfig(DataConfig):
...
     output_tokens_min: int | None = Field(
         description="The minimum number of text tokens retained or added to outputs.",
-        gt=0,
+        ge=0,
         default=None,
     )
     output_tokens_max: int | None = Field(
         description="The maximum number of text tokens retained or added to outputs.",
-        gt=0,
+        ge=0,
```
sjmonson: The `PreprocessDatasetConfig` is unrelated to synthetic data. Drop these changes.

src/guidellm/data/schemas.py
```diff
@@ -131,12 +132,12 @@ class SyntheticTextDatasetConfig(DataConfig):
...
     output_tokens_stdev: int | None = Field(
         description="The standard deviation of the tokens generated for outputs.",
-        gt=0,
+        ge=0,
         default=None,
     )
     output_tokens_min: int | None = Field(
         description="The minimum number of text tokens generated for outputs.",
-        gt=0,
+        ge=0,
```
sjmonson: Since these are `None` by default I don't think this change is needed.

src/guidellm/data/schemas.py
```diff
@@ -131,12 +132,12 @@ class SyntheticTextDatasetConfig(DataConfig):
     )
```
sjmonson: This was left unchanged. I think the intention was to make it `ge=0` However I think it would be better to do:

```suggestion
    output_tokens: int | None = Field(
        description="The average number of text tokens generated for outputs.",
        gt=0,
        default=None,
    )
```

And omit output token columns in the dataset if None.

src/guidellm/utils/random.py
```diff
@@ -24,14 +24,17 @@ def __iter__(self) -> Iterator[int]:
...
             )
         calc_max = self.max_value
         if calc_max is None:
             calc_max = (
                 self.average + 5 * self.variance if self.variance else self.average
             )
 
+        # Ensure calc_max >= calc_min
+        calc_max = max(calc_min, calc_max)
+
```
sjmonson: Due to `output_tokens` change above this won't be needed.

src/guidellm/backends/openai/http.py
```diff
@@ -118,6 +119,7 @@ def validate_request_format(cls, v: str | None) -> str | None:
 LEGACY_API_ALIASES = {
     "text_completions": "/v1/completions",
     "chat_completions": "/v1/chat/completions",
+    "embeddings": "/v1/embeddings",
```
sjmonson: `LEGACY_API_ALIASES` is legacy because we are moving away from this format. Please remove.

src/guidellm/benchmark/outputs/console.py
```diff
@@ -209,6 +209,18 @@ def validated_kwargs(cls, *_args, **_kwargs) -> dict[str, Any]:
...
+        Check if all benchmarks have zero output tokens (embeddings mode).
+
+        :param report: The benchmark report to check
+        :return: True if all benchmarks have zero output tokens
+        """
+        return all(
+            benchmark.metrics.output_token_count.successful.total_sum == 0
+            for benchmark in report.benchmarks
+        )
+
```
sjmonson: I think this is too fragile; embedding are not the only reason that we could and up with 0 successful tokens. Need a better metric to determine this.

src/guidellm/benchmark/outputs/console.py
```diff
@@ -446,21 +466,22 @@ def print_request_latency_table(self, report: GenerativeBenchmarksReport):
...
+                columns.add_stats(
+                    benchmark.metrics.inter_token_latency_ms,
+                    group="ITL",
+                    name="ms",
+                )
+                columns.add_stats(
+                    benchmark.metrics.time_per_output_token_ms,
+                    group="TPOT",
+                    name="ms",
+                )
```
sjmonson: These should be `None` if we are not streaming, if values are `None` then `columns.add_stats` will dash out entries. If I revert the changes to this file and run a test it shows zeros so that is a bug. Would prefer the dashed out route. E.g.

```md
ℹ Request Latency Statistics (Completed Requests)
|============|=========|========|=====|=====|=====|=====|=====|=====|
| Benchmark  | Request Latency || TTFT     || ITL      || TPOT     ||
| Strategy   | Sec             || ms       || ms       || ms       ||
|            | Mdn     | p95    | Mdn | p95 | Mdn | p95 | Mdn | p95 |
|------------|---------|--------|-----|-----|-----|-----|-----|-----|
| concurrent | 0.1     | 0.2    | --  | --  | --  | --  | --  | --  |
|============|=========|========|=====|=====|=====|=====|=====|=====|
```

Looks a little nicer then a two column table. I.e. Revert this change and figure out why `benchmark.metrics.time_to_first_token_ms` and others are `0.0` and not `None` when non-streaming.

src/guidellm/benchmark/outputs/console.py
```diff
@@ -426,11 +441,16 @@ def print_request_counts_table(self, report: GenerativeBenchmarksReport):
             title="Request Token Statistics (Completed Requests)",
         )
 
-    def print_request_latency_table(self, report: GenerativeBenchmarksReport):
+    def print_request_latency_table(
+        self,
+        report: GenerativeBenchmarksReport,
+        hide_streaming_metrics: bool = False,
+    ):
```
sjmonson: With change to streaming metrics this will no longer be necessary.

src/guidellm/benchmark/outputs/console.py
sjmonson: See other comments below, I think we can get to the point where this is completely unnecessary.

src/guidellm/benchmark/outputs/console.py
```diff
@@ -219,14 +231,17 @@ async def finalize(self, report: GenerativeBenchmarksReport) -> str:
         :param report: The completed benchmark report
         :return: Status message indicating output location
         """
+        is_embeddings = self._is_embeddings_benchmark(report)
+
         self.print_run_summary_table(report)
-        self.print_text_table(report)
+        if not is_embeddings:
+            self.print_text_table(report)
```
sjmonson: I actually think the info in this table is useful for embeddings, at least the input tokens/words/chars side is. So maybe just remove this check?

src/guidellm/benchmark/outputs/console.py
```diff
@@ -219,14 +231,17 @@ async def finalize(self, report: GenerativeBenchmarksReport) -> str:
...
         self.print_image_table(report)
         self.print_video_table(report)
         self.print_audio_table(report)
         self.print_tool_call_table(report)
         self.print_request_counts_table(report)
-        self.print_request_latency_table(report)
+        self.print_request_latency_table(report, hide_streaming_metrics=is_embeddings)
```
sjmonson: With change to streaming metrics this will no longer be necessary.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -1119,3 +1120,115 @@ def format(
...
+        if len(input_texts) == 1:
+            arguments.body["input"] = input_texts[0]
+        else:
+            arguments.body["input"] = input_texts
+
+        # Add optional parameters
+        if kwargs.get("encoding_format"):
+            arguments.body["encoding_format"] = kwargs["encoding_format"]
+        if kwargs.get("dimensions"):
+            arguments.body["dimensions"] = kwargs["dimensions"]
```
sjmonson: This does not work since there is no way to set `kwargs["encoding_format"]` or `kwargs["dimensions"]` in the current backend. Just drop these lines altogether since the extras kwargs below handles it.

docs/guides/embeddings.md
```diff
@@ -0,0 +1,304 @@
...
+
+```bash
+# Specify embedding model
+--model "BAAI/bge-large-en-v1.5"
+
+# Set encoding format (float or base64)
+--backend-extra '{"encoding_format": "base64"}'
+
+# Limit embedding dimensions
+--backend-extra '{"dimensions": 512}'
```
sjmonson: `--backend-extra` does not exist. The correct approach is `--backend-kwargs '{"extras": {"body": {"dimensions": 512}}}'`

docs/guides/embeddings.md
```diff
@@ -0,0 +1,304 @@
...
+
+## Quick Start
+
+### Basic Embeddings Benchmark
+
+```bash
+guidellm benchmark \
+  --target http://localhost:8000/v1 \
+  --model "BAAI/bge-small-en-v1.5" \
+  --request-format embeddings \
```
sjmonson:
```suggestion
  --request-format /v1/embeddings \
```

Due to legacy naming comment. This is used a lot in this document so make sure to update all of them.

docs/guides/embeddings.md
```diff
@@ -0,0 +1,304 @@
...
+### Benchmark Profiles
+
+Embeddings support all standard benchmark profiles:
+
+```bash
+# Constant rate
+--profile constant --rate 10
+
+# Sweep multiple rates
+--profile sweep --rate 1 5 10 20
```
sjmonson: Invalid syntax for rate. Should be `--rate 1,5,10,20`.

sjmonson (changes requested):
Overall a much better design. See nits below.

docs/guides/embeddings.md
```diff
@@ -0,0 +1,304 @@
...
+### Benchmark Profiles
+
+Embeddings support all standard benchmark profiles:
+
+```bash
+# Constant rate
+--profile constant --rate 10
+
+# Sweep multiple rates
+--profile sweep --rate 1,5,10,20
```
dbutenhof: FYI: Sweep actually uses only *one* "rate" (the first value of a list if one is given) as the number of asynchronous "strategy" steps following the synchronous & throughput determination of the minimum and maximum async rate. (It defaults to 5.)

concurrent, constant/async, and poisson will accept a list of rates or a single rate. throughput requires a single rate (and will ignore additional rates).

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -1119,3 +1120,109 @@ def format(
...
+
+        # Build response (no text output for embeddings)
+        return GenerationResponse(
+            request_id=request.request_id,
+            request_args=arguments.model_dump_json(),
+            text="",  # Embeddings don't generate text
+            input_metrics=UsageMetrics(
+                text_tokens=usage.get("prompt_tokens", 0),
+            ),
+            output_metrics=UsageMetrics(
```
jaredoconnell: Is there any need for any other type of output metric, given that it doesn't output tokens?

jaredoconnell (dismissed):
Looks good to me. Sam will also re-review this.

dbutenhof (dismissed):
I think this looks OK, although I made a comment on some documentation that could be "tweaked" for clarity.

docs/guides/embeddings.md
```diff
@@ -76,20 +77,24 @@ Or use synthetic data generation:
...
+--profile constant --rate 10,20,50  # Run successive constant streams
+
+# Poisson (probabilistic rate distribution, accepts list)
 --profile poisson --rate 10
+
+# Concurrent (fixed parallel streams, accepts list)
+--profile concurrent --rate 10,20,50
+
+# Sweep (runs synchronous, throughput, then interpolated constant strategies)
+--profile sweep --rate 5  # Total number of strategies (default: 5)
```
sjmonson: Default is not 5

```suggestion
# Sweep (runs synchronous, throughput, then interpolated constant strategies)
--profile sweep --rate 5  # Total number of strategies
```

docs/guides/embeddings.md
```diff
@@ -76,20 +77,24 @@ Or use synthetic data generation:
...
-
-# Maximum throughput
---profile throughput
-
 # Synchronous (one request at a time)
 --profile synchronous
 
-# Poisson distribution
+# Throughput (maximum throughput discovery)
+--profile throughput --rate 10  # Number of concurrent request streams (default: 10)
```
sjmonson: Default is not 10

```suggestion
--profile throughput --rate 10  # Number of concurrent request streams
```

sjmonson (changes requested):

jaredoconnell:
One option to consider is getting less detailed with the use of profiles in GuideLLM. Because it's redundant, and the CLI format will change soon.

sjmonson:
> One option to consider is getting less detailed with the use of profiles in GuideLLM. Because it's redundant, and the CLI format will change soon.

Yeah I had considered this to. I think overall the format of the guide needs a little bit of work and more linking to other relevant sections in the docs, but I didn't want to block the whole feature on it.

docs/guides/embeddings.md
dbutenhof: Yeah, that was my mistake -- I added the default as a parenthetical comment (without even bothering to confirm my incorrect memory), and this really isn't an appropriate place to be documenting default values anyway.

docs/guides/embeddings.md
dbutenhof: FYI, there is no default rate for throughput. It's not really necessary to show every possible profile here. You did mention that all standard profiles apply, and an example is fine, but we don't need to be comprehensive in this context.

maryamtahhan: Ok i will simplify the docs here

dbutenhof (approved):
I think a few sections of the expanded documentation were nice (like the sample JSONL dataset, which appears in some other specialized modes) -- but I also don't think any of the documentation you dropped was critical, so let's move ahead.

sjmonson (approved):

src/guidellm/benchmark/outputs/console.py
sjmonson: This was never fixed but its a preexisting but so I won't block this PR on it.

PR #712, opened 2026-04-29
Multi-turn tool call chat completions conversations

### LINKED ISSUE #416: Support benchmarking of reasoning and tool calling Chat Completions requests

jaredoconnell:
## Summary

This PR adds client-side chat completions conversations to the http backend.

## Details

- The design is data-driven. The only data passed to the backend is an API field not included in datasets (required vs auto) and the behavior on how to handle missing tool calls.
- Allows external datasets or synthetic data. If synthetic data, it's a simple JSON result with a field populated with the same generation logic as any other synthetic data in GuideLLM.
- Allows specifying tool calls as `auto` or `required` to the model. Good for testing various scenarios. Models behave differently depending on the value set. `required` is best for predictability.
- Allows specifying how to handle missing tool calls. Useful to set whether it's okay or an error condition, and if it is okay, whether to end the conversation early or continue the conversation.

## Test Plan

- Run the tests
- Follow the documentation to run vLLM with tool calls enabled

## Related Issues

- Resolves #416 

---

## Use of AI


src/guidellm/backends/openai/http.py
```diff
@@ -82,6 +82,26 @@ class OpenAIHttpBackendArgs(BackendArgs):
...
+    @field_validator("tool_call_missing_behavior")
+    @classmethod
+    def validate_tool_call_missing_behavior(cls, v: str) -> str:
+        valid = {"ignore_continue", "ignore_stop", "error_stop"}
+        if v not in valid:
+            raise ValueError(
+                f"Invalid tool_call_missing_behavior '{v}'. "
+                f"Must be one of: {', '.join(sorted(valid))}"
+            )
+        return v
```
sjmonson:
```suggestion
    tool_call_missing_behavior: Literal["ignore_continue", "ignore_stop", "error_stop"] = Field(
        default="error_stop",
        description=(
            "What happens when a tool call is expected but the model does not "
            "produce one. Options: ignore_continue (continue to next turn), "
            "ignore_stop (cancel remaining turns), error_stop (error and "
            "cancel remaining turns)."
        ),
    )
```

Unnecessary `field_validator`. Pydantic has this builtin

src/guidellm/benchmark/outputs/html.py
```diff
@@ -375,6 +375,8 @@ def _build_workload_details(
     sample_outputs = [
         req.output.replace("\n", " ").replace('"', "'")
         if (req := successful_requests[i]).output
+        else json.dumps(req.tool_calls).replace('"', "'")
+        if req.tool_calls
```
sjmonson: Updating this could break HTML. Might as well drop the change since it will all be rewritten soon.

```suggestion
```

src/guidellm/data/preprocessors/mappers.py
```diff
@@ -151,10 +161,12 @@ def datasets_mappings(
...
 
-                # Re-enumerate to ensure we don't have a gap in turns
-                for turn, (_, column_name) in enumerate(sorted(turn_columns)):
+                # Preserve the original turn index from the column name so
+                # that sparse columns (e.g. tools_0, tools_3) stay aligned
+                # with the turns they belong to.
+                for original_turn, column_name in sorted(turn_columns):
                     column_type = cast("GenerativeDatasetColumnType", column_type)
-                    mappings[(column_type, turn)].append((index, column_name))
+                    mappings[(column_type, original_turn)].append((index, column_name))
```
sjmonson: I'm fine with this change since I debated this behavior initially, but remove the comment and rename the var back to `turn` since `original_turn` is meaningless to someone who is not looking at the original code.

src/guidellm/schemas/info.py
```diff
@@ -168,6 +168,15 @@ class RequestInfo(StandardBaseModel):
...
     )
+    stop_conversation: bool = Field(
+        default=False,
+        description=(
+            "When True, the worker cancels all remaining turns in the "
+            "conversation after this request completes successfully. "
+            "Set by the backend when a turn's result means the conversation "
+            "should not continue (e.g. expected tool call not produced)."
+        ),
+    )
```
sjmonson: We already use exceptions to trigger early stops from the backend so I prefer that route. `"error_stop"` can throw any exception and `"ignore_stop"` might work to throw `asyncio.CancelledError` (See future comment for examples).

src/guidellm/scheduler/worker.py
sjmonson, line 1: Don't need these changes with above suggestion.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -460,7 +514,20 @@ def format(  # noqa: C901
...
+        When ``response`` carries ``tool_calls`` (from a prior turn), the
+        assistant message is emitted with the full tool_calls array and
+        synthetic ``role: "tool"`` result messages are appended so the model
+        sees a complete multi-turn tool-calling exchange.  Per-request tool
+        definitions from ``data.columns["tools_column"]`` are injected into
+        the request body.
+
+        On turns where ``data.expects_tool_call`` is ``False``, the
+        ``tool_choice`` API parameter is overridden to ``"none"`` so the model
+        produces a plain text response.
```
sjmonson: This function is not about tool calling and the actual functionality this describes is now handled in other functions.

```suggestion
```

src/guidellm/backends/openai/http.py
```diff
@@ -521,6 +550,38 @@ def _build_headers(
...
+        """
+        if not request.expects_tool_call or response.tool_calls:
+            return
+
+        if self.tool_call_missing_behavior == "ignore_continue":
+            pass
+        elif self.tool_call_missing_behavior == "ignore_stop":
+            request_info.stop_conversation = True
+        elif self.tool_call_missing_behavior == "error_stop":
+            request_info.error = "Expected tool call but model produced none"
```
sjmonson:
```suggestion
            raise ValueError("Expected tool call but model produced none")
```

src/guidellm/backends/openai/http.py
```diff
@@ -521,6 +550,38 @@ def _build_headers(
...
+        :param response: The compiled response from the model.
+        :param request_info: Mutable request metadata; fields may be set.
+        """
+        if not request.expects_tool_call or response.tool_calls:
+            return
+
+        if self.tool_call_missing_behavior == "ignore_continue":
+            pass
+        elif self.tool_call_missing_behavior == "ignore_stop":
+            request_info.stop_conversation = True
```
sjmonson:
```suggestion
            raise asyncio.CancelledError("Expected tool call but model produced none")
```

sjmonson (changes requested):
Not finished reviewing; will add more comments in a bit. Github is acting up and not letting me add to this review for some reason.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -6,11 +6,18 @@
 both streaming and non-streaming responses. Each handler implements the
 GenerationRequestHandler protocol to format json requests, parse API responses,
 extract usage metrics, and convert results into standardized GenerationResponse.
+
+The chat completions handler supports multi-turn tool calling with pre-planned
+turns: it captures ``tool_calls`` from model responses, reconstructs them from
+streaming deltas, and formats follow-up messages with synthetic tool results.
+The ``expects_tool_call`` flag on each request controls per-turn
+``tool_choice`` overrides.
```
sjmonson:
```suggestion
```

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -450,7 +504,96 @@ def _format_prompts(
...
+        # break the server's constrained decoding grammar.
+        body.pop("ignore_eos", None)
+        body.pop("stop", None)
+
+        # On tool-call turns, let the model finish valid JSON naturally;
+        # max_completion_tokens would truncate output mid-JSON and corrupt
+        # the arguments sent in conversation history on follow-up turns.
+        if data.expects_tool_call:
+            body.pop("max_completion_tokens", None)
+            body.pop("max_tokens", None)
```
sjmonson, line 549: This does not seem right. Either all of these key pops belong under `if data.expects_tool_call` or none of them do. If you pop `ignore_eos` but not `max_tokens` then you end up with the same problem as is mentioned in the first comment.

sjmonson: Its also a very bad idea to remove these keys unless we absolutely have to because it could destroy benchmark reproducibility. So if you haven't already I would suggest you double check the behavior.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -545,9 +719,13 @@ def compile_non_streaming(
         Process a complete chat completion response.
 
         Extracts content from the message object within choices, handling the nested
-        structure specific to chat completion endpoints.
+        structure specific to chat completion endpoints.  When the model responds
+        with ``tool_calls``, they are captured on the returned
+        :class:`GenerationResponse` so the scheduler can detect them and
+        dynamically create a follow-up turn.
```
sjmonson:
```suggestion
        structure specific to chat completion endpoints.
```

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -22,15 +29,57 @@
...
+    def to_dict(self) -> dict[str, Any]:
+        """Serialize to the OpenAI tool_call dict format."""
+        return {
+            "id": self.id,
+            "type": self.type,
+            "function": {
+                "name": self.function.name,
+                "arguments": self.function.arguments,
+            },
+        }
```
sjmonson: Pydantic much better here; especially if this gets serialized.

```suggestion
class StreamingToolCallFunction(BaseModel):
    """Accumulated function name and arguments for a single streamed tool call."""

    name: str = ""
    arguments: str = ""


class StreamingToolCall(BaseModel):
    """A single tool call reassembled from streaming deltas. """

    id: str = ""
    type: str = "function"
    function: StreamingToolCallFunction = Field(
        default_factory=StreamingToolCallFunction
    )
```

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -634,11 +841,21 @@ def compile_streaming(
             else:  # mixed content + tool call turn
                 output_metrics.mixed_content_tool_tokens = output_metrics.text_tokens
 
+        # Assemble full tool call payloads from the streaming deltas so they
+        # can be stored on GenerationResponse for multi-turn history.
+        raw_tool_calls: list[dict[str, Any]] | None = None
+        if self.streaming_tool_calls:
+            raw_tool_calls = [
+                self.streaming_tool_calls[i].to_dict()
```
sjmonson: Due to Pydantic change above.

```suggestion
                self.streaming_tool_calls[i].model_dump()
```

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -22,15 +29,57 @@
...
     "PoolingRequestHandler",
     "ResponsesRequestHandler",
+    "StreamingToolCall",
+    "StreamingToolCallFunction",
     "TextCompletionsRequestHandler",
 ]
 
+# Placeholder content returned in synthetic ``role: "tool"`` messages during
+# multi-turn tool calling.  The model sees this as the tool's output.
+DEFAULT_SYNTHETIC_TOOL_RESPONSE = '{"status": "ok"}'
```
sjmonson: Add this to config in `settings.py`.

src/guidellm/data/deserializers/synthetic.py
```diff
@@ -18,13 +18,36 @@
...
 
 __all__ = [
+    "DEFAULT_SYNTHETIC_TOOLS",
+    "DEFAULT_SYNTHETIC_TOOL_RESPONSE",
     "SyntheticTextDataset",
     "SyntheticTextDatasetDeserializer",
 ]
 
+# Short placeholder used when no tool_response_tokens size is configured.
+DEFAULT_SYNTHETIC_TOOL_RESPONSE = '{"status": "ok"}'
```
sjmonson: Reuse value from `settings.py`.

src/guidellm/data/deserializers/synthetic.py
```diff
@@ -18,13 +18,36 @@
     DatasetDeserializerFactory,
 )
 from guidellm.data.schemas import SyntheticTextDatasetConfig
+from guidellm.utils.imports import json
 from guidellm.utils.random import IntegerRangeSampler
 
 __all__ = [
+    "DEFAULT_SYNTHETIC_TOOLS",
+    "DEFAULT_SYNTHETIC_TOOL_RESPONSE",
```
sjmonson:
```suggestion
```

Don't export these.

src/guidellm/data/deserializers/synthetic.py
sjmonson, line 1: Fine for now, but in the future if we still want synthetic tool-calling it should be a separate dataset.

src/guidellm/data/schemas.py
```diff
@@ -159,6 +196,28 @@ class SyntheticTextDatasetConfig(DataConfig):
...
+            )
+
+        if self.tool_response_tokens is not None and self.tool_call_turns == 0:
+            raise ValueError(
+                "tool_response_tokens was set but tool_call_turns is 0. "
+                "Set tool_call_turns > 0 to enable tool calling."
+            )
+
+        return self
+
```
sjmonson:
```suggestion
```

src/guidellm/backends/openai/request_handlers.py
sjmonson: Additionally, move these into `schemas/request.py` since it is functally used in `GenerativeRequestStats`.

src/guidellm/schemas/request_stats.py
```diff
@@ -53,6 +53,10 @@ class GenerativeRequestStats(StandardBaseDict):
     output: str | None = Field(
         default=None, description="Generated text output from the request"
     )
+    tool_calls: list[dict[str, Any]] | None = Field(
+        default=None,
+        description="Raw tool call payloads from the model response in OpenAI format",
+    )
```
sjmonson:
```suggestion
    tool_calls: list[StreamingToolCall] | None = Field(
        default=None,
        description="Raw tool call payloads from the model response in OpenAI format",
    )
```

src/guidellm/schemas/response.py
```diff
@@ -24,8 +26,12 @@ class GenerationResponse(StandardBaseModel):
...
 
     Captures the output and metrics from a generation request, providing structured
-    data for text output, token usage statistics, and compilation of detailed
-    request statistics for analysis and monitoring purposes.
+    data for text output, tool call payloads, token usage statistics, and compilation
+    of detailed request statistics for analysis and monitoring purposes.
+
+    When the model responds with tool calls instead of (or in addition to) plain
+    text, the ``tool_calls`` field carries the raw payloads so the scheduler can
+    dynamically create follow-up turns for multi-turn tool calling benchmarks.
```
sjmonson:
```suggestion
```

src/guidellm/schemas/response.py
```diff
@@ -52,6 +58,16 @@ class GenerationResponse(StandardBaseModel):
...
+    tool_calls: list[dict[str, Any]] | None = Field(
+        default=None,
+        description=(
+            "Raw tool call payloads from the model response, each containing "
+            "id, type, and function (name + arguments) in OpenAI format. "
+            "Populated for both streaming and non-streaming chat completions. "
+            "Used by the scheduler to detect tool call turns and dynamically "
+            "create follow-up requests in multi-turn conversations."
+        ),
+    )
```
sjmonson:
```suggestion
    tool_calls: list[StreamingToolCall] | None = Field(
        default=None,
        description=(
            "Raw tool call payloads from the model response, each containing "
            "id, type, and function (name + arguments) in OpenAI format."
        ),
    )
```

src/guidellm/__main__.py
```diff
@@ -408,6 +408,29 @@ def benchmark():
...
+    "--tool-call-missing-behavior",
+    type=click.Choice(["ignore_continue", "ignore_stop", "error_stop"]),
+    default=None,
+    help=(
+        "What the worker does when a tool call is expected but the model "
+        "does not produce one. ignore_continue: continue to next turn, "
+        "ignore_stop: cancel remaining turns, error_stop: error and "
+        "cancel remaining turns. Default: error_stop."
+    ),
+)
```
sjmonson: No new top-level args. Handle this in `--backend-kwargs`

src/guidellm/__main__.py
```diff
@@ -429,6 +452,24 @@ def run(**kwargs):  # noqa: C901
...
+    tool_call_missing_behavior = kwargs.pop("tool_call_missing_behavior", None)
+
+    if tool_choice is not None:
+        extras = backend_kwargs.setdefault("extras", {})
+        body = extras.setdefault("body", {})
+        body["tool_choice"] = tool_choice
+
+    if tool_call_missing_behavior is not None:
+        backend_kwargs["tool_call_missing_behavior"] = tool_call_missing_behavior
+
```
sjmonson:
```suggestion
```

docs/guides/multiturn.md
sjmonson, line 1: This should be its own guide, not inserted into the generic Multiturn one.

sjmonson (changes requested):

docs/guides/multiturn.md
jaredoconnell: Moved.

src/guidellm/data/deserializers/synthetic.py
jaredoconnell: Done.

src/guidellm/scheduler/worker.py
```diff
@@ -434,6 +449,12 @@ async def _process_next_request(  # noqa: C901
...
                     f"Backend exception for request {request_info.request_id}"
                 )
+                # Cancel remaining conversation turns on backend error
+                for skip_req, skip_info in conversation:
+                    skip_info.error = f"Cancelled: {request_info.error}"
+                    skip_info.timings.resolve_end = time.time()
+                    self._send_update("cancelled", None, skip_req, skip_info)
+                conversation.clear()
         finally:
             if request_info is not None:
```
jaredoconnell, line 439: What do you think of this design?

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: I think it was designed to unconditionally pop some of those when tools were present was in case the model created some anyway. For now I moved them all to the conditional, and added a section to the documentation on why this happens.

dbutenhof (commented):
Finally got through this. Looks good. Needs a rebase.

src/guidellm/schemas/info.py
jaredoconnell: This is removed now.

src/guidellm/data/preprocessors/mappers.py
jaredoconnell: Changed back.

sjmonson:
augment review

src/guidellm/scheduler/worker.py
sjmonson: `request_info.error` is designed to be set by this function; not the backend. Also this is duplicate, same thing happens in the `finally` block below.

src/guidellm/scheduler/worker.py
sjmonson: Comment stands: `git checkout main -- src/guidellm/scheduler/worker.py`

src/guidellm/data/preprocessors/mappers.py
```diff
@@ -168,8 +178,10 @@ def datasets_mappings(
                     dataset_columns_str,
                 )
 
-                # Re-enumerate to ensure we don't have a gap in turns
-                for turn, (_, column_name) in enumerate(sorted(turn_columns)):
+                # Preserve the original turn index from the column name so
+                # that sparse columns (e.g. tools_0, tools_3) stay aligned
+                # with the turns they belong to.
```
sjmonson:
```suggestion
```

We are not "preserving" anything; its just a lack of change.

src/guidellm/backends/openai/http.py
```diff
@@ -480,7 +498,9 @@ async def resolve(  # type: ignore[override, misc]
                     request_info.timings.token_iterations += iterations
 
             request_info.timings.request_end = time.time()
-            yield request_handler.compile_streaming(request, arguments), request_info
+            gen_response = request_handler.compile_streaming(request, arguments)
+            yield gen_response, request_info
+            self._check_tool_call_expectations(request, gen_response)
```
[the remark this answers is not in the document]
sjmonson: This is correct, should be:

1. gen_response
2. _check_tool_call_expectations
3. yield

since an error will kick out to `except` block and yield.

sjmonson (changes requested):

src/guidellm/backends/openai/http.py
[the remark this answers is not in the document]
jaredoconnell: Should be addressed now.

docs/guides/tool_calling.md
```diff
@@ -0,0 +1,168 @@
...
+
+Plain-text turns (where `tool_choice="none"`) are unaffected and continue to respect `output_tokens` / `ignore_eos` as normal, even when tool definitions are present in the request.
+
+## Edge Cases
+
+- **Single-turn tool calling** (`turns=1, tool_call_turns=1` or `tool_call_turns=[0]`) is supported. The conversation has one turn that expects a tool call and no plain-text response.
+- **All-tool conversations** (e.g. `tool_call_turns=3` with `turns=3`, or `tool_call_turns=[0,1,2]`) are supported. Every turn is a tool-call turn and the model never produces a final plain-text response. The `output` field in `benchmarks.json` will be `None` for every request; use the `tool_calls` field to inspect model output.
+- **Non-contiguous tool turns** (e.g. `tool_call_turns=[0, 2]` with `turns=4`) are supported. Only the specified turns expect tool calls; other turns produce plain text.
+- **Tool definitions on non-tool turns** are still sent in the request (they're part of the data), but `tool_choice` is forced to `none` so the model produces text. This matches real-world agentic patterns where the tools remain available but the model is instructed to respond in natural language.
```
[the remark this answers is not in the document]
jaredoconnell: Addressed.

src/guidellm/scheduler/worker.py
```diff
@@ -426,6 +426,7 @@ async def _process_next_request(  # noqa: C901
                 self._send_update("cancelled", response, request, request_info)
             raise
         except Exception as exc:  # noqa: BLE001
+            premature_exit = True
```
jaredoconnell, line 429: This is still required. Is there any reason why the rest of a conversation should not be cancelled if a request fails? If it's not cancelled, that seems like bad behavior, because the conversation would be missing pieces.

sjmonson: Yeah you are right. This is definitely a mistake in the original code.

sjmonson:
@mergifyio rebase

jaredoconnell:
I will rebase locally. I enabled `rerere` in anticipation of this.

dbutenhof (dismissed):
The code looks good; I ran it against my server with Qwen/Qwen3-0.6B and the GuideLLM sample command from `tool_calling.md`, and it seems to run OK:

```console
ℹ Tool Call Metrics Statistics (Completed Requests)
|===========|======|=======|=======|=======|======|=======|======|======|
| Benchmark | Output Tokens             |||| Output Count            ||||
| Strategy  | Per Request || Per Second   || Per Request || Per Second ||
|           | Mdn  | p95   | Mdn   | Mean  | Mdn  | p95   | Mdn  | Mean |
|-----------|------|-------|-------|-------|------|-------|------|------|
| constant  | 77.0 | 692.0 | 152.7 | 309.4 | 1.0  | 19.0  | 1.5  | 4.8  |
|===========|======|=======|=======|=======|======|=======|======|======|
```

dbutenhof (dismissed):

sjmonson (dismissed):

dbutenhof (dismissed):
This should be safe given that `stream` is a `Request` object from the original connection that should only change if another call is made.

dbutenhof (approved):
Still looks good

dbutenhof:
🥳 🎆 🧰

PR #715, opened 2026-05-04
Refactor CSV tests

jaredoconnell:
## Summary

It now isolates specific functionality better, and covers more edge cases.

## Details

- Moves functionality into `_align_columns`, which can be more easily tested in isolation.
- Adds unit tests that cover edge cases
- Has an integration test.

## Test Plan

- I ran a real benchmark with CSV to confirm that it still works.
- You can run the unit tests.

---

## Use of AI

### DIFF
--- a/tests/unit/benchmark/test_csv_output.py
+++ b/tests/unit/benchmark/test_csv_output.py
@@ -163,7 +163,7 @@ async def test_finalize_aligns_columns_in_written_csv(tmp_path: Path):
     Integration test: finalize writes a CSV where all rows (headers + data)
     have the same column count, even when benchmarks produce different columns.
 
-    Uses monkeypatching to control the column shape without constructing full
+    Uses patching to control the column shape without constructing full
     benchmark objects.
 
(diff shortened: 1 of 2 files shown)


tests/unit/benchmark/test_csv_output.py
```diff
@@ -7,21 +7,181 @@
...
+
+
+@pytest.mark.asyncio
+@pytest.mark.sanity
+async def test_finalize_aligns_columns_in_written_csv(tmp_path: Path):
+    """
+    Integration test: finalize writes a CSV where all rows (headers + data)
+    have the same column count, even when benchmarks produce different columns.
+
+    Uses monkeypatching to control the column shape without constructing full
```
dbutenhof: Technically, you're stubbing method attributes directly, not "monkeypatching", and it's kinda interesting that the LLM would have generated this alias ...

dbutenhof (dismissed):
I like the cleaner set of test cases...

dbutenhof (approved):

PR #717, opened 2026-05-04
Refactor CLI into nested structure

sjmonson:
## Summary

Migrates CLI code out of `__main__.py` and into its own submodule.

## Details

The PR is purely refactoring. Making this change now will make the incoming CLI refactor a little cleaner to review.

## Test Plan

Run any existing CLI commands to confirm functionality.

---

## Use of AI

### DIFF
--- a/src/guidellm/__main__.py
+++ b/src/guidellm/__main__.py
@@ -1,0 +1,6 @@
+"""GuideLLM command-line interface entry point"""
+
+from guidellm.cli import cli
+
+if __name__ == "__main__":
+    cli()
--- a/src/guidellm/cli/__init__.py
+++ b/src/guidellm/cli/__init__.py
@@ -1,0 +1,46 @@
+"""
+GuideLLM command-line interface entry point.
+
+Primary CLI application providing benchmark execution, dataset preprocessing, and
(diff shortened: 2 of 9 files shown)


src/guidellm/__main__.py
```diff
@@ -1,784 +1,6 @@
...
-
-    server = MockServer(config)
-    console = Console()
-    console.print_update(
-        title="GuideLLM mock server starting...",
-        details=f"Listening on http://{host}:{port} for model {model}",
-        status="success",
-    )
-    server.run()
+"""Entry points for guidellm."""
```
dbutenhof: OK, I would have stuck with the first line of the old comment: "GuideLLM command-line interface entry point".

dbutenhof (approved):
It's hard to review this sort of refactoring without simply trusting that it's mostly a cut-and-paste; but that appears to be essentially the case, (and so I've avoided making some random commentary along the way that would equally apply to the current code) ... so I've really just got one minor comment comment...

dbutenhof (dismissed):

src/guidellm/__main__.py
sjmonson: Done

dbutenhof (approved):

PR #718, opened 2026-05-05
Add instructions against common AI poor code quality habits

jaredoconnell:
## Summary

AI bots very frequently tend to import things inline and use `getattr`. But we have always found that to be a bad design, so this documents AI chatbots not to do these things.

## Details

- Inline imports are not ideal from a code style and maintainability standpoint 
- getattr is frequently used by AI to speculatively use values that don't exist on a type, reducing quality of the code.

---

## Use of AI

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -72,8 +72,8 @@ tox -e tests -- -m regression
 - All tests must pass before committing
 - Markdown files must be properly formatted
 - Public functions in `src/` code must use the reStructuredText docstring format
-- All imports must be done at the top of the file unless necessary for functionality
-- Use of `getattr` should be avoided if possible as it hides incorrect usage of types
+- All imports **SHALL** be done at the top of the file
+- **DO NOT** use `getattr` or `setattr` as it hides incorrect usage of types
 
 ## Common Tasks
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -65,15 +65,23 @@ tox -e tests -- -m regression
 - Use appropriate markers (`smoke`, `sanity`, `regression`)
 - Tests should be placed in files matching the name and path of the file under tests. E.g. `src/guidellm/benchmark/schemas/generative/entrypoints.py` -> `tests/unit/benchmark/schemas/generative/test_entrypoints.py`.
 
-### Style Requirements
+### Quality Requirements
 
 - All Python code must pass linting and formatting
 - All Python code must pass type checking
 - All tests must pass before committing
 - Markdown files must be properly formatted
+
+### Style Requirements
+
 - Public functions in `src/` code must use the reStructuredText docstring format
 - All imports **SHALL** be done at the top of the file
 - **DO NOT** use `getattr` or `setattr` as it hides incorrect usage of types
+
+### Design Requirements
+
+- Only touch sections of code that need to be changed for the given task
+- Prefer solutions that keep non-generic code out of generic pathways
 
 ## Common Tasks
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -81,7 +81,8 @@ tox -e tests -- -m regression
 ### Design Requirements
 
 - Only touch sections of code that need to be changed for the given task
-- Prefer solutions that keep non-generic code out of generic pathways
+- When handling variant-specific logic, encapsulate it in methods on registry class implementations rather than adding if/else branches to generic code paths
+- Class implementations must fully encapsulate their unique logic and that logic must not leak into caller code paths.
 
 ## Common Tasks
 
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -72,6 +72,8 @@ tox -e tests -- -m regression
 - All tests must pass before committing
 - Markdown files must be properly formatted
 - Public functions in `src/` code must use the reStructuredText docstring format
+- All imports must be done at the top of the file unless necessary for functionality
```
dbutenhof: Can we be more specific? I think the only time we legitimately do that is to narrow the scope of optional extra packages; so maybe even go very specific like "unless necessary to minimize the scope of optional imports listed under the `pyproject.toml` "project.optional-dependencies" section."

AGENTS.md
```diff
@@ -72,6 +72,8 @@ tox -e tests -- -m regression
...
+- Use of `getattr` should be avoided if possible as it hides incorrect usage of types
```
dbutenhof: We use `getattr` quite a bit -- again, I wonder if we can be more specific here. I kinda worry about how an over-eager-to-please LLM might choose to interpret "if possible". I'm not really coming up with a better condition right now, but this feels like a plot hole ... 😆

dbutenhof (changes requested):
I like the idea of accumulating protections and advice as we go, so this is a fantastic precedent. We need to be cautious about the wording, so I'm expressing some concerns.

AGENTS.md
sjmonson: I would make this more broad, not more specific. If you give AI an excuse to ignore a rule it will twist its logic to match.

```suggestion
- All imports **SHALL** be done at the top of the file
```

AGENTS.md
sjmonson: This is general guidance for AIs, not humans. If a human wants to override this guidance for a good reason they are free to do so.


```suggestion
- **DO NOT** use `getattr` or `setattr` as it hides incorrect usage of types
```

sjmonson (changes requested):
Please restructure the sections a bit as we will probably add more guidance in the future.

```md
### Quality Requirements

- All Python code must pass linting and formatting
- All Python code must pass type checking
- All tests must pass before committing
- Markdown files must be properly formatted

### Style Requirements

- Public functions in `src/` code must use the reStructuredText docstring format
- All imports must be done at the top of the file unless necessary for functionality
- Use of `getattr` should be avoided if possible as it hides incorrect usage of types
```

Also see wording suggestions below.

AGENTS.md
dbutenhof: My concern is that an over-eager AI assistant will notice & decide to "fix" existing instances. Ah well -- we can burn that bridge when we reach it.

dbutenhof (commented):
I've been playing with my GitHub PM tool, and I added an `AGENTS.md` to be sure it checked & fixed all formatting & tests before "finishing". I found it's lazy about keeping the rules in context. It suggested adding a section at the beginning, and that seems to be working (so far), so we should consider something like this:

```markdown
## Read this file when starting or resuming work

- **Open `AGENTS.md` again** when you begin a task on this repo or return after a long gap, so required checks (below) stay in context until all required checks are green.
```

AGENTS.md
jaredoconnell: I added a design requirements section that should address the over-eager changes.

dbutenhof (dismissed):
Sure; I commented earlier about a measure that might help to convince the agent to keep these rules in active context (from my experience with Cursor), but we can always just see what happens...

AGENTS.md
```diff
@@ -65,13 +65,23 @@ tox -e tests -- -m regression
...
+### Style Requirements
+
 - Public functions in `src/` code must use the reStructuredText docstring format
+- All imports **SHALL** be done at the top of the file
+- **DO NOT** use `getattr` or `setattr` as it hides incorrect usage of types
+
+### Design Requirements
+
+- Only touch sections of code that need to be changed for the given task
+- Prefer solutions that keep non-generic code out of generic pathways
```
sjmonson: What does this mean? I am not entirely sure what this entails so idk if an AI would figure it out.

sjmonson (changes requested):

AGENTS.md
jaredoconnell: How about this?

> Do not add if/else branches for variant-specific logic in generic code paths. Variant-specific logic belongs in methods on registry class implementations

Or
> When handling variant-specific logic, encapsulate it in methods on registry class implementations rather than adding if/else branches to generic code paths

dbutenhof: Or -- class implementations must fully encapsulate their unique logic and that logic must not leak into caller code paths.

sjmonson: I like @jaredoconnell's option 2 and @dbutenhof's. Maybe add both?

sjmonson (approved):
I am a little worried that the wording for points 2 and 3 in "Design Requirements" it too unique for it to really be effective; LLMs do much better with common phrases, but we can give it a shot.

dbutenhof (approved):

PR #723, opened 2026-05-07
[v0.7 CLI Refactor] Rework BackendArgs to be the authoritative config location 

### LINKED ISSUE #725: Rework interface between `backends` and `benchmark`
### Problem Statement

Currently the `benchmark` subsystem handles validating backend configuration and creating a backend with given parameters. The means every change to the way a backend is configured takes place in both `backends` and `benchmark` and limits the ability to extend backends.

### Proposed Solution

Handle backend configuration in an interface expose from each backed to the benchmark subsystem. The existing `BackendArgs` data class can be reused for this interface.

### Alternatives Considered

_No response_

### Usage Examples

```markdown

```

### Additional Context

_No response_

sjmonson:
## Summary

Reworks `BackendArgs` into a `PydanticClassRegistry` containing everything needed to create a backend.

## Details

This is the first (or third depending on how you count) patch to refactor our CLI and the internal way submodules are configured and spawned. Since we already had a partial base for it this will likely be the simplest PR and subsequent PRs will have to rework a lot more code.

Changes to the entrypoints and CLI are temporary as that code will be refactored in a follow-up.

### Changes

I tried to keep most of the original functionality but there were a few things that just did not make sense as implemented or were planned for removal:

#### Dropped `LEGACY_API_ALIASES`

They only exist to keep existing scripts from breaking and this release those scripts will break for other reasons so might as well make the change now.

#### Dropped `request_handlers` backend argument

This let the user provide custom request type handlers as a backend argument. The better way to do this is to just call `OpenAIRequestHandlerFactory.register` on your custom handler.

#### `validate_backend` no longer has the option to provide a custom health check endpoint

You can get nearly the same functionality by passing `api_routes={"/health": "custom endpoint"}`. However users usually just set `validate_backend=False` since its too much of a hassle to provide a custom endpoint.

## Test Plan

Test setting `--backend` and `--backend-kwargs` with various values to make sure nothing has changed (except for exceptions detailed below).

## Related Issues

- Related to #724
- Resolves #725

---

## Use of AI

### DIFF
--- a/src/guidellm/backends/backend.py
+++ b/src/guidellm/backends/backend.py
@@ -9,7 +9,7 @@ provide a standard interface for distributed execution across worker processes.
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import ClassVar, Literal
+from typing import ClassVar
 
 from pydantic import ConfigDict, Field
 
@@ -24,11 +24,7 @@ from guidellm.utils.registry import RegistryMixin
 __all__ = [
     "Backend",
     "BackendArgs",
-    "BackendType",
 ]
-
-
-BackendType = Literal["openai_http", "vllm_python"]
 
 
 class BackendArgs(PydanticClassRegistryMixin["BackendArgs"], ABC):
@@ -63,7 +59,7 @@ class BackendArgs(PydanticClassRegistryMixin["BackendArgs"], ABC):
 
         return BackendArgs
 
-    type_: BackendType = Field(
+    type_: str = Field(
         alias="type",
         description="Type identifier for the backend configuration.",
     )
--- a/src/guidellm/backends/backend.py
+++ b/src/guidellm/backends/backend.py
@@ -88,14 +88,15 @@ class Backend(
     ::
         @Backend.register("my_backend")
         class MyBackend(Backend):
-            def __init__(self, api_key: str):
-                super().__init__("my_backend")
-                self.api_key = api_key
+            def __init__(self, args: MyBackendArgs):
+                super().__init__(args)
+                self.api_key = args.api_key
 
             async def process_startup(self):
                 self.client = MyAPIClient(self.api_key)
 
-        backend = Backend.create("my_backend", api_key="secret")
+        args = MyBackendArgs(api_key="secret")
+        backend = Backend.create(args)
     """
 
     @classmethod
--- a/src/guidellm/backends/backend.py
+++ b/src/guidellm/backends/backend.py
@@ -8,13 +8,17 @@ provide a standard interface for distributed execution across worker processes.
 
 from __future__ import annotations
 
-from abc import abstractmethod
-from typing import Literal
+from abc import ABC, abstractmethod
+from typing import ClassVar, Literal
 
-from pydantic import BaseModel, ConfigDict
+from pydantic import ConfigDict, Field
 
 from guidellm.scheduler import BackendInterface
-from guidellm.schemas import GenerationRequest, GenerationResponse
+from guidellm.schemas import (
+    GenerationRequest,
+    GenerationResponse,
+    PydanticClassRegistryMixin,
+)
 from guidellm.utils.registry import RegistryMixin
 
 __all__ = [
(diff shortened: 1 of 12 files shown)


sjmonson:
*sigh* Fix one CI job and break another. I will have a fix up for the tests soon.

sjmonson:
augment review

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -59,54 +59,68 @@
...
 
+@BackendArgs.register("vllm_python")
 class VLLMPythonBackendArgs(BackendArgs):
     """Pydantic model for VLLM Python backend creation arguments."""
 
+    type_: Literal["vllm_python"] = Field(
+        alias="type",
+        default="vllm_python",
+        description="Backend type identifier for VLLM Python backend.",
+    )
```
jaredoconnell, line 70: Technically, using a type override here is a type violation.

Ideally you could switch this to a `ClassVar`.

This pattern is used in lots of places, but it's technically wrong in those places, too.

src/guidellm/cli/benchmark/run.py
```diff
@@ -386,6 +381,8 @@ def run(**kwargs):  # noqa: C901
 
     # Map top-level CLI options to backend_kwargs
     backend_kwargs = kwargs.pop("backend_kwargs", {})
```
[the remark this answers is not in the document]
sjmonson: Wrong because line 369 does `kwargs = cli_tools.set_if_not_default(ctx, **kwargs)` but easy mistake to make.

src/guidellm/backends/backend.py
```diff
@@ -8,30 +8,61 @@
...
+        if cls.__name__ == "BackendArgs":
+            return cls
 
-class BackendArgs(BaseModel):
-    """Base class for backend creation argument models."""
+        return BackendArgs
 
-    # Allow for extra fields until we make BackendArgs the sole source of truth
-    model_config = ConfigDict(extra="allow")
+    type_: str = Field(
```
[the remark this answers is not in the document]
sjmonson: Good catch. Needs `serialize_by_alias=True` in model config.

src/guidellm/backends/openai/http.py
```diff
@@ -30,50 +30,101 @@
...
         description="Model identifier for generation requests",
-        json_schema_extra={
-            "error_message": (
-                "Backend '{backend_type}' requires a model parameter. "
-                "Please provide --model with a valid model identifier."
-            )
-        },
     )
-    request_format: str | None = Field(
+    request_format: Literal[
```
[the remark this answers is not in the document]
sjmonson: Good catch

src/guidellm/backends/backend.py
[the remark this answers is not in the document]
dbutenhof: Fascinating. Wanna try a unit test to be sure that the round-trip serialization works?

src/guidellm/backends/openai/http.py
```diff
@@ -30,50 +30,102 @@
...
-            "error_message": (
-                "Backend '{backend_type}' requires a target parameter. "
-                "Please provide --target with a valid endpoint URL."
-            )
-        },
     )
-    model: str | None = Field(
-        default=None,
+    model: str = Field(
+        default_factory=str,
```
jaredoconnell, line 65: Why is the None option removed?

src/guidellm/backends/openai/http.py
```diff
@@ -30,50 +30,102 @@
...
+
+@BackendArgs.register("openai_http")
+class OpenAIHTTPBackendArgs(BackendArgs):
     """Pydantic model for OpenAI HTTP backend creation arguments."""
 
+    type_: Literal["openai_http"] = Field(
+        alias="type",
+        default="openai_http",
+        description="Type identifier for the backend configuration.",
+    )
```
dbutenhof, line 60: Such an odd construct -- a type that's a literal of a single value with that single value as the default. Yet it certainly is true that the only valid value for the inherited field in *this* class is that one. No matter how many times I scan past this it just looks weird. 😆

dbutenhof (dismissed):
I pulled it and tried various good and bad option values e.g. for the backend args, and all seems to work, which is one definition of "good".

src/guidellm/backends/backend.py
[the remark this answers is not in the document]
sjmonson: Probably smart. I did manually confirm but a regression test is a good idea.

sjmonson:
Forgot I was going to use this chance to move `GenerationRequestArguments` to the OpenAI backend. Will add that next week.

jaredoconnell (approved):
Looks good to me. I tested basic use cases with the HTTP and vLLM-Python backend.

tests/unit/backends/test_backend.py
```diff
@@ -29,6 +29,175 @@ class _TestBackendArgs(BackendArgs):
...
 
 
+class TestBackendArgs:
+    """Test cases for BackendArgs base class."""
+
+    @pytest.mark.smoke
+    def test_class_signatures(self):
+        """Verify BackendArgs inheritance, discriminator, and available methods.
+
+        ### WRITTEN BY AI ###
```
dbutenhof, line 39: I'd love to transition to the more standard Generated-by. Want to try it here? 🙃

dbutenhof (approved):

src/guidellm/backends/vllm_python/vllm.py
sjmonson: This is how [type discriminators](https://pydantic.dev/docs/validation/2.0/usage/types/unions/#discriminated-unions-aka-tagged-unions) work in Pydantic. It cannot be a ClassVar.

tests/unit/backends/test_backend.py
sjmonson: If we really want to go this path, lets do it in a separate PR and just convert everything at once. Our `AGENTS.md` still recommends the `## WRITTEN BY AI ##` method.

src/guidellm/backends/openai/http.py
sjmonson: Because internally in the `OpenAIHTTPBackend` class `model=None` became `model=""` in `__init__` since it avoids having to check if model is a string in a lot of places.

jaredoconnell:
This looks ready to merge. I have no further comments after seeing your responses.

sjmonson:
> Forgot I was going to use this chance to move `GenerationRequestArguments` to the OpenAI backend. Will add that next week.

Changed my mind about this. I'll do it in a follow-up after #713 lands since that PR adds a common file which will make it easier to move.

PR #733, opened 2026-05-19
[v0.7 CLI Refactor] Rework Data Deserialization Config

sjmonson:
## Summary

First set of changes to `data` creation and configuration. Most changes are hidden interally but does break `--data` usage by requiring the new format (e.g. `kind=sythetic_text,prompt_tokens=32,...`).

## Details

This is missing some key pieces like how `random_seed` and `processor` will be configured. Also currently everything is configured in the dataloader, that may move to some nice helper function under `src/data/entrypoints.py`.

Updated docs changes are generated and mostly untested since they we need to be adjusted again after the full CLI refactor.

## Test Plan

Run any existing workloads. Everything should be the same except for the format of `--data`.

## Related Issues

- Related To #

---

## Use of AI


docs/guides/datasets.md
```diff
@@ -8,17 +8,12 @@ GuideLLM supports various dataset configurations to enable benchmarking and eval
...
+  - `kind=synthetic_text` — generates synthetic prompts on the fly. Required fields: `prompt_tokens`, `output_tokens`. Optional: `turns`, `prefix_tokens`, `prefix_count`, `prefix_buckets`, and distribution controls (`prompt_tokens_stdev`, `output_tokens_stdev`, etc.).
+  - `kind=huggingface` — loads from HuggingFace Hub or a local directory/file. Required field: `data` (dataset ID or path). Pass dataset loading arguments (e.g. `split`, `name`) via the `load_kwargs` field.
+  - `kind=json_file`, `kind=csv_file`, `kind=text_file`, `kind=parquet_file`, `kind=arrow_file`, `kind=hdf5_file` — loads from a local file. Required field: `path`.
+  - `kind=trace_synthetic` — loads a JSONL trace file for replay benchmarking. Required field: `path`. Optional: `timestamp_column` (default: `timestamp`), `prompt_tokens_column` (default: `input_length`), `output_tokens_column` (default: `output_length`).
+  - Can be specified as a key=value string (`kind=synthetic_text,prompt_tokens=256,output_tokens=128`), a JSON string (`'{"kind": "huggingface", "data": "my/dataset", "load_kwargs": {"split": "train"}}'`), or repeated for multiple sources.
```
dbutenhof: This feels out of place in a list of "kind"s. Arguably it's a sentence in the `--data` first-level list item, since you can't really put it "alone" in the context of a list.

dbutenhof (commented):
Wow. Lots of changes, but a few repeating patterns. I think the new patterns are great, and the "tragic" loss of `_deserialize_with_registered_deserializers` made me want to cheer.

I should probably pull this and experiment a bit before hitting "Approve".

dbutenhof (commented):
Interesting. Previously, if I included an unrecognized data option or excluded a "required" like `output_tokens`, the test run just went ahead without complaint.

With the new code, in either case, I get unhelpful errors. An unrecognized (e.g., misspelled) parameter gets "Error: Invalid value for --data: Extra inputs are not permitted", while a missing expected parameter (e.g., our friend "output_tokens") fails with "Error: Invalid value for --data: Field required".

docs/guides/datasets.md
sjmonson: Yeah... most of the docs changes are the AI equivalent of find and replace. Do you want to just note them now and fix them in the PR that does the full CLI change or just address them now?

dbutenhof: The basic information is useful and correct, and I guess if we can get this massive restructuring through more quickly, "nits" like this and the error behavior on DataArgs deserialization can be deferred to the next pass.

sjmonson:
> Interesting. Previously, if I included an unrecognized data option or excluded a "required" like `output_tokens`, the test run just went ahead without complaint.
> 
> With the new code, in either case, I get unhelpful errors. An unrecognized (e.g., misspelled) parameter gets "Error: Invalid value for --data: Extra inputs are not permitted", while a missing expected parameter (e.g., our friend "output_tokens") fails with "Error: Invalid value for --data: Field required".

Yeah the validation error is translated here:

https://github.com/vllm-project/guidellm/blob/03358ed980f4404f03fea0a73463aa9252b7edc5/src/guidellm/cli/benchmark/run.py#L427-L432

But it does not handle nested args yet. Just needs some adjustments. It will be easier to do after the CLI is redone since there will be less conditional cases.

src/guidellm/data/deserializers/huggingface.py
```diff
@@ -17,27 +16,45 @@
...
 __all__ = ["HuggingFaceDatasetDeserializer"]
 
 
+@DataArgs.register("huggingface")
+class HuggingFaceDataArgs(DataArgs):
+    kind: str = Field(
+        default="huggingface",
+        description="Type identifier for the Hugging Face dataset deserializer.",
+    )
+    data: str | Dataset | IterableDataset | DatasetDict | IterableDatasetDict = Field(
```
jaredoconnell: `src` makes more sense for this type of input. It's the source of the data.

src/guidellm/benchmark/profiles.py
```diff
@@ -46,6 +45,11 @@
 if TYPE_CHECKING:
     from guidellm.benchmark.schemas import Benchmark
 
+# NOTE: These imports should be removed after changes to trace
```
jaredoconnell, line 48: As in the bigger refactor to fully support trace datasets?

src/guidellm/cli/benchmark/run.py
```diff
@@ -379,10 +342,31 @@ def run(**kwargs):  # noqa: C901
             kwargs["output_dir"] = path.parent
             kwargs["outputs"] = (path.name,)
 
+    # NOTE: Hardcode some of the defaults for now
```
jaredoconnell, line 345: What's the longer-term plan regarding these defaults?

jaredoconnell (commented):
This is a large improvement. I just have a few comments.

src/guidellm/benchmark/profiles.py
sjmonson: Yes-ish, these are only needed because the profile has to load the trace dataset. We can drop that entirely, doesn't need to be with full support though we can do it earlier.

src/guidellm/cli/benchmark/run.py
sjmonson: `BenchmarkArgs` will hold the default `kind` for each `...Args` and each `...Args` subclass can hold its own default fields.

src/guidellm/data/deserializers/huggingface.py
sjmonson: Maybe `source` instead? I don't like making the word longer but I think its probably better to use the full word.

jaredoconnell: Sure. Maybe have `src` as an alias. It's such a common term in CS.

dbutenhof: While I see the point about "data" being a little awkward in the "--data data args", I don't really see a problem, and I don't see "source" as an improvement in UX clarity.

But to get into the spirit, if "data" doesn't have enough context, we can avoid the extra two letters with "from" --data "type=huggingface,from=org/dataset-name"

jaredoconnell: To me "data" infers more inlined data, rather than a place to get data from. So data makes sense for the synthetic data, for example.

sjmonson: Done

src/guidellm/data/deserializers/huggingface.py
```diff
@@ -17,27 +17,46 @@
...
 
-@DatasetDeserializerFactory.register("huggingface")
+@DataArgs.register(["huggingface", "hf"])
+class HuggingFaceDataArgs(DataArgs):
+    kind: Literal["huggingface", "hf"] = Field(
+        default="huggingface",
+        description="Type identifier for the Hugging Face dataset deserializer.",
+    )
+    source: str | Dataset | IterableDataset | DatasetDict | IterableDatasetDict = Field(
+        validation_alias=AliasChoices("source", "src", "from", "path", "name"),
```
jaredoconnell, line 40: Since it's aliased now, I wonder if it makes sense to also alias "id".

jaredoconnell (commented):
In several places the markdown shows the old arg for huggingface inputs "data".

dbutenhof (dismissed):

jaredoconnell (approved):

dbutenhof (approved):

PR #736, opened 2026-05-20
Add vLLM Python batch backend for micro-batched inference

maryamtahhan:
## Summary

Adds `VLLMPythonBatchBackend` (`kind=vllm_python_batch`) for throughput-oriented inference using vLLM's synchronous `LLM` engine. Requests queue and flush in configurable micro-batches via `LLM.generate()`.

Renames the existing in-process backend to `vllm_python_async`. The legacy `vllm_python` kind remains accepted as an alias for `vllm_python_async`.

Benchmark logging silencing, worker stdio suppression, and Rich progress refresh were split to #1035 per review feedback.

## Key design

- **Engine lifecycle**: engine creation deferred to worker `_ensure_engine()`; main preflight validates without loading weights; workers preload in `validate()` via `multiprocessing.parent_process()`.
- **Batching**: `_batch_lock` for queue/shutdown; `_generate_lock` serializes `LLM.generate()`; partial batches flush after `batch_timeout`.
- **Cancellation safety**: executor `generate()` is shielded so constraint/task cancellation waits for in-flight generation before releasing `_generate_lock`.
- **Fork/spawn**: asyncio locks recreated in `process_startup()`; locks stripped for pickle.
- **Metrics**: `RequestStateStats` mapped to wall-clock timings; chat-template fallback for plain LMs (e.g. opt-125m).

## Usage

```bash
guidellm run \
  --backend kind=vllm_python_batch,model=Qwen/Qwen3-0.6B,batch_size=8 \
  --profile kind=throughput,max_concurrency=20 \
  --constraint kind=max_requests,count=100
```

## Files

| Area | Files |
|------|-------|
| Batch backend | `src/guidellm/backends/vllm_python/batch.py` |
| Async rename + alias | `src/guidellm/backends/vllm_python/vllm.py` |
| Shared utilities | `src/guidellm/backends/vllm_python/common.py` |
| Docs | `docs/guides/vllm-python-batch-backend.md`, `docs/guides/backends.md` |
| Tests | `tests/unit/backends/vllm_python/test_vllm_batch.py` |

## Test plan

- [x] Unit tests for batch engine laziness, shutdown drain, generate-lock serialization, cancellation shielding, metrics, chat-template fallback
- [x] `vllm_python` alias registration test
- [x] EC2 validation with `facebook/opt-125m` (20/20 requests, JSON output)
- [ ] Re-request review from @sjmonson (logging split addressed in #1035)

## Related

- Logging/stdio/progress follow-up: #1035

---

## Use of AI

Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

---

# git log

commit 5bcb62b3e247e90dfd398b4f19d823733c341145
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Fri Jul 17 14:30:31 2026 +0100

    Add vLLM Offline Backend for batch processing
    
    Introduces VLLMOfflineBackend inheriting from VLLMPythonBackend,
    using vLLM's synchronous LLM engine for micro-batched inference.
    Wires RequestStateStats metrics into response timings. Includes
    registration test, docs, and backends.md update.
    
    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 62c919fc750f46ee2e67963791335302a36eb717
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Fri Jul 17 15:33:26 2026 +0100

    Fix tokenizer access path for vLLM V1 LLM engine
    
    Use llm.get_tokenizer() instead of llm.llm_engine.tokenizer.tokenizer
    which doesn't exist in the V1 engine.
    
    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit beb1c0a30fd24d82e1d2702bf784ac4f88f2a397
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Fri Jul 17 15:41:25 2026 +0100

    Fix formatting, metrics, and mdformat issues
    
    - Apply ruff format to offline.py
    - Fix _wire_vllm_metrics: only use token counts, not monotonic
      timestamps that produce garbage when mixed with wall-clock times
    - Apply mdformat to markdown docs
    
    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit d8d690ef9cdfa4929c302957d21884b0cc4a61c6
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Mon Jul 20 17:27:48 2026 +0100

    Defer vLLM engine creation to first request and reset CPU affinity
    
    The vLLM LLM engine is now created lazily on the first resolve() call
    instead of during process_startup(). This avoids a double-startup that
    wastes resources reloading model weights and can cause CPU-affinity
    degradation when the engine is created in both the validation path and
    the worker process.
    
    Uses double-checked locking with asyncio.Lock to prevent concurrent
    engine creation when multiple requests arrive simultaneously.
    
    Adds _reset_cpu_affinity() to restore the full cgroup cpuset before
    engine creation, working around OpenMP/torch restricting CPU affinity
    in forked worker processes.
    
    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 211e74f8133e25df57d650907ca816d9dcb22eee
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Tue Jul 21 12:57:05 2026 +0100

    Wire vLLM RequestStateStats metrics and fix batch lock contention
    
    Map vLLM's monotonic RequestStateStats timestamps to wall-clock
    values so TTFT, ITL, and TPOT are reported for offline batches.
    Split _process_batch into _take_pending_batch + _run_generate to
    release the batch lock during LLM.generate(), and loop the deferred
    flush until all pending requests are drained.
    
    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit f375739b9b6265b7a9d50a372316d30811675d78
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Tue Jul 21 14:31:04 2026 +0100

    Address review: move CPU affinity to common module, use _shutting_down, call shutdown
    
    - Move _reset_cpu_affinity to common.py so both vllm_python backends
      can reuse it; add comment explaining cgroup v2/v1 fallback loop
    - Guard resolve() with _shutting_down check to reject requests during
      shutdown
    - Call llm.shutdown() before deleting the engine in process_shutdown
    
    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 3b1b060d6d012da70eaf93bd291d525ad3e14cb2
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Tue Jul 21 15:55:21 2026 +0100

    Fix concurrency hazards and clean up batch request dataclass
    
    - Add _generate_lock to serialize LLM.generate() calls, preventing
      overlapping generates on the non-thread-safe sync LLM engine
    - Move _shutting_down check + enqueue under the same batch lock
      acquisition so shutdown cannot strand in-flight resolve() waiters
    - Set _shutting_down under lock in process_shutdown for atomicity
    - Remove unused _BatchedRequest fields (request, request_info,
      request_id) and the uuid import
    - Drop duplicate validate_vllm_config; inherited from parent
    
    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit cb0684f4f011c736466a948f44c238d3f2898bc0
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Tue Jul 21 16:11:06 2026 +0100

    Add unit tests for vLLM offline backend
    
    Covers engine laziness, lifecycle/shutdown, batch processing,
    deferred flush, generate-lock serialization, shutdown-guard,
    and metrics wiring with 31 test cases.
    
    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit a784db2d79907ef494ea3407a75ef3b35c51b461
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Tue Jul 21 16:26:23 2026 +0100

    Fix shutdown race, add batch_timeout, and finish test coverage
    
    Shutdown now waits for the deferred-flush task to complete and
    acquires _generate_lock before tearing down the engine, preventing
    a race where an in-flight LLM.generate() outlives _llm teardown.
    
    Add batch_timeout field (default 0.01s) to VLLMOfflineBackendArgs
    so partial batches accumulate for a configurable window before
    flushing; full batches still flush immediately.
    
    Add ## WRITTEN BY AI ## docstrings to all 34 test functions, replace
    the old cancellation test with test_shutdown_waits_for_inflight_generate,
    and add batch_timeout validation tests.
    
    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 9925b20dee81d8da914c1851bbe598a0e54057bd
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Tue Jul 21 16:41:47 2026 +0100

    Replace guide example with throughput profile and remove getattr from metrics
    
    Use throughput,max_concurrency=20 instead of constant,rate=3 in the
    offline backend guide example to better reflect the backend's
    throughput-oriented purpose.
    
    Replace all getattr calls in _wire_vllm_metrics with hasattr + direct
    attribute access per AGENTS.md coding standards.
    
    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 602b0511205ada535a509f2b3bfcacb4f48a3cbe
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Tue Jul 21 17:33:45 2026 +0100

    Reset asyncio locks in process_startup to fix forked worker hang
    
    The benchmark framework calls process_startup/validate/shutdown in
    the parent process, then forks worker subprocesses that call
    process_startup again. Asyncio locks created on the parent event
    loop are unusable in the child, causing the worker to hang and
    produce 0 completed requests.
    
    Recreate all asyncio primitives (_batch_lock, _generate_lock,
    _engine_lock) and reset batch state in process_startup() so each
    worker gets fresh locks bound to its own event loop.
    
    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 2faa3e202090860ce77c84996ae2ab89538f157a
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Wed Jul 22 10:31:21 2026 +0100

    Preload engine in forked workers to fix 0-completion benchmark
    
    The vLLM engine cold-start (42s on CPU) was happening during the
    timed benchmark phase inside resolve(), causing SIGTERM to kill the
    EngineCore before any requests could complete.
    
    Record the creator PID in __init__ so validate() can detect forked
    workers and eagerly preload the engine before the timed phase.
    Clear _llm in process_startup so workers never inherit a stale
    engine handle from the parent process.
    
    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 727adc0dbdd8b63c0673644da18b316c8be52d3f
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Wed Jul 22 10:41:26 2026 +0100

    Polish engine lifecycle docs and strengthen lock-reset test
    
    Rewrite the Engine lifecycle section to clearly describe the
    three-stage flow: parent preflight (cheap check), worker preload
    (PID-based eager engine creation), and resolve() idempotent
    fallback.  Mention both fork and spawn workers.
    
    Strengthen test_startup_recreates_locks: use `is not` per lock,
    assert batch state is reset, and acquire/release each new lock
    to prove they work on the current event loop.
    
    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 51d1154b7c1f23ef685dc043bb21c54ad21d75fe
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Tue Aug 4 10:38:30 2026 +0100

    Address PR review comments for vllm_offline backend
    
    - Replace hasattr guards in _wire_vllm_metrics with direct field access
      on vllm.RequestOutput / RequestStateStats; type request_output
      properly and cast at the call site in resolve()
    - Add explanatory comment in reset_cpu_affinity() for the cgroup v2/v1
      loop-with-return pattern that reads as odd at first glance
    - Call reset_cpu_affinity() in VLLMPythonBackend.process_startup() so
      the regular in-process vLLM backend benefits from the same CPU-affinity
      fix as the offline backend
    - Replace 'resolve() fallback' with 'Inference-time safety net' in
      docs to avoid exposing the internal method name to end users
    
    Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit d5d17d2180bf3e9e3657099af5822c7e5fe7212d
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Tue Aug 4 10:44:04 2026 +0100

    Fix partial-metrics crash in _wire_vllm_metrics and address review nits
    
    Use getattr with 0.0 defaults for timing fields so that metrics objects
    carrying only num_generation_tokens (e.g. test stubs) no longer raise
    AttributeError after the metrics-is-None guard. Also update the validate()
    docstring to say "inference-time safety net", fix the test_no_metrics_no_crash
    docstring, and add a smoke test asserting reset_cpu_affinity() is called in
    VLLMPythonBackend.process_startup().
    
    Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 26013505b113c3b7371737c2f599c6da8ae11a88
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Tue Aug 4 11:21:35 2026 +0100

    Detect offline worker processes via parent_process() for engine preload
    
    Replace fragile _creator_pid comparison with multiprocessing.parent_process()
    so scheduler workers reliably preload the vLLM engine before the timed phase,
    avoiding 0-completion runs when cold-start lands inside resolve().
    
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 82036c691c03b9bbbd0a004c0e93b6d6105b2b08
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Tue Aug 4 14:22:05 2026 +0100

    Fix benchmark progress panel stacking during vLLM offline worker init.
    
    Rich Live's background auto-refresh thread re-rendered the Benchmarks panel
    every 250 ms.  Between renders, vLLM worker processes wrote to the inherited
    terminal file descriptors during engine initialisation, shifting the cursor
    so each re-render landed below the previous one instead of overwriting it.
    
    Switch to manual refresh (auto_refresh=False) and drive repaints explicitly:
    
    * _force_refresh()     — immediate render on state-transition events
                             (benchmark start, complete, finalize)
    * _throttled_refresh() — caps update rate at refresh_per_second (~4 fps)
                             during active benchmarking; no-op between events
    
    The window between on_benchmark_start() and the first on_benchmark_update()
    now contains zero renders, so worker-process writes cannot cause stacking.
    entrypoints.py calls prepare_vllm_benchmark_logging() before workers start
    so env-var silencing is inherited by child processes.
    
    Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 1cb5dcc52694981b45bda5f6d58a3c88055e5f50
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Tue Aug 4 14:22:24 2026 +0100

    Suppress vLLM worker startup output and fix offline batch waiter hang.
    
    Worker processes write HF Hub progress bars, tqdm model-loading bars, and
    C-level PyTorch warnings (e.g. NUMA) directly to the inherited terminal
    file descriptors during engine initialisation.  These writes interleave
    with the main-process Rich live panel even when Python-level log levels
    are lowered.
    
    Fixes:
    
    * Add suppress_worker_stdio() (utils/terminal.py) — redirects OS-level
      fd 1/fd 2 via dup2() to /dev/null and replaces sys.stdout/sys.stderr,
      silencing both C-level and Python-level writes for the duration.
    * Wrap backend.validate() in worker._processing_startup() with
      suppress_worker_stdio() so the vLLM LLM engine cold-start is silent.
    * Add VLLM_CONFIGURE_LOGGING=0 to prepare_vllm_benchmark_logging() so
      vLLM's EngineCore subprocesses do not re-run their own log handler
      setup after the level has been lowered.
    * Call prepare_vllm_benchmark_logging() in VLLMPythonBackend.process_startup()
      and use vllm_benchmark_engine_config() for the async engine so the
      online backend also benefits from quieter defaults.
    * Fix offline batch waiter hang: replace zip(..., strict=True) outside
      the try/except in _run_generate() with an explicit length check that
      routes mismatches through the error path, ensuring all batch waiters
      always receive ready.set() and resolve() calls never hang.
    
    Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 0e892da439cdfaa938d8547d120d1e623873b15a
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Tue Aug 4 14:30:12 2026 +0100

    Address low-severity gaps: fd leak, markers, and missing tests.
    
    * terminal.py: store both open(os.devnull) handles before the try block
      and close them in finally, eliminating the two-fd leak per call.
    * test_progress.py: add ## WRITTEN BY AI ## module docstring, smoke/sanity
      markers, and descriptive per-test docstrings per AGENTS.md; extract the
      repeated patch target into a module-level constant to stay within the
      79-char line limit.
    * test_vllm_offline.py: add test_run_generate_output_count_mismatch_signals
      _all_waiters — verifies the len(outputs) != len(batch) path routes all
      waiters through the error handler so resolve() never hangs.
    * test_terminal.py: new POSIX unit test file for suppress_worker_stdio()
      covering stdout/stderr silencing, sys.stdout/stderr restoration,
      fd restoration, fd-leak detection, and exception-safety.
    
    Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 72a88e914d7b3bd21e9145b07331e0ef09d86c3e
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Tue Aug 4 15:20:28 2026 +0100

    Fix offline batch waiter hangs and spawn worker pickling.
    
    Defer asyncio lock creation to process_startup() and strip locks in
    pickle state so spawn workers can unpickle the backend. Route all
    _run_generate() failures through _signal_batch_failure() so every
    batch waiter is unblocked on any exception.
    
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit c65a0200a604e94b07277e381f8db12c7cb12eac
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Wed Aug 5 08:58:41 2026 +0100

    Fix quality check failures: mypy union-attr, ruff lint, and ruff format
    
    - offline.py: replace Lock | None with if-is-None guards so mypy can
      narrow the type without assert (avoids S101); add noqa: BLE001 on
      intentional broad except in _run_generate; use strict=False on the
      post-length-check zip
    - test_terminal.py: replace os.listdir/os.path.exists/os.readlink
      with pathlib equivalents (PTH208/PTH110/PTH115), drop unused probe
      variable (F841), remove f-string with no placeholders (F541), combine
      nested with statements (SIM117), shorten long docstrings (E501),
      add Path import
    - test_progress.py: remove parentheses from @pytest.fixture (PT001),
      shorten two over-limit docstrings (E501)
    
    Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit ee6b03c15f65b1558f45d92b7c5448ed85ced225
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Wed Aug 5 11:44:45 2026 +0100

    Fix remaining ruff lint failures blocking quality CI.
    
    Resolve import ordering, line length, pathlib, and pickle test lint issues so precommit and quality-checks pass on PR #736.
    
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 431b93daa62ec9c04642be9155bd355308edb551
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Fri Aug 7 11:48:22 2026 +0100

    Remove Windows guard from suppress_worker_stdio.
    
    GuideLLM does not support Windows; the no-op branch was dead code.
    
    Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit b29b1aacd9c937b13153d425a2da0356e45cd475
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Fri Aug 7 11:50:45 2026 +0100

    Simplify prepare_vllm_benchmark_logging to env vars only.
    
    Remove the loggerDict iteration and the private vllm.logger._configure_vllm_root_logger
    call. Both were workarounds for reconfiguring vLLM logging after it had already been
    imported, but guidellm.extras lazy-loads vLLM so the env vars are always set before
    the first vLLM import in any worker process. A single logging.getLogger("vllm").setLevel()
    call is kept as a clean stdlib safety net.
    
    Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit f0628e8ec8cb6b8ebdfcef4269a6ce1fa9a83536
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Fri Aug 7 12:29:45 2026 +0100

    Fall back to plain prompt when model has no chat template.
    
    For default-template request format, apply_chat_template() raises ValueError
    when the model tokenizer has no chat_template set (e.g. facebook/opt-125m).
    Catch ValueError in both vllm and offline backends and fall back to
    _extract_prompt_chat_plain() so plain-LM models work without extra config.
    
    Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 4c8597a93496d1ae36d81d187739f747efe15259
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Fri Aug 7 12:51:39 2026 +0100

    Harden prepare_vllm_benchmark_logging with handler lowering and defensive vLLM reconfigure.
    
    Lower handler levels on the vllm logger in addition to the logger itself,
    add a safety-net call to _configure_vllm_root_logger when vllm.logger is
    already in sys.modules (guarded by try/except for API drift), and extend
    the docstring to explain why in-process reconfiguration is still needed
    despite lazy loading. Adds a regression test covering the error-suppression path.
    
    Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit a0ab3f19ebe2b84d3a28edc1857506d451371f7d
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Fri Aug 7 12:57:01 2026 +0100

    Fix ruff lint errors in prepare_vllm_benchmark_logging hardening.
    
    Add noqa: SLF001 for the intentional private-member access on
    _configure_vllm_root_logger, and shorten two lines that exceeded
    the 88-character limit.
    
    Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 86c92f5f9920e9f3fa38424d67006a73c62b3eb3
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Fri Aug 7 12:59:19 2026 +0100

    Lower vllm.* child logger levels during benchmark logging setup.
    
    Extend prepare_vllm_benchmark_logging to quiet already-registered vllm.*
    child loggers and narrow the defensive catch around vLLM's private root
    logger reconfigure API.
    
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit d45c064bf2a19d0c2f93931396ba8caff393b19f
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Fri Aug 7 13:49:53 2026 +0100

    Address PR #736 review nits for VLLMOfflineBackend.
    
    - Split _validate_backend_initialized() into _validate_process_started()
      (process check only) and _require_llm() (also asserts engine exists),
      removing the fragile pattern of returning a potentially-None LLM.
    - Extract _try_reconfigure_vllm_root_logger() in common.py to isolate
      the private vLLM API call with documented drift risk; tighten docstring
      to explain all three silencing layers and when each is needed.
    - Guard _schedule_deferred_flush() task creation under _batch_lock to
      prevent concurrent resolve() calls from spawning duplicate flush tasks.
    - Fix _wire_vllm_metrics() clock anchoring: use queued_ts as mono_base
      (arrival_time is vLLM's queue-arrival wall-clock time); fall back to
      scheduled_ts when queued_ts is absent. Wire scheduled_at from vLLM
      RequestStateStats when both queued_ts and scheduled_ts are available.
    - Replace hasattr(self._llm, "shutdown") with contextlib.suppress(AttributeError).
    
    Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 3d77915400793e361f209a29c46d36ac1eb1671b
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Thu Aug 20 10:06:36 2026 +0100

    Split logging UX from batch backend and rename vLLM backends.
    
    Remove benchmark logging silencing, worker stdio suppression, and manual
    Rich progress refresh from this PR per review feedback; those changes
    live on feat/vllm-benchmark-logging-silencing.
    
    Rename backend kinds to vllm_python_async and vllm_python_batch, with
    matching class and documentation updates.
    
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 34936194e616a7cda5679b1b146057e869dbe44f
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Thu Aug 20 10:30:57 2026 +0100

    Fix mdformat for vLLM batch and migration guide docs.
    
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 4146ca77726d17d9f71bc800c460b43ddeec6dbf
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Thu Aug 20 10:36:32 2026 +0100

    Rename remaining offline references to batch backend naming.
    
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit 933263b0780c028cd73694df933d59fe7bda559a
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Thu Aug 20 10:49:16 2026 +0100

    Shield batch generate from cancellation and register vllm_python alias.
    
    - Shield executor LLM.generate() so cancellation waits before releasing lock
    - Register vllm_python as alias for vllm_python_async
    - Export VLLMPythonBatchBackend from guidellm.backends
    - Fix batch doc example and v0.6 migration guide backend names
    - Add chat-template fallback tests for async and batch backends
    
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

commit fd8af30c3ded50bdf70aea05eaeb7d65203b6caa
Author: Maryam Tahhan <mtahhan@redhat.com>
Date:   Thu Aug 20 10:55:26 2026 +0100

    Re-raise CancelledError after shielded batch generate completes.
    
    Extract _await_shielded_executor so cancellation always propagates even
    when the executor raises, and unblock the cancellation unit test promptly.
    
    Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>

---------

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
Signed-off-by: Maryam Tahhan <mtahhan@redhat.com>


maryamtahhan:
@sjmonson @jaredoconnell this PR has been rebased and is green again

src/guidellm/backends/vllm_python/common.py
```diff
@@ -0,0 +1,347 @@
...
+        tokenize=False,
+        add_generation_prompt=True,
+    )
+    if isinstance(prompt, str):
+        return prompt
+    raise RuntimeError("Backend received unexpected type from tokenizer.")
+
+
+def create_sampling_params(
+    vllm_module: Any,
```
sjmonson: Don't pass modules as arguments. Just move `from guidellm.extras import vllm` out of the `TYPE_CHECKING` block. `guidellm.extras` builds lazy stubs so it won't actually import vLLM until you call something on the module.

src/guidellm/backends/vllm_python/offline.py
```diff
@@ -0,0 +1,567 @@
...
+
+from guidellm.backends.backend import Backend, BackendArgs
+from guidellm.backends.vllm_python import common
+from guidellm.backends.vllm_python.vllm_response import VLLMResponseHandler
+from guidellm.logger import logger
+
+if TYPE_CHECKING:
+    from guidellm.extras import vllm
+else:
+    from guidellm.extras import vllm
```
sjmonson: Its funny this is not a linting error.

```suggestion
from guidellm.extras import vllm
```

src/guidellm/backends/vllm_python/offline.py
```diff
@@ -0,0 +1,567 @@
...
+        await self._maybe_process_batch()
+
+        # Wait for result
+        await batched_req.ready.wait()
+
+        # Build response
+        request_info.timings.request_end = time.time()
+
+        if batched_req.result is not None:
+            output = batched_req.result
```
sjmonson: So poking around the vLLM docs a bit it looks like vLLM will tell us a lot about the request from what it returns. Maybe take a look at `batched_req.result.metrics` which is a [`RequestStateStats`](https://docs.vllm.ai/en/latest/api/vllm/v1/metrics/stats/#vllm.v1.metrics.stats.PromptTokenStats.get_by_source) and wire all of the latencies to the correct metrics. You can get pretty much all of the latency metrics we normally track. Start/end time I would probably keep how you currently have them though since that is what the "user" sees as start/end.

src/guidellm/backends/vllm_python/offline.py
```diff
@@ -0,0 +1,567 @@
...
+
+    @property
+    def processes_limit(self) -> int | None:
+        """Limit to single process for batch coordination."""
+        return 1
+
+    @property
+    def info(self) -> dict[str, Any]:
+        """Get backend configuration details."""
+        return self._args.model_dump()
```
sjmonson: Will need to be dropped after #912

src/guidellm/backends/vllm_python/offline.py
```diff
@@ -0,0 +1,567 @@
...
+        self.resolved_prompt = resolved_prompt
+        self.multi_modal_data = multi_modal_data
+        self.max_tokens = max_tokens
+        self.request_id = str(uuid.uuid4())
+        self.result: vllm.RequestOutput | None = None
+        self.ready = asyncio.Event()
+
+
+@Backend.register("vllm_offline")
+class VLLMOfflineBackend(Backend):
```
sjmonson: I didn't check too closely but it seems like even with `common.py` the normal and offline backends share a lot of code. Would it make more sense if the base class was the other vLLM backend? E.g.

```suggestion
@Backend.register("vllm_offline")
class VLLMOfflineBackend(VLLMPythonBackend):
```

Either way I think you should keep `common.py` as it makes the classes easier to read.

sjmonson (changes requested):

src/guidellm/backends/vllm_python/batch.py
```diff
@@ -0,0 +1,567 @@
...
+            batch_size=64,
+            vllm_config={"tensor_parallel_size": 2}
+        )
+        backend = VLLMOfflineBackend(args)
+        await backend.process_startup()
+        async for response, info in backend.resolve(request, request_info):
+            process_response(response)
+        await backend.process_shutdown()
+    """
+
```
dbutenhof, line 141: add

```python
    _args: VLLMOfflineBackendArgs
```

and remove the `self._args = arguments` in `__init__`, along with removing the `info` method as Sam mentioned. This supports the new inherited `info` that correctly handles JSON-ifying the arguments in the generated report files.

docs/guides/vllm-offline-backend.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Installation
+
+The offline backend requires vLLM to be installed. See the [vLLM Python Backend installation guide](vllm-python-backend.md#installation) for recommended installation methods.
+
+## Basic Usage
+
+```bash
+guidellm benchmark run \
+  --backend vllm_offline \
```
dbutenhof: This command is now obsolete. Replace it with:

```console
guidellm run \
    --backend kind=vllm_offline,model=Qwen/Quen3-0.6B,batch_size=64 \
    --constraint kind=max_requests,count=1000
```

docs/guides/vllm-offline-backend.md
```diff
@@ -0,0 +1,235 @@
...
+  --model "Qwen/Qwen3-0.6B" \
+  --backend-kwargs '{"batch_size": 64}' \
+  --data "prompt_tokens=256,output_tokens=128" \
+  --max-requests 1000
+```
+
+## Backend Options
+
+Configure the offline backend via `--backend-kwargs` with JSON:
```
dbutenhof: Specify backend parameters through the `--backend` with `kind=vllm_offline`, for example

```console
guidellm run \
    --backend `{
      "kind": "vllm_offline",
      "model": ""meta-llama/Llama-2-7b-hf",
      "batch_size": 64,
      "vllm_config": {
          "tensor_parallel_size": 2,
          "gpu_memory_utilization": 0.9
      }
     }'
```

docs/guides/vllm-offline-backend.md
```diff
@@ -0,0 +1,235 @@
...
+  - `"plain"`: No chat template, plain text concatenation
+  - Path or string: Custom Jinja2 chat template
+- **`image_placeholder`**: Placeholder for images (default: `"<image>"`)
+- **`audio_placeholder`**: Placeholder for audio (default: `"<|audio|>"`)
+
+## How Micro-Batching Works
+
+The offline backend uses a **micro-batching** approach:
+
+1. **Buffering**: As requests arrive via `resolve()`, they're added to a buffer
```
dbutenhof: At this level, I don't think I'd use the `resolve()` method name: it won't mean much to most readers, and isn't obvious. Maybe, "as requests are generated from the dataset" ...

docs/guides/vllm-offline-backend.md
```diff
@@ -0,0 +1,235 @@
...
+4. **Result Distribution**: Return cached results to waiting requests
+5. **Flush on Shutdown**: Remaining requests processed when backend shuts down
+
+This gives you 10-100x fewer model forward passes compared to per-request processing while working within GuideLLM's scheduler architecture.
+
+## Examples
+
+### Basic Throughput Benchmark
+
+```bash
```
dbutenhof: Again, this (and the following examples) need to be updated to the new CLI model.

docs/guides/vllm-offline-backend.md
```diff
@@ -0,0 +1,235 @@
...
+| 32-64      | Good       | Medium  | Medium | General use, balanced        |
+| 128-256    | High       | High    | High   | Large GPUs, max throughput   |
+
+**Rule of thumb**: Start with 32, increase until GPU utilization >90% or OOM.
+
+### Memory Optimization
+
+```bash
+# Reduce memory usage
+--backend-kwargs '{
```
dbutenhof: Again, these are all under `--backend` now, as I showed above.

docs/guides/vllm-offline-backend.md
```diff
@@ -0,0 +1,235 @@
...
+### "Backend not started up for process"
+
+The backend wasn't initialized. Ensure your benchmark calls the backend lifecycle correctly (this should happen automatically).
+
+### Out of Memory (OOM)
+
+Reduce `batch_size` or `gpu_memory_utilization`:
+
+```bash
+--backend-kwargs '{"batch_size": 16, "vllm_config": {"gpu_memory_utilization": 0.7}}'
```
dbutenhof: And here... plus a few more below

dbutenhof (changes requested):
I think the code is mostly OK (except it should be updated to match the current backend init conventions). The documentation needs to be updated to 0.7 CLI conventions.

src/guidellm/backends/vllm_python/offline.py
```diff
@@ -0,0 +1,507 @@
...
+        async with self._batch_lock:
+            if self._pending_batch:
+                await self._process_batch()
+
+        if self._processing_task is not None:
+            self._processing_task.cancel()
+            self._processing_task = None
+
+        if self._llm is not None:
+            del self._llm
```
dbutenhof: Should this be calling `self._llm.shutdown()`?

src/guidellm/backends/vllm_python/offline.py
```diff
@@ -0,0 +1,507 @@
...
+    async def process_shutdown(self):
+        """
+        Drain pending batch and tear down the vLLM LLM engine.
+
+        :raises RuntimeError: If backend was not properly initialised
+        """
+        if not self._in_process:
+            raise RuntimeError("Backend not started up for process.")
+
+        self._shutting_down = True
```
dbutenhof: You set `_shutting_down` as if you intend to use it somewhere, but it's not used. Should it terminate `resolve`, for example?

src/guidellm/backends/vllm_python/offline.py
```diff
@@ -0,0 +1,613 @@
...
+        and reset the affinity to the full set.
+        """
+        if sys.platform != "linux":
+            return
+
+        current = os.sched_getaffinity(0)
+
+        for path_str in (
+            "/sys/fs/cgroup/cpuset.cpus.effective",
+            "/sys/fs/cgroup/cpuset/cpuset.cpus",
```
dbutenhof: I assume this is basically a Cgroups v2/v1 accommodation ... this deserves a comment, since at first glance the "loop" looks odd.

Furthermore, this likely should also apply to the normal in-process vLLM backend, so it should ideally be implemented in a common module under `vllm_python`. (Even if you don't add it to the standard vllm_python backend now.)

dbutenhof (commented):
Some comments ... the ones around shutdown are actually from a Cursor AI review of your branch -- there's more, but I didn't copy everything.

maryamtahhan:
> Some comments ... the ones around shutdown are actually from a Cursor AI review of your branch -- there's more, but I didn't copy everything.

Thank you @dbutenhof I will have a look

maryamtahhan:
not ready for re-review... will add a comment when it is. Still doing some testing

maryamtahhan:
Ready for Re-review

src/guidellm/backends/vllm_python/batch.py
```diff
@@ -0,0 +1,618 @@
...
+            "partial batch.  Full batches bypass this delay."
+        ),
+    )
+
+    # Hide the inherited ``stream`` field -- offline is never streaming.
+    stream: Literal[False] = Field(  # type: ignore[assignment]
+        default=False,
+        exclude=True,
+        description="Offline backend does not support streaming.",
+    )
```
dbutenhof: It would be cleaner to build a common base class for the in-process vLLM backends (possibly in `common.py`) and then add `stream` only to the asynchronous subclass rather than trying to hide it here. I guess I won't quite go to the extent of demanding a change here, since it is more work & touch surface and this will probably suffice ... but it *would* be cleaner. 😁

dbutenhof (dismissed):

src/guidellm/backends/vllm_python/offline.py
```diff
@@ -0,0 +1,618 @@
...
+            return
+
+        arrival = metrics.arrival_time if hasattr(metrics, "arrival_time") else 0.0
+        scheduled = metrics.scheduled_ts if hasattr(metrics, "scheduled_ts") else 0.0
+        queued = metrics.queued_ts if hasattr(metrics, "queued_ts") else 0.0
+        mono_base = scheduled or queued
+        first_tok = (
+            metrics.first_token_ts if hasattr(metrics, "first_token_ts") else 0.0
+        )
+        last_tok = metrics.last_token_ts if hasattr(metrics, "last_token_ts") else 0.0
```
jaredoconnell: I don't like the repeated use of `hasattr`. Can you import `from vllm import RequestOutput as RequestOutput`, and use that to change the type from `Any` to `vllm.RequestOutput`? And then you can access the field directly, with the type `RequestStateStats | None`.

The only reason I could see this being necessary is if some versions of vLLM that you want to support have different fields here. But this kind of thing makes static type analysis fail, resulting in successful runs with blank metrics if vLLM's types change.

jaredoconnell (commented):
Looks pretty good. I have one comment.

jaredoconnell (dismissed):

src/guidellm/utils/terminal.py
```diff
@@ -0,0 +1,51 @@
...
+    Used in scheduler workers during heavyweight backend startup (e.g. vLLM
+    engine initialisation) so that C-level writes from subprocesses and
+    third-party libraries cannot corrupt the main-process Rich live display.
+    Python-level ``sys.stdout``/``sys.stderr`` are also replaced so that any
+    higher-level writes (HF Hub progress bars, tqdm) are silenced too.
+    POSIX only; a no-op on Windows.
+    """
+    if sys.platform == "win32":
+        yield
+        return
```
sjmonson: We do not support windows. Please remove this bit.

sjmonson (changes requested):
Not a huge fan on how all of the logging redirection is set up but fine for now I guess. Otherwise just one high-level nit.

src/guidellm/utils/terminal.py
maryamtahhan: Thanks @sjmonson 
Done — Windows guard removed from suppress_worker_stdio. 

Also simplified prepare_vllm_benchmark_logging while I was in there: dropped the loggerDict iteration and the _configure_vllm_root_logger private API call. Workers inherit env vars set in entrypoints.py, so vLLM's VLLM_LOGGING_LEVEL/VLLM_CONFIGURE_LOGGING are already in place before its first import — the reconfigure-after-the-fact code was dead weight.

And found a bug during testing: apply_chat_template raises ValueError for plain LMs with no chat template (e.g. facebook/opt-125m) when request_format='default-template'. Added a fallback to _extract_prompt_chat_plain in both backends.

maryamtahhan:
This is quite a large PR... I can try to break it down into several if it would help with the reviews

sjmonson:
Sorry for all the delays. We have had a lot of vacations on the team recently. 

> This is quite a large PR... I can try to break it down into several if it would help with the reviews

I think overall most of this PR is fairly self-contained so probably OK with keeping the backend stuff all together. Most of the stuff I am hesitant on is the logging/stdout changes. I have been working on another PR that should let us keep logging enabled without breaking the console output. If you could maybe drop the stuff related from that from this PR (`prepare_vllm_benchmark_logging()`, `utils/terminal.py`, changes to `benchmark/progress.py`, etc) and move them to a new PR until we determine if our other solution will work, that would be helpful.

jaredoconnell (commented):
I retested it. It still works as intended.

I wonder if it makes sense to rename this backend and the old vllm python backend. Both use vLLM's offline inference APIs. But the main difference is the new one is intended for batching, and the existing implementation is more intended for streaming.
It might make sense to rename them both to `vllm_python_async` and `vllm_python_batch`. `offline` isn't bad, but I think there's some ambiguity to those not familiar with vLLM's API documentation.

Regarding logging, I do think the vLLM logs are useful. The intention is to soon improve handling of log messages alongside the interactive display. At that point the best way to handle vLLM log messages is just to show them but annotate that they are from vLLM and not GuideLLM, if possible. But that can be handled later.

maryamtahhan:
@jaredoconnell @sjmonson — re-requesting review.

- Renamed to vllm_python_async / vllm_python_batch; vllm_python still works as an alias
- Logging/stdio/progress moved to #1035 per Sam’s feedback — this PR is batch-only now
- Fixed cancellation/shutdown race on shielded LLM.generate(), updated docs + PR description, EC2 validated with opt-125m

jaredoconnell (approved):

sjmonson (approved):
LGTM, thanks for all of the work!

PR #738, opened 2026-05-22
Made E2E test theoretically run better in CI

jaredoconnell:
## Summary

Setup so that the server restarts on retries, and extend time available. Designed to improve reliability until the E2E tests are refactored.

The current design made it so that it the server failed, it didn't restart on retries. So the retries didn't help with the situation.

## Test Plan

If the CI passes reliably, that's what we're looking for.

---

## Use of AI


dbutenhof (approved):
Sounds like a plausible mitigation; worth a try anyway.

sjmonson (approved):

PR #742, opened 2026-05-22
Fix TTFT measurement for reasoning-capable models

### LINKED ISSUE #737: TTFT is miscalculated for models that emit `delta.reasoning` before `delta.content`
**Description**
When benchmarking a model through the OpenAI-compatible backend, the Time to First Token (TTFT) is incorrectly measured if the model first streams reasoning tokens (the `delta.reasoning` field) before any content tokens.  
The current implementation only considers `delta.content` as the “first token” and completely ignores `delta.reasoning`. As a result, the first token event is not recorded until the first content chunk arrives, which makes TTFT appear larger than it really is.

**Streaming chunk examples**
A typical response stream from a reasoning model:

```json
// Chunk 1 – reasoning token
data: {"id":"chatcmpl-af430a941c7a2bf7","object":"chat.completion.chunk","created":1779347167,"model":"lz-models-http","choices":[{"index":0,"delta":{"reasoning":"Okay"},"finish_reason":null}]}

// … potentially more reasoning chunks …

// Later chunk – first content token
data: {"id":"chatcmpl-af430a941c7a2bf7","object":"chat.completion.chunk","created":1779347167,"model":"lz-models-http","choices":[{"index":0,"delta":{"content":"\n\n"},"finish_reason":null}]}
```

**Root cause**
In `backends/openai/request_handlers.py` around line 573:

```python
if choices and (content := choice.get("delta", {}).get("content")):
```

This condition only triggers when `delta.content` is present. Since reasoning models first populate `delta.reasoning` (e.g., `"Okay"`), the first-token detection is delayed until the first `content` chunk arrives, artificially inflating TTFT.

**Expected behavior**
The first token (whether it is a reasoning token or a content token) should be recognized immediately, and TTFT should reflect the latency of that first chunk.

**Suggested fix**
Extend the condition to also accept `delta.reasoning` as a valid first token:

```python
if choices and (
    content := choice.get("delta", {}).get("content")
    or choice.get("delta", {}).get("reasoning")
):
```

This ensures that the first meaningful token (reasoning or content) is correctly identified.

**Temporary workaround**
Manually applied the patch above on the v0.6.0 release. It resolves the issue in our internal benchmarks.

**Impact**
- All metrics that depend on TTFT (e.g., end-to-end latency percentiles, time-to-first-token distributions) are **wrong** for reasoning-capable models (DeepSeek-R1, o1-style models, etc.).
- Users relying on `guidellm` for such evaluations will obtain misleading performance data and may draw incorrect conclusions about model responsiveness.

**Environment**
- `guidellm` version: v0.6.0 (release)
- Backend: OpenAI-compatible endpoint serving a model that produces reasoning tokens
- Issue also applies to `main` (based on code inspection)

**Additional context**
Consider also checking for any other future “initial token” fields that the OpenAI API may introduce. At a minimum, `reasoning` must be handled correctly.

soyr-redhat:
## Summary

Fixes TTFT (Time to First Token) measurement for reasoning-capable models that emit `delta.reasoning` tokens before `delta.content` tokens.

When benchmarking models through the OpenAI-compatible backend, the TTFT was incorrectly measured if the model first streams reasoning tokens before content tokens. The implementation only considered `delta.content` as the "first token" and ignored `delta.reasoning`, causing TTFT to appear larger than it really is.

This extends `ChatCompletionsRequestHandler.add_streaming_line` to detect both `delta.reasoning` and `delta.content` tokens, ensuring accurate TTFT measurement regardless of which token type arrives first.

## Details

- [x] Add `delta.reasoning` detection in `ChatCompletionsRequestHandler.add_streaming_line`
- [x] Check reasoning tokens before content tokens to maintain logical ordering
- [x] Handle edge case where both fields exist in the same chunk
- [x] Add comprehensive unit tests for reasoning token detection
- [x] Add integration test validating full TTFT measurement flow

## Test Plan

- Run unit tests: `uv run pytest tests/unit/backends/openai/test_request_handlers.py::TestChatCompletionsRequestHandler`
- Run integration test: `uv run pytest tests/unit/backends/openai/test_http.py::TestOpenAIHTTPBackend::test_resolve_stream_reasoning_tokens_ttft`
- All 199 OpenAI backend tests pass
- Code formatting and linting checks pass (`uv run ruff check` and `uv run ruff format --check`)

## Related Issues

- Resolves #737

---

## Use of AI

### DIFF
--- a/src/guidellm/backends/openai/request_handlers.py
+++ b/src/guidellm/backends/openai/request_handlers.py
@@ -815,8 +815,7 @@ class ChatCompletionsRequestHandler(TextCompletionsRequestHandler):
         choice: dict[str, dict] = choices[0] if choices else {}
         delta = choice.get("delta", {}) if choices else {}
 
-        if reasoning := delta.get("reasoning"):
-            self.streaming_texts.append(reasoning)
+        if delta.get("reasoning"):
             updated = True
         if content := delta.get("content"):
             self.streaming_texts.append(content)
--- a/src/guidellm/backends/openai/request_handlers.py
+++ b/src/guidellm/backends/openai/request_handlers.py
@@ -815,6 +815,9 @@ class ChatCompletionsRequestHandler(TextCompletionsRequestHandler):
         choice: dict[str, dict] = choices[0] if choices else {}
         delta = choice.get("delta", {}) if choices else {}
 
+        if reasoning := delta.get("reasoning"):
+            self.streaming_texts.append(reasoning)
+            updated = True
         if content := delta.get("content"):
             self.streaming_texts.append(content)
             updated = True
--- a/tests/unit/backends/openai/test_http.py
+++ b/tests/unit/backends/openai/test_http.py
@@ -964,3 +964,80 @@ class TestCheckToolCallExpectations:
 
         with pytest.raises(ValueError, match="tool call"):
             backend._check_tool_call_expectations(req, resp)
+
+    @pytest.mark.sanity
+    @pytest.mark.asyncio
+    @async_timeout(10.0)
+    async def test_resolve_stream_reasoning_tokens_ttft(
+        self,
+        httpx_mock: HTTPXMock,
+    ):
+        """
+        Test that TTFT is measured correctly when reasoning tokens arrive first.
+
+        Validates that the first_token_iteration timing is set on the first
+        reasoning token, not waiting for the first content token.
+        This is an integration test for issue #737.
+
+        ### WRITTEN BY AI ###
+        """
+        # Create a realistic reasoning model response stream
+        stream_chunks = [
+            # First chunk: reasoning token (should trigger TTFT)
+            b'data: {"id": "chatcmpl-123", "choices": '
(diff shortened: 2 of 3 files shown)


jaredoconnell (changes requested):
After talking with the team, we've determined that reasoning tokens should not be mixed with the content tokens. There are two reason for this. One, reasoning should be differentiated from content. And two, reasoning tokens should not be included in multi-turn conversations. This creates an unrealistic scenario.

For this PR the simplest option is to keep `updated = True`, but remove the appending of reasoning tokens. Another option that's more more involved is to add a variable separate from `streaming_texts` to track reasoning texts in the output, but not include them in the output.

sjmonson (approved):

jaredoconnell (approved):
The primary concern is now fixed. Since I made some changes it's okay since another maintainer also merged.

dbutenhof (approved):

soyr-redhat:
Understood. Thank you all for the help! And my apologies @jaredoconnell  and @sjmonson for failing to rebase/make immediate changes, I was out for the week for my graduation!

Issue #735, opened 2026-05-20
enable multiprocessing support for trace replay


VincentG1234:
### Problem Statement

The initial replay implementation in #620 temporarily forced max_processes=1 to avoid replay ordering / scheduling issues under multiprocessing. 

### Proposed Solution

I may have since found a way to support multiprocessing for replay benchmarks without reintroducing the previous issues.

Instead of assigning replay timestamps at dequeue time, the target replay time is now bound deterministically when requests are enqueued. Workers then consume requests with an already assigned targeted_start, avoiding the shared-index race condition.

Opening this issue before proposing a small follow-up PR to enable multiprocessing support for replay benchmarks.

sjmonson:
We actually had a similar idea; based on some previous design prototypes for another feature where the dataset needed to have control over scheduling. Basically something like what was reverted in this commit: https://github.com/vllm-project/guidellm/commit/a5751de4567b0fa2ac3537240aa32d30acfef19f

The existing design with separate timing and requests is necessary for more dynamic workloads so the thought is that the `RequestSettings` can have another field called something like `relative_timestamp` and on requests scheduling the scheduler passes a request's `RequestSettings` to `SchedulingStrategy.next_request_time()` and then each Strategy can decide how to handle it.

PR #745, opened 2026-05-26
Enable multiprocessing support for trace replay strategy

VincentG1234:
## Summary:

Trace replay previously forced single-process execution because timestamps were assigned at dequeue time via a shared `next_request_index()` counter, causing ordering races under multiprocessing (#620). This PR sources per-request trace offsets from the dataset pipeline via `RequestSettings.relative_timestamp`, following the
design in #735.

## Details
- `trace_synthetic` deserializer emits a `relative_timestamp` column (offset from
  the earliest sorted trace event)
- `GenerativeColumnMapper` maps it to `relative_timestamp_column`
- `GenerativeRequestFinalizer` attaches `RequestSettings` on `GenerationRequest`
- `WorkerGroupState` copies `request.settings` into `RequestInfo` at enqueue
- Add `resolve_dequeued_target_start()` on `SchedulingStrategy`; trace replay
  applies `start_time + time_scale * relative_timestamp` at dequeue
- `TraceReplayStrategy.next_request_time` returns benchmark start (immediate dequeue)
- Remove `processes_limit = 1` on trace replay
- `ReplayProfile` keeps `time_scale` only; default `max_requests` derived from trace row count

## Test Plan:

- [x] Unit tests pass
- [x] Manual trace replay benchmark with `--processes > 1` and verify inter-arrival timing matches the trace

## Related Issues

 #735

## 

- [x]  I used an AI coding assistant while working on this PR

---

# git log

commit 25484a64d8afc95d7467074b2ac235c448c35f10
Author: Vincent Gimenes <vincent.gimenes@gmail.com>
Date:   Fri Jun 5 11:16:54 2026 +0200

    Enable multiprocess trace replay via per-request RequestSettings
    Move trace replay timing from ReplayProfile/TraceReplayStrategy into the
    dataset pipeline so each request carries its own scheduling metadata:
    - trace_synthetic deserializer emits a relative_timestamp column (offset
      from the earliest sorted trace event)
    - GenerativeColumnMapper maps it to relative_timestamp_column
    - GenerativeRequestFinalizer attaches RequestSettings on GenerationRequest
    - WorkerGroupState copies request.settings into RequestInfo at enqueue
    TraceReplayStrategy no longer holds a global relative_timestamps list.
    next_request_time schedules dequeue at benchmark start; resolve_dequeued_target_start
    applies start_time + time_scale * relative_timestamp per request at dequeue.
    This removes the single-process constraint and enables correct multiprocess
    trace replay timing.
    ReplayProfile keeps only time_scale (from rate) and default max_requests
    when data_samples truncates the dataset.
    Add integration test exercising the full trace_synthetic → mapper →
    finalizer → TraceReplayStrategy multiprocess path, plus unit test updates.
    
    Signed-off-by: Vincent Gimenes <vincent.gimenes@gmail.com>

---------

Signed-off-by: Vincent Gimenes <vincent.gimenes@gmail.com>

### DIFF
--- a/src/guidellm/scheduler/strategies.py
+++ b/src/guidellm/scheduler/strategies.py
@@ -26,7 +26,7 @@ from typing import Annotated, ClassVar, Literal, TypeVar
 
 from pydantic import Field, NonNegativeFloat, NonNegativeInt, PositiveInt, PrivateAttr
 
-from guidellm.schemas import PydanticClassRegistryMixin, RequestInfo
+from guidellm.schemas import PydanticClassRegistryMixin, RequestInfo, RequestSettings
 from guidellm.utils.mixins import InfoMixin
 
 __all__ = [
@@ -220,6 +220,26 @@ class SchedulingStrategy(PydanticClassRegistryMixin["SchedulingStrategy"], InfoM
         :param request_info: Completed request metadata including timing details
             and completion status
         """
+
+    async def resolve_dequeued_target_start(
+        self,
+        worker_index: NonNegativeInt,
+        provisional_start: float,
+        settings: RequestSettings,
+    ) -> float:
+        """
+        Resolve scheduled start time after dequeue using per-request settings.
+
+        Default returns ``provisional_start`` unchanged. Strategies with
+        enqueue-bound timing metadata can override to reinterpret settings.
+
+        :param worker_index: Worker process index handling the request
+        :param provisional_start: Start time from the worker's scheduling slot
+        :param settings: Per-request scheduling metadata attached at enqueue
+        :return: Unix timestamp when the request should begin processing
+        """
+        _ = (worker_index, settings)
+        return provisional_start
 
     def requeue_delay(self) -> float:
         """
@@ -679,18 +699,18 @@ class TraceReplayStrategy(SchedulingStrategy):
     """
     Replay scheduling from a trace of timestamps.
 
-    Schedules each request at start_time + time_scale * relative_timestamp[i],
-    so the trace's inter-arrival pattern is reproduced with an optional time scale.
+    Each request carries a ``relative_timestamp`` in ``RequestSettings`` from the
+    dataset finalizer. ``next_request_time`` schedules dequeue immediately at
+    benchmark start; ``resolve_dequeued_target_start`` applies the trace offset via
+    ``start_time + time_scale * relative_timestamp``, reproducing inter-arrival
+    timing under multiprocessing.
     """
 
     type_: Literal["trace"] = "trace"  # type: ignore[assignment]
-    relative_timestamps: list[float] = Field(
-        description="Request start times relative to first event (first = 0)",
-    )
     time_scale: float = Field(
         default=1.0,
         gt=0,
-        description="Scale factor applied to relative timestamps",
+        description="Scale factor applied to relative timestamps from the dataset",
     )
 
     def __str__(self) -> str:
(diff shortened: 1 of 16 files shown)


src/guidellm/scheduler/strategies.py
```diff
@@ -197,15 +197,36 @@ def next_request_index(self) -> PositiveInt:
...
+
+        Default implementation returns empty settings so strategies that derive
+        timing at runtime can keep existing behavior.
+
+        :param request_index: Zero-based request index from producer ordering
+        :return: Scheduling settings attached to the request at enqueue
+        """
+        _ = request_index
+        return RequestSettings()
+
```
sjmonson: The `RequestSettings` object should come from the dataset; not the profile. Something like the finallizer interface here:

 https://github.com/vllm-project/guidellm/blob/12a836cd41f64c22d2774d71118f10a0c0f705dc/src/guidellm/data/finalizers.py#L23-L35

except instead of a float, the `DatasetFinalizer` creates a `RequestSettings` object from a column (or columns) of the dataset.

src/guidellm/scheduler/strategies.py
```diff
@@ -197,15 +197,36 @@ def next_request_index(self) -> PositiveInt:
...
     @abstractmethod
-    async def next_request_time(self, worker_index: NonNegativeInt) -> float:
+    async def next_request_time(
+        self,
+        worker_index: NonNegativeInt,
+        settings: RequestSettings | None = None,
```
sjmonson:
```suggestion
        settings: RequestSettings,
```

I think its probably safe to assume every request has a `RequestSettings` which will simplify usage.

sjmonson: Ah ok, my bad. I had forgotten that the request is not available when we dequeue a time. I think instead remove `RequestSettings` from the arguments of `next_request_time` and see the next comment on the `replay` profile.

src/guidellm/scheduler/strategies.py
```diff
@@ -698,29 +770,42 @@ def __str__(self) -> str:
...
 
-        return start_time + self.time_scale * self.relative_timestamps[idx - 1]
+    async def resolve_dequeued_target_start(
+        self,
+        worker_index: NonNegativeInt,
+        provisional_start: float,
+        settings: RequestSettings,
+    ) -> float:
+        _ = provisional_start
+        return await self.next_request_time(worker_index, settings)
```
sjmonson: Instead of having `resolve_dequeued_target_start` call `next_request_time`. Move the timestamp logic to `resolve_dequeued_target_start` and have `next_request_time` always emit `get_processes_start_time()`. That way the request schedules immediately and then defers to the trace time.

sjmonson (changes requested):

sjmonson (approved):
Looks good and seems to run with no issues!

sjmonson:
^ Sorry missclick

dbutenhof (approved):

src/guidellm/schemas/request.py
```diff
@@ -252,3 +253,10 @@ class GenerationRequest(StandardBaseModel):
         default_factory=UsageMetrics,
         description="Output statistics including counts, sizes, and durations.",
     )
+    settings: RequestSettings = Field(
+        default_factory=RequestSettings,
+        description=(
+            "Per-request scheduling metadata from the dataset finalizer, "
+            "for example trace replay relative timestamps."
+        ),
+    )
```
jaredoconnell, line 262: Is it worth having a new field for this, when the value added is already present in the columns field?

jaredoconnell (approved):
This seems like a good solution to get trace working for now. We'll need do a larger refactor later for more advanced trace formats that have more advanced requirements. One downside of this design is everything gets queued at the same time, which could be a problem for large datasets, but it would take some work to fix that without concurrency problems.

src/guidellm/schemas/request.py
sjmonson: Yes because the scheduler cannot read the columns; or more accurately it should not.

sjmonson:
Looks like e2e tests broke temporarily.

sjmonson:
@mergifyio queue

PR #747, opened 2026-05-28
feat: add automated docs deployment workflow

1195343015:
## Summary

Add automated CI workflows to deploy documentation to GitHub Pages via mike. Currently docs are only deployable manually, and the live site has not been updated since mid-April, causing stale and incorrect content (e.g. the developing page still reads "Developing for Speculators" instead of "Developing for GuideLLM").

## Details

- [x] Add `docs-deploy` job to `main.yml`, triggered on push to `main` when docs-related files change (`docs/**`, `mkdocs.yml`, `DEVELOPING.md`, etc.), deploying as the `main` version
- [x] Add `docs-deploy` job to `release.yml`, triggered on `v*.*.*` tags, deploying the version tag and updating the `stable` alias
- [x] Fix `site_url`, `repo_url`, and `edit_uri` in `mkdocs.yml` from `neuralmagic` to `vllm-project`

## Test Plan

- After merge, verify the `docs-deploy` job succeeds in the Main workflow
- Confirm a new commit from `github-actions[bot]` appears on the `gh-pages` branch
- Visit `https://vllm-project.github.io/guidellm/main/developer/developing/` and verify content matches the current `DEVELOPING.md`

## Related Issues

N/A

---

## Use of AI

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,6 +4,12 @@ on:
   push:
     branches:
       - main
+    paths:
+      - "docs/**"
+      - "mkdocs.yml"
+      - "DEVELOPING.md"
+      - "CONTRIBUTING.md"
+      - "CODE_OF_CONDUCT.md"
 
 jobs:
   tests:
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -22,3 +28,28 @@ jobs:
...
+      contents: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+      - name: Setup Python with UV
+        uses: ./.github/actions/python-uv
+        with:
+          python-version: "3.10"
```
dbutenhof: Python 3.10 goes end-of-life this year. I think we should be using 3.13 (the highest we currently support, and the lowest still getting bug fixes) except where there's a specific reason for something else (as in matrix testing).

(Yes, we have other 3.10 references in the GitHub workflows that ought to be updated -- but let's at least avoid adding more.)

.github/workflows/main.yml
```diff
@@ -4,6 +4,12 @@ on:
   push:
     branches:
       - main
+    paths:
+      - "docs/**"
+      - "mkdocs.yml"
+      - "DEVELOPING.md"
+      - "CONTRIBUTING.md"
+      - "CODE_OF_CONDUCT.md"
```
dbutenhof: This workflow also runs quality and smoke tests: it shouldn't be restricted to running on changes to "documentation" paths.

This would make sense if the workflow definition was specific to documentation publication, but it's not.

dbutenhof (changes requested):

dbutenhof (approved):
OK. Another option would have been to split off a separate documentation workflow, with the path patterns, that triggered on both push to main and release tag; but this should be fine.

sjmonson (approved):
Might move this to nightly later and like @dbutenhof said combine it into one shared file but for now looks good.

dbutenhof:
Ah -- I eagerly went to check the doc site: and of course this didn't update anything because no doc files changed. Next time. 😁

1195343015:
> Ah -- I eagerly went to check the doc site: and of course this didn't update anything because no doc files changed. Next time. 😁

Good catch! Opened #758 as a follow-up — it fixes the diff detection (git diff-tree HEAD → git diff event.before/sha, since merge commits don't carry file changes) and also includes a couple of doc fixes so the deploy actually triggers on merge this time.

PR #751, opened 2026-05-28
Ensure system env vars don't taint tests

jaredoconnell:
## Summary

I've had problems where env vars I set are tainting tests.

## Details

- The env var I set was to set the console log level to DEBUG, because I want to see backend errors. But that breaks a test. I had `GUIDELLM__LOGGING__CONSOLE_LOG_LEVEL=DEBUG`
- I applied this to all tests that could theoretically break due to env vars being set. 

## Test Plan

- Running tests to ensure the standard case didn't break
- Run tests with env vars set to ensure it still works. 

---

## Use of AI

### DIFF
--- a/tests/unit/test_settings.py
+++ b/tests/unit/test_settings.py
@@ -1,3 +1,5 @@
+import os
+
 import pytest
 
 from guidellm.settings import (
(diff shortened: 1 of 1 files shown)


dbutenhof (approved):
OK

sjmonson (approved):

PR #752, opened 2026-05-29
Fix TypeError when streaming delta has tool_calls=null

rgerganov:
## Summary

Some OpenAI-compatible servers emit ``"tool_calls": null`` inside streaming chat completion deltas. ``dict.get("tool_calls", [])`` only returns the default when the key is missing, so for an explicit ``null`` the iteration was attempted on ``None``, raising
``TypeError: 'NoneType' object is not iterable`` and erroring the request. Coerce ``None`` to an empty list before iterating.

---

## Use of AI

### DIFF
--- a/src/guidellm/backends/openai/request_handlers.py
+++ b/src/guidellm/backends/openai/request_handlers.py
@@ -726,7 +726,9 @@ class ChatCompletionsRequestHandler(TextCompletionsRequestHandler):
 
         # Accumulate streamed tool_calls deltas.  Each tool call may be split
         # across multiple chunks; we reassemble by ``index``.
-        for tc_delta in delta.get("tool_calls", []):
+        # ``tool_calls`` could be either missing or ``null`` in the delta
+        # (some OpenAI-compatible servers emit this), handle both cases
+        for tc_delta in delta.get("tool_calls") or []:
             self._accumulate_tool_call_delta(tc_delta)
             updated = True
 


src/guidellm/backends/openai/request_handlers.py
```diff
@@ -726,7 +726,9 @@ def add_streaming_line(self, line: str) -> int | None:
 
         # Accumulate streamed tool_calls deltas.  Each tool call may be split
         # across multiple chunks; we reassemble by ``index``.
-        for tc_delta in delta.get("tool_calls", []):
+        # ``tool_calls`` may be present but explicitly ``null`` in the delta
+        # (some OpenAI-compatible servers emit this), so coerce to a list.
```
dbutenhof: You're not "coercing" -- you're supplying a default when the `tool_calls` key exists but `get` returns a falsey value like `None` (or `[]`...).

I suspect in this context the difference between `null` and "falsey" doesn't really matter (e.g., `delta["tool_calls"] == 0`); but, technically, the "may be present but explicitly 'null'" is slightly misleading about what the code actually does.

dbutenhof (commented):
For reference, it'd be useful to know what server and model did this.

Given that this behavior has been seen, the change seems reasonable as long as we're OK with non-null falsey values silently slipping through this path -- but the comment is a bit misleading.

rgerganov:
Thanks for the review. This happens when running `guidellm` against model endpoint deployed on VCF Private AI Services. It is also a regression because we don't have such problem with the latest release (v0.6.0).

src/guidellm/backends/openai/request_handlers.py
rgerganov: I have fixed the comment.

dbutenhof (approved):
The "correctness" of this change hinges on whether we care about bizarre falsey values (`0`, `false`, `{}`, etc). However in context I'm not sure that's something we'd want to track as an "error" so I'm satisfied as long as nobody else objects.

"For the record" it might be helpful for the commit message to identify the server and model where you see this behavior.

sjmonson (approved):

Issue #724, opened 2026-05-08
Refactor CLI/Config in v0.7.0


sjmonson:
> [!NOTE]
> This proposal is human written.

The goal of this proposal is to define a new standard CLI and config (previously known as scenario) format which solves many usability issues and allows us to implement new features which are prohibitively difficult to add to the existing design.

## Usability Improvements

Here are a few issues that often crop up in the current design and how this proposal addresses them.

### Better data configuration

One critical limitation of the existing (<=v0.6.0) design for the `--data` argument is it is really hard to determine what a user intended when something is entered incorrectly. Most users of GuideLLM are likely familiar with some variant of this error (truncated to avoid taking up an entire page):

```python
ValueError: Data deserialization failed, likely because the input doesn't match any of the input formats. See the 15 error(s) that occurred while attempting to deserialize the data prompt_tokens=e,output_tokens=50:
  - Deserializer 'huggingface': (HFValidationError) Repo id must use alphanumeric chars, '-', '_' or '.'. The name cannot start or end with '-' or '.' and the maximum length is 96: 'prompt_tokens=100,output_tokens=50'.
...
```

This is because a data string like `prompt_tokens=100,output_tokens=50` contains no information on what kind of dataset it is. GuideLLM will attempt to parse it as every kind of dataset and if they all fail, it will return the error of every dataset type. With this proposal we have eliminated this ambiguity by adding a `type` field to configure explicitly which kind of dataset is being requested. This also has the added benefit of making it possible to have multiple dataset formats which closely match. Such as mookcake dataset which would normally conflict with plain jsonl.

### Built-in option documentation

The current GuideLLM CLI is very limited in what is documented, but internally every field has a description. Since the run CLI exposes users to the internal type lookup tables we can make a very intuitive help system for describing available options (note final format will differ):

```
$ guidellm describe backend openai_http
HTTP backend for OpenAI-compatible servers.

Supports OpenAI API, vLLM servers, and other compatible endpoints with
text/chat completions, streaming, authentication, and multimodal inputs.
Handles request formatting, response parsing, error handling, and token
usage tracking with flexible parameter customization.

Fields:
  - target: str
    Base URL of the OpenAI-compatible server
  - model: str | None
    Model identifier for generation requests
  - request_format: Literal["/v1/completions"] (default "/v1/completion")
    Request format for OpenAI-compatible server. 
```

### Config layering

[...]

PR #753, opened 2026-05-29
CLI profile refactor

dbutenhof:
## Summary

Continuing CLI refactoring: modify the Profile classes to use ProfileArgs instances built from `kind=<name>` payloads like backends and datasets.

## Details

Use CLI to construct a ProfileArgs instance, which is then used to build the Profile.

Like datasets and backends, we now identify a profile "kind", as `--profile kind=constant[,param=value...]`; directly specified parameters supercede inherited global CLI values like `--rate`.

Use more meaningful aliases for the global "rate" to match the previous internal names (e.g., "sweep_size"). This also exposes previously hidden parameters like the sweep's `strategy_type` to run poisson instead of constant async.

## Test Plan

Manually tested all profiles, including Replay.

## Related Issues

- Related to #724 

---

## Use of AI

Assisted-by: Cursor

---

# git log

commit 6b3d67569da7550f5b57b6d99dc6c995ccffee92
Date:   Tue May 26 15:24:51 2026 -0400

    CLI profile refactor
    
    Use CLI to construct a ProfileArgs instance, which is then used to build
    the Profile.
    
    Like datasets and backends, we now identify a profile "kind", as
    `--profile kind=constant[,param=value...]`; directly specified
    parameters supercede inherited global CLI values like `--rate`.
    
    Use more meaningful aliases for the global "rate" to match the previous
    internal names (e.g., "sweep_size"). This also exposes previously hidden
    parameters like the sweep's `strategy_type` to run poisson instead of
    constant async.
    
    Manually tested all profiles, including Replay.
    
    Assisted-by: Cursor

commit 9f24445ada63ac0d72671a56e75772e330800d67
Date:   Tue Jun 2 09:27:53 2026 -0400

    Review comments & cleanup
    
    - Remove Pydantic aspect of Profile class, holding a reference to ProfileArgs.
    - Remove constraints from ProfileArgs, to isolate construction and remove some
      report serialization issues.
    - Change ProfileArgs deserialization to "forbid" unknown keys. We have two
      classes of profile parameter that aren't "universal":
      - data and data_samples are used only by Replay: we remove these from other
        ProfileArgs -- this will be refactored later.
      - Although random_seed isn't used by all profiles, it's a global used
        throughout GuideLLM, so we promote this to the base ProfileArgs for
        convenience.
    
    Assisted-by: Cursor

commit f49f3b7545e7997d00be3740cf70b06fbef6f7d3
Date:   Tue Jun 2 12:13:06 2026 -0400

    Odd ..
    
    CI "pre-commit" check found something local pre-commit didn't.
    
commit 2c64d1ae3643219746e21ec011923410ef0a5d4e
Date:   Tue Jun 2 19:54:56 2026 -0400

    More review changes
    
    - move random_seed to Profile
    - formalize abstraction of Profile with ABC
    
commit 744b3ec6c6a2832e9f5c306027247864bdd8ea32
Date:   Tue Jun 2 19:59:17 2026 -0400

    Just in case the CI went crazy due to pre-commit updates ...
    
---------

Assisted-by: Cursor


src/guidellm/benchmark/profiles/profile.py
```diff
@@ -0,0 +1,331 @@
...
+    This class serves as a base for defining argument models used in the creation
+    of profile instances. It inherits from PydanticClassRegistryMixin to enable
+    automatic registration of subclasses, allowing for flexible and extensible
+    backend configurations.
+
+    :cvar schema_discriminator: Field name for polymorphic deserialization
+    """
+
+    model_config = ConfigDict(
+        extra="ignore",
```
sjmonson: Extras need to be forbidden on args classes otherwise we won't error on unknown fields.

```suggestion
        extra="forbid",
```

src/guidellm/benchmark/profiles/profile.py
```diff
@@ -0,0 +1,331 @@
...
+        "profile",
+        "synchronous",
+        "concurrent",
+        "throughput",
+        "async",
+        "constant",
+        "poisson",
+        "sweep",
+        "replay",
+    ] = Field(
```
sjmonson: Base class deliminator needs to be `str`.

```suggestion
    kind: str = Field(
```

src/guidellm/benchmark/profiles/profile.py
```diff
@@ -0,0 +1,331 @@
...
```
sjmonson: Base class deliminator needs to be `str`.


```suggestion
    kind: str = Field(
```

src/guidellm/benchmark/profiles/profile.py
```diff
@@ -0,0 +1,331 @@
...
+        "concurrent",
+        "throughput",
+        "async",
+        "constant",
+        "poisson",
+        "sweep",
+        "replay",
+    ],
+    "Profile type identifiers for polymorphic deserialization",
+]
```
sjmonson: Delete this as we are trying to move away from the hardcoded types list.

src/guidellm/benchmark/profiles/profile.py
```diff
@@ -0,0 +1,331 @@
...
+        "poisson",
+        "sweep",
+        "replay",
+    ] = Field(
+        description="Profile type discriminator for polymorphic serialization",
+    )
+    constraints: dict[str, Any | dict[str, Any] | ConstraintInitializer] | None = Field(
+        default=None,
+        description="Runtime constraints applied to strategy execution",
+    )
```
sjmonson: Adding `constraints` inside the `ProfileArgs` isn't going to work because constraints will have there own separate `ConstraintsArgs`. What I did for `DataLoader` was just make the `create` method (currently on `Profile` here) take instances of things created by other registries. For now I think here that just means moving this argument into `create`.

sjmonson (changes requested):
The way this is currently written each `ProfileArgs` is essentially a duplicate of its `Profile` without its business logic. That is going to be hard to maintain going forward because we will have to maintain identical fields between Args and Profiles. Instead I recommend looking at converting `Profiles` to a normal non-pydantic registry. See `BackendRegistry` or `DataLoaderRegistry` (after #754) for examples.

sjmonson (changes requested):
Forgot one thing. There is a bunch of migration code in the CLI run function:

https://github.com/vllm-project/guidellm/blob/8fdbd2862d1ca0405e947e40e55e21740c6fb02d/src/guidellm/cli/benchmark/run.py#L345-L364

If you add the following then you can update `BenchmarkGenerativeTextArgs` to store the profile args rather then profile name and avoid changing the existing CLI usage.

```python
kwargs["profile"] = {"kind": kwargs.pop("profile", sweep)}
```

src/guidellm/benchmark/profiles/profile.py
dbutenhof: Yeah, I realize that ... but I don't see a great alternative. For example, `synchronous` doesn't accept a "rate", and only `replay` wants `data`/`data_samples`.

To let Pydantic complain, I think we'd need to determine up front whether the profile we're building needs data/data_samples, and drop them if it doesn't... but that means determining the final class before calling the factory, and essentially hardcoding the list of "valid but non-universal" parameters we're willing to silently exclude.

I suppose that's effectively what the old `resolve_args` was doing. But that's still just silently ignoring what we don't want for each class.

Or, I suppose, we could add data/data_samples to `ProfileArgs` so they're not errors... but the idea of promoting parameters that *aren't* universal to the base class to avoid complaints isn't pleasant.

Basically, `Replay` is anomalous right now. We always have "data", and nobody else wants it.

src/guidellm/benchmark/profiles/profile.py
dbutenhof: That actually both solved & created some problems, but it makes sense. I added a `resolve_constraints` to deserialize, and passed that list into `Profile.create`.

dbutenhof (commented):
> The way this is currently written each ProfileArgs is essentially a duplicate of its Profile without its business logic.

Yeah; I'd started out turning Profile into a plain class, but ran into problems ... and without noticing I'd forgotten to remove the Pydantic registry mixin, I let Cursor try to fix them. Which it did by completing the Pydantic infrastructure. I let it stand to get some other things straightened out "first", and then never remembered to fix it.

> store the profile args rather then profile name and avoid changing the existing CLI usage.

Well, I started out with code to transparently promote a `str` to `{"kind": str}`, and decided it made more sense to require the kind to be explicit. I suppose it's not entirely similar to the new data deserializer since you *can* specify a profile without additional parameters.

Anyway, I spent today doing most of the rework, which got messier than I'd expected. I'll resume tomorrow morning, and hopefully have another tested commit up ...

src/guidellm/benchmark/profiles/profile.py
sjmonson: Needing `data` and `data_samples` is a temporary hack which I am hoping is resolved after #745. So I think for now adding it to `ProfileArgs` is fine.

dbutenhof: I decided to try something ugly and "less integrated", which seemed appropriate for this kind of hack. I added a "before" model validator that removes `data` and `data_samples` at the `ProfileArgs` level, and overrode/nullified that validator for `ReplayProfileArgs`.

I also realized I had to deal with `random_seed`, which is also only used/needed by a few but will always be available globally -- here it seemed more appropriate to move it into `ProfileArgs` rather than conditionally strip it out. `--random-seed` is always going to be there regardless, so we might as well just accept it.

src/guidellm/benchmark/profiles/profile.py
dbutenhof: I moved them to `Profile` instead of `ProfileArgs`, which solved some report-time serialization issues.

I'm also "compiling" constraints in entrypoints with a new `resolve_constraints` that'll fit the new model, and passing through the `dict[str, Constraint]`. (Actually last night I'd realized I should have shifted the `resolve_constraints` call out of `resolve_profile`, but I forgot this morning.)

src/guidellm/benchmark/profiles/profile.py
```diff
@@ -76,188 +71,111 @@ def __pydantic_schema_base_type__(cls) -> type[ProfileArgs]:
...
+    def registered_names(cls) -> tuple[str, ...]:
+        """
+        Get all registered names from the registry.
+        """
+        return tuple(cls.registry.keys() if cls.registry else [])
+
 
-    @field_serializer("constraints")
-    def _constraints_serializer(
+class Profile:
```
sjmonson: `class Profile(ABC)`? Some the other components use a `Protocol` for this, but since there is actual functionality in this class I think ABC is more approprate.

src/guidellm/benchmark/entrypoints.py
```diff
@@ -349,7 +374,7 @@ def _build_profile_from_args(
             "Profile arguments must be a string, dictionary, or ProfileArgs."
         )
     args = ProfileArgs.model_validate(profile_args)
```
sjmonson, line 375: I would still prefer you plumb this up to `BenchmarkGenerativeTextArgs` since the other refactor pieces did that but not a blocker since it could also fit into the CLI-side refactor.

src/guidellm/benchmark/entrypoints.py
```diff
@@ -330,7 +356,6 @@ def _build_profile_from_args(
     profile_args: dict[str, Any] = {
         "random_seed": random_seed,
```
sjmonson: Since so many different things need random seed, I have been treating it as an object instead of a config value (like how constraints work) and on the final PR wee will have a separate `RandomArgs` which configures the random number.

sjmonson (changes requested):
Few more nits. Also I forgot to mention earlier but in the original proposal I moved `warmup` and `cooldown` config into the `ProfileArgs` (see #724) since currently they are fed directly into the scheduler. We can tackle that later though if we just want to get this in to unblock other work.

src/guidellm/benchmark/entrypoints.py
dbutenhof: I do like that idea, and it nearly works, except for Replay, which needs `data_samples`, which it pulls out of the depths of the `request_loader` data iterator... which hasn't been built as the time we instantiate `BenchmarkGenerativeTextArgs`; meaning, I think, that I'd have to duplicate/simulate a lot of loader logic. (ReplayProfileArgs defaults it to -1, which isn't ideal if the loader *has* a value, but it ought to be working.)

I spent some more time experimenting this afternoon; but I haven't figured it out. Pydantic is apparently not mapping the raw `--data` JSON payload to Replay's `list[DataArgs]`. I'm not sure why that would be, but I guess even Pydantic has its limits. I could try `list[DataArgs | dict[str, str]]`, I suppose...

Ultimately, it's a matter of how far I chase the rabbit down the rabbit hole before giving up... I went further than I'd gone before, but Replay complicates the universe.

src/guidellm/benchmark/entrypoints.py
dbutenhof: Makes sense as a strategy -- but the profiles still need to access it, and I'm not sure there's a better way right now than just pulling it in.

sjmonson: `ProfileFactory.create(args, resolve_constraints(constraints), args.random_seed)`? And then down the line we can replace `args.random_seed` with however random is handled. Let me know if I am missing something.

src/guidellm/benchmark/entrypoints.py
sjmonson: Yeah I forgot you made it a global argument. Good to defer after we can drop that.

src/guidellm/benchmark/profiles/profile.py
dbutenhof: Oops ... good point. I missed that. It's easy to get lazy about the niceties. 😦

src/guidellm/benchmark/entrypoints.py
dbutenhof: Fine -- just move it out of `ProfileArgs` into the `Profile` subclasses for now. That's logical.

sjmonson (approved):

sjmonson (approved):

sjmonson:
@mergifyio queue

PR #758, opened 2026-05-30
fix: fix docs deploy trigger and minor doc issues

1195343015:
## Summary

The docs deployment workflow added in #747 uses `git diff-tree HEAD` to detect docs changes, which only checks the last commit and misses changes in merge commits. This prevents the first deployment from triggering after merge. Also fix a broken markdown link and stale logo URLs.

## Details

- [x] Use `git diff --name-only ${{ github.event.before }} ${{ github.sha }}` instead of `git diff-tree HEAD` so docs changes across the entire push are detected
- [x] Fix swapped markdown link syntax in DEVELOPING.md (`[jest-runner-groups](url)` instead of `[url](text)`)
- [x] Update `neuralmagic` -> `vllm-project` in docs/index.md logo image URLs

## Test Plan

- After merge, the Main workflow should detect the DEVELOPING.md and docs/index.md changes and run the docs-deploy job
- Verify https://vllm-project.github.io/guidellm/main/ shows the updated content

## Related Issues

- Follow-up to #747

---

## Use of AI

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Check docs changes
         id: docs-changed
         run: |
-          if git diff-tree --no-commit-id --name-only -r HEAD \
+          if git diff --name-only ${{ github.event.before }} ${{ github.sha }} \
             | grep -qE '^(docs/|mkdocs\.yml|DEVELOPING\.md|CONTRIBUTING\.md|CODE_OF_CONDUCT\.md)'; then
             echo "changed=true" >> $GITHUB_OUTPUT
           else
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
sjmonson, line 44: Looking at this again; I think it might be better move this workflow to `nightly` and drop this check.

sjmonson: Actually, we can just defer that change till later, since this will at least get modern docs published.

sjmonson (approved):

dbutenhof (approved):

dbutenhof:
Not sure that it's due to this PR (I haven't investigated any details), but the doc deployment tried and failed: https://github.com/vllm-project/guidellm/actions/runs/26829976219/job/79108799035

PR #759, opened 2026-06-01
Make synthetic_text output_tokens optional and improve CLI errors

rgerganov:
## Summary

Synthetic data generation already handled output_tokens being absent (samplers, features, and downstream max_tokens fallback all gated on None), but the schema declared it required. This forced users benchmarking endpoints with no output tokens (e.g. /v1/embeddings) to pass a meaningless value. Make the field default to None.

Also fix the misleading "Invalid value for --data: Field required" CLI error: the pydantic-to-click translation in `benchmark run` was dropping the nested field location and only reporting the top-level CLI option. Errors now include the full path (e.g. data[0].synthetic_text.prompt_tokens) and aggregate multiple validation errors into a single message.

Assisted-by: Cursor AI Claude Opus 4.7

---

## Use of AI

### DIFF
--- a/src/guidellm/cli/benchmark/run.py
+++ b/src/guidellm/cli/benchmark/run.py
@@ -408,11 +408,13 @@ def run(**kwargs):  # noqa: C901
             scenario=kwargs.pop("scenario", None), **kwargs
         )
     except ValidationError as err:
-        # Translate pydantic valdation error to click argument error
+        # Translate pydantic validation error to click argument error
         errs = err.errors(include_url=False, include_context=True, include_input=True)
-        param_name = "--" + str(errs[0]["loc"][0]).replace("_", "-")
+        first_loc = errs[0]["loc"]
+        top_field = str(first_loc[0]) if first_loc else ""
+        param_name = "--" + top_field.replace("_", "-")
         raise click.BadParameter(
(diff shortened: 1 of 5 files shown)


src/guidellm/cli/benchmark/run.py
```diff
@@ -28,6 +29,45 @@
...
+
+    :param errs: Pydantic error dicts as returned by ``ValidationError.errors()``
+    :return: Single error message; multiple errors are rendered as a bullet list
+    """
+    formatted = [_format_validation_error(e) for e in errs]
+    if len(formatted) == 1:
+        return formatted[0]
+    return "\n  - " + "\n  - ".join(formatted)
+
+
```
sjmonson: Yeah sorry this was missing. `main` is halfway through the refactor detailed in #724 so there are a bunch of rough edges. Can you move these functions to `src/guidellm/utils/cli.py`.

sjmonson (changes requested):
Thanks for the fixes! One minor nit.

src/guidellm/cli/benchmark/run.py
rgerganov: Fixed.

sjmonson (approved):

PR #765, opened 2026-06-03
Mask SecretStr api_key in model_dump so reports serialize

### LINKED ISSUE #764: SecretStr api_key crashes report serialization (no output when api_key is set)
## Bug Description

## Expected Behavior

## Steps to Reproduce

2. Run:
   ```bash
   guidellm benchmark run \
     --api-key sk-test-123 \
     --profile sweep \
     --max-requests 10 \
   ```

## Environment

- **Python Version:** 3.12.12

```
```

## Additional Context

MadCreeper:
## Summary

Backend args with an `api_key` (`SecretStr`) crash report serialization. `OpenAIHTTPBackend.info` returns `self._args.model_dump()` (python mode), leaving `api_key` as a raw `SecretStr` in `benchmark.config.backend`; the CSV/HTML writers `json.dumps` that dict and raise `TypeError: Object of type SecretStr is not JSON serializable`, so any run with an `api_key` set produces no reports. This adds a `field_serializer` on `OpenAIHTTPBackendArgs` that masks the key in every dump mode.

## Details

- [x] Add `field_serializer("api_key", when_used="always")` to `OpenAIHTTPBackendArgs`, masking the key (`**********`) in all dump modes.
- [x] Add unit tests asserting `model_dump()` / `model_dump_json()` are JSON-safe and masked, and that `get_secret_value()` (runtime auth) is unchanged.

## Test Plan

- `tox -e tests -- tests/unit/backends/openai/test_http.py`
- `tox -e lint-check && tox -e type-check`
- Manual: run a benchmark with `--api-key` set and `--outputs "benchmark.json,report.csv"`; confirm files are written and the key shows as `**********` (not the real value).

## Related Issues

- Resolves #764

---

- [x] "I certify that all code in this PR is my own, except as noted below." (AI assistance noted via the commit trailer.)

## Use of AI

### DIFF
--- a/src/guidellm/backends/openai/http.py
+++ b/src/guidellm/backends/openai/http.py
@@ -16,7 +16,14 @@ from collections.abc import AsyncIterator
 from typing import Any, Literal
 
 import httpx
-from pydantic import AliasChoices, Field, SecretStr, field_validator, model_validator
+from pydantic import (
+    AliasChoices,
+    Field,
+    SecretStr,
+    field_serializer,
+    field_validator,
+    model_validator,
+)
 
 from guidellm.backends.backend import Backend, BackendArgs
 from guidellm.backends.openai.request_handlers import (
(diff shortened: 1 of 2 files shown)


dbutenhof (approved):
So, effectively, you're converting `model_dump`, which leaves the Python object as value, to `model_dump_json`, which doesn't.

Looks like it ought to solve the problem, and indeed as the backend uses `.get_secret_value()` it shouldn't break the one case where we want the cleartext.

jaredoconnell (approved):

src/guidellm/backends/openai/http.py
```diff
@@ -148,6 +155,27 @@ class OpenAIHTTPBackendArgs(BackendArgs):
...
+        A bare ``SecretStr`` is not JSON-serializable under python-mode
+        ``model_dump()``. Because ``OpenAIHTTPBackend.info`` returns
+        ``model_dump()`` and that dict is embedded in each benchmark's
+        ``config.backend`` (which the CSV/HTML writers pass to ``json.dumps``),
+        an unmasked ``SecretStr`` raises ``TypeError`` and no report is written.
+        Masking, rather than exposing ``get_secret_value()``, matches Pydantic's
+        json-mode default and keeps real keys out of persisted reports. Runtime
+        auth is unaffected: it reads the live ``SecretStr`` via
+        ``get_secret_value()``, and workers receive the backend object by pickle,
+        not by ``model_dump()``.
```
sjmonson, line 172:
```suggestion
```

Can you drop this. Its a bit rambley and doesn't really help with documentation here.

sjmonson (changes requested):

PR #777, opened 2026-06-05
Add Mooncake trace format support

SkiHatDuckie:
## Summary
Add support for generating synthetic prompts from a Mooncake formatted trace file. This involves handling the *hash_ids* column, as well as validating that two different hash IDs do not represent the exact same sequence of tokens if they share the same previous ID (parent node).

Future PRs should create a common file/class for the similar functionality of `TraceMooncakeDatasetDeserializer` and `TraceSyntheticDatasetDeserializer`, as well as to make adding support for additional trace formats easier.

## Details
- Add `src/guidellm/data/deserializers/trace_mooncake.py`
  - `TraceMooncakeDataArgs`: Adds two new fields `hash_ids_column` and `hash_id_block_size`
  - `TraceMooncakeDatasetDeserializer`: Takes a processor, and a path to a .jsonl file, and returns a `datasets.Dataset` with the columns *("prompt", "prompt_token_count", "output_token_count", "hash_ids")*
- Register new classes with `DataArgs` and `DatasetDeserializerFactory`
- Add unit test coverage in `test_trace_mooncake.py`

## Test Plan
- `tox -e test-unit -- tests/unit/data/deserializers/test_trace_mooncake.py`
- `tox -e lint-check && tox -e type-check`

## Related Issues
- Addresses issue [#597](https://github.com/vllm-project/guidellm/issues/597) (1. Trace Parser)

---

## Use of AI

---

# git log

commit de90d4061affdeb4fe416e383ae817c2a4328a89
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Wed Jun 3 15:10:36 2026 -0400

    Add TraceMooncake objects
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit a58605e97d25a7134d000cd473384814983c4837
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Wed Jun 3 16:59:02 2026 -0400

    Populate trace_mooncake.py
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit fe5a3f6fab585a63c40b3713775c4e0b54a316e9
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Thu Jun 4 09:52:27 2026 -0400

    Update type hints
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit d770bfa61df4a3348a2405a5d5519ee580d495ac
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Fri Jun 5 13:33:27 2026 -0400

    Add test_trace_mooncake.py
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit b2f0cde5e346f84c67a0be19248473f2a59548eb
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Fri Jun 5 15:02:30 2026 -0400

    Add additional tests
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit 214fcef80bd9de50a2a3f25ac03c10f35bc5ff82
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Fri Jun 5 15:03:26 2026 -0400

    Remove unused imports
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit b9ffd891673935c7c323025dcefd6d4199ef13ab
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Fri Jun 5 15:24:01 2026 -0400

    Update type hints
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit 68c530f1d573413c6db7a553d62f265e7a54b6a2
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Fri Jun 5 15:32:41 2026 -0400

    Register Mooncake trace objects
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit a38db6304afbc1c0388a488743d181b92a5d8cf4
Author: SkiHatDuckie <63932363+SkiHatDuckie@users.noreply.github.com>
Date:   Sat Jun 6 21:36:22 2026 -0400

    Fix formatting in _generate_trace function
    
    Signed-off-by: SkiHatDuckie <63932363+SkiHatDuckie@users.noreply.github.com>
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit b26e3d702fd011d303dc446b6b845573f2acd5d0
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Mon Jun 8 13:12:37 2026 -0400

    Fix _generate_trace formatting
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit 3f67e4624c7c482c8e83c02f374455fee190c976
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Tue Jun 9 12:41:59 2026 -0400

    Add `kind` field to TraceMooncakeDataArgs
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit 2a9811029f395571714a7f83d12c3061cad3dd26
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Tue Jun 9 13:33:57 2026 -0400

    Rename parameters
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit 3021f8bfbcb7e36729250ddd81cdeba92188be6e
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Tue Jun 9 15:31:12 2026 -0400

    Update docstrings, DataArgs.register name
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit 8954956998a586cf35630d2aa7db107e0b3375ef
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Tue Jun 9 16:57:39 2026 -0400

    Remove raise in _generate_token_ids
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit a6bfb8b07f27c71684a37834acf9999e5f7ed406
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Wed Jun 10 11:53:04 2026 -0400

    Remove hash_ids column from generated Dataset
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit ca85a658573040d6000282ec4654d03d7224656c
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Wed Jun 10 17:05:08 2026 -0400

    Return IterableDataset to avoid pre-generating for entire trace
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit b1b85e5d47e0600cbeef3babad5f01f10a566fac
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Thu Jun 11 12:20:39 2026 -0400

    Move iteration_count increment to end of __iter__
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit f7ab87e35746ac526a9c2d6e9c5c3dd9de9a8a4f
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Thu Jun 11 12:25:10 2026 -0400

    Track row index state locally to examples interable
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit ac9233f401000e5032e249793943c16d7e4fade7
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Thu Jun 11 13:21:17 2026 -0400

    Increase `margin_of_safety` to minimize re-attempts
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit 7c7d1e499ef1431d1c0e18f657167877e46fedec
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Thu Jun 11 13:33:11 2026 -0400

    Fix: increment row_idx
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit 903a443d4ea9cb0a37d7e524fc3bf096c531a817
Author: SkiHatDuckie <63932363+SkiHatDuckie@users.noreply.github.com>
Date:   Thu Jun 11 14:26:04 2026 -0400

    Change `kind` to match registered name
    
    Signed-off-by: SkiHatDuckie <63932363+SkiHatDuckie@users.noreply.github.com>

commit 2100dbd78f76b3fc41a98bf5e969b2dfd8cd87ef
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Thu Jun 11 15:50:27 2026 -0400

    Implement  methods required by _BaseExamplesIterable
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

commit 6472f554bbd03992cb6c9760f44e4ea6c39ba466
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Fri Jun 12 09:15:58 2026 -0400

    Satisfy linting
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

---------

Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>
Signed-off-by: SkiHatDuckie <63932363+SkiHatDuckie@users.noreply.github.com>

### DIFF
--- a/src/guidellm/data/deserializers/trace_mooncake.py
+++ b/src/guidellm/data/deserializers/trace_mooncake.py
@@ -9,6 +9,7 @@ benchmarks.
 
 import math
 from collections.abc import Callable
+from typing import Any
 
 from datasets import Dataset
 from faker import Faker
@@ -60,7 +61,7 @@ def _validate_row(row: dict, config: TraceMooncakeDataArgs) -> None:
         )
 
 
-def _is_in_table(hash_id_table: list[list[int]], hash_id: int) -> bool:
+def _is_in_table(hash_id_table: list[Any], hash_id: int) -> bool:
     return (
         hash_id < len(hash_id_table)
         and hash_id >= 0
@@ -68,7 +69,7 @@ def _is_in_table(hash_id_table: list[list[int]], hash_id: int) -> bool:
     )
 
 
-def _resize_to_hold_id(hash_id_table: list[list[int]], hash_id: int) -> None:
+def _resize_to_hold_id(hash_id_table: list[Any], hash_id: int) -> None:
     num_new_entries = hash_id - (len(hash_id_table) - 1)
     hash_id_table.extend(None for _ in range(num_new_entries))
 
@@ -176,8 +177,8 @@ class TraceMooncakeDatasetDeserializer(DatasetDeserializer):
         processor = processor_factory()
         faker = Faker()
         faker.seed_instance(random_seed)
-        hash_id_table = []
-        sibling_token_blocks = {}
+        hash_id_table: list[Any] = []
+        sibling_token_blocks: dict[Any, list[list[int]]] = {}
         prompts = []
         for row in rows:
             _validate_row(row, config)
--- a/src/guidellm/data/deserializers/trace_mooncake.py
+++ b/src/guidellm/data/deserializers/trace_mooncake.py
@@ -64,8 +64,9 @@ def _calculate_required_prompt_tokens(
     """Returns the number of prompt tokens needed to satisfy the row input length.
     This will be less than the block_size if the input length is not divisible by it
     and `hash_id` is the final ID for the row."""
-    if row[config.hash_ids_column][-1] == hash_id:
-        return row[config.prompt_tokens_column] % config.hash_id_block_size
+    remainder = row[config.prompt_tokens_column] % config.hash_id_block_size
+    if row[config.hash_ids_column][-1] == hash_id and remainder != 0:
+        return remainder
     return config.hash_id_block_size
 
 
--- a/src/guidellm/data/deserializers/trace_mooncake.py
+++ b/src/guidellm/data/deserializers/trace_mooncake.py
@@ -81,7 +81,9 @@ def _generate_token_ids(
     faker: Faker,
 ) -> list[int]:
     """Generate `token_count` synthetic token ids for trace prompt construction."""
-    margin_of_safety = 2
+    # Ideally, `margin_of_safety` should be set to slighty more than
+    # the average number of characters used by tokenizers to form one token.
+    margin_of_safety = 8
     attempt = 0
     while True:
         attempt += 1
--- a/src/guidellm/data/deserializers/trace_mooncake.py
+++ b/src/guidellm/data/deserializers/trace_mooncake.py
@@ -6,6 +6,7 @@ Reads a trace file (timestamp, input_length, output_length, hash_ids) and yields
 row per line with a synthetic prompt matching the requested input_length for replay
 benchmarks.
 """
+
 from __future__ import annotations
 
 import dataclasses
@@ -14,11 +15,11 @@ from collections.abc import Callable, Iterable
 from pathlib import Path
 from typing import Any, Literal
 
+import numpy as np
 from datasets import Dataset, DatasetInfo, Features, IterableDataset, List, Value
 from datasets.exceptions import DatasetGenerationError
 from datasets.iterable_dataset import _BaseExamplesIterable
 from faker import Faker
-import numpy as np
 from pydantic import Field
 from transformers import PreTrainedTokenizerBase
 
--- a/src/guidellm/data/deserializers/trace_mooncake.py
+++ b/src/guidellm/data/deserializers/trace_mooncake.py
@@ -6,6 +6,7 @@ Reads a trace file (timestamp, input_length, output_length, hash_ids) and yields
 row per line with a synthetic prompt matching the requested input_length for replay
 benchmarks.
 """
+from __future__ import annotations
 
 import dataclasses
 import math
@@ -17,6 +18,7 @@ from datasets import Dataset, DatasetInfo, Features, IterableDataset, List, Valu
 from datasets.exceptions import DatasetGenerationError
 from datasets.iterable_dataset import _BaseExamplesIterable
 from faker import Faker
+import numpy as np
 from pydantic import Field
 from transformers import PreTrainedTokenizerBase
 
@@ -266,6 +268,22 @@ class _TraceMooncakeExamplesIterable(_BaseExamplesIterable):
     @property
     def num_shards(self) -> int:
         return 1
+
+    def shuffle_data_sources(
+        self,
+        generator: np.random.Generator,  # noqa: ARG002
+    ) -> _TraceMooncakeExamplesIterable:
+        """Returns self as sharding is not implemented yet."""
+        return self
+
+    def shard_data_sources(
+        self,
+        num_shards: int,  # noqa: ARG002
+        index: int,  # noqa: ARG002
+        contiguous: bool = True,  # noqa: ARG002
+    ) -> _TraceMooncakeExamplesIterable:
+        """Returns self as sharding is not implemented yet."""
+        return self
 
     def load_state_dict(self, state_dict: dict) -> None:
         """Load the state from a state dict."""
--- a/src/guidellm/data/deserializers/trace_mooncake.py
+++ b/src/guidellm/data/deserializers/trace_mooncake.py
@@ -235,15 +235,15 @@ class _TraceMooncakeExamplesIterable(_BaseExamplesIterable):
                     )
                     sibling_token_blocks[prev_id].append(hash_id_table[hash_id])
             prompt = _create_prompt_from_hash_ids(ids, hash_id_table, self.processor)
-            self.iteration_count += 1
             yield (
-                self.iteration_count - 1,
+                self.iteration_count,
                 {
                     "prompt_tokens_count": row[self.config.prompt_tokens_column],
                     "output_tokens_count": row[self.config.output_tokens_column],
                     "prompt": prompt,
                 },
             )
+            self.iteration_count += 1
 
     @property
     def is_typed(self) -> bool:
--- a/src/guidellm/data/deserializers/trace_mooncake.py
+++ b/src/guidellm/data/deserializers/trace_mooncake.py
@@ -9,12 +9,13 @@ benchmarks.
 
 import dataclasses
 import math
-from collections.abc import Callable
+from collections.abc import Callable, Iterable
 from pathlib import Path
 from typing import Any, Literal
 
-from datasets import Dataset, List, Value
+from datasets import Dataset, DatasetInfo, Features, IterableDataset, List, Value
 from datasets.exceptions import DatasetGenerationError
+from datasets.iterable_dataset import _BaseExamplesIterable
 from faker import Faker
 from pydantic import Field
 from transformers import PreTrainedTokenizerBase
@@ -107,7 +108,7 @@ def _create_distinct_token_block(
         if token_ids not in sibling_token_blocks:
             return token_ids
         attempt += 1
-    raise DataNotSupportedError(
+    raise ValueError(
         f"Failed to generate distinct synthetic token block after {attempt} attempts"
     )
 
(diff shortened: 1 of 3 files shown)


src/guidellm/data/deserializers/trace_mooncake.py
```diff
@@ -0,0 +1,245 @@
...
+    This will be less than the block_size if the input length is not divisible by it
+    and `hash_id` is the final ID for the row."""
+    remainder = row[config.prompt_tokens_column] % config.hash_id_block_size
+    if row[config.hash_ids_column][-1] == hash_id and remainder != 0:
+        return remainder
+    return config.hash_id_block_size
+
+
+def _generate_token_ids(
+    amount: int, processor: PreTrainedTokenizerBase, faker: Faker, max_attempts: int = 8
```
dbutenhof: "amount" is a rather odd name--that usually implies measured rather than counted quantities. How about "count" -- or "token_count" would be even more clear.

src/guidellm/data/deserializers/trace_mooncake.py
```diff
@@ -0,0 +1,245 @@
...
+            return token_ids[:amount]
+    raise DataNotSupportedError(
+        "Failed to generate enough synthetic prompt tokens for "
+        f"{amount} tokens after {attempt} attempts"
+    )
+
+
+def _create_distinct_token_block(
+    block_size: int,
+    exhausted_token_blocks: list[list[int]],
```
dbutenhof: Another rather odd name. "Exhausted" implies we've already tried them, but I don't think that's quite the sense you want here.

You use "sibling nodes" in the docstring, which might be a better name. (Or sibling blocks, since you're referring to block size: or "prompts" or "input token blocks", since that's really what we're talking about.)

src/guidellm/data/deserializers/trace_mooncake.py
```diff
@@ -0,0 +1,245 @@
...
+    processor: PreTrainedTokenizerBase,
+    faker: Faker,
+    max_distinctness_attempts: int = 20,
```
dbutenhof: Just "max_attempts" would probably be better here. "Distinct" (although I'm not sure that's the best term, either) is already in the function name.

VincentG1234:
Hello, thanks for working on this.

It looks like the branch may have been merged with main instead of rebased on top of it. The current diff contains many unrelated commits/changes, which makes the Mooncake-specific changes difficult to review. Would it be possible to rebase or clean the branch so the PR only contains the trace_mooncake changes?

I also tried the PR locally before the merge and hit the following validation error:

```bash
guidellm benchmark run \
  --profile kind=replay \
  --data "kind=trace_mooncake,path=trace.jsonl" \
  --target http://<url of your endpoint> \
  --model <model-name>
```
Error:
```terminal
pydantic_core._pydantic_core.ValidationError: 1 validation error for BenchmarkGenerativeTextArgs
data.0.trace_mooncake.kind
  Input should be 'trace_synthetic' [type=literal_error, input_value='trace_mooncake', input_type=str]
```

It looks like `trace_mooncake` is not currently understood by the CLI for replay data. The command fails at Pydantic validation because the schema expects `trace_synthetic` only.

SkiHatDuckie:
Branch has now been cleaned up. I was still working on getting guidellm/vllm to run on my current machine, so the initial commits were only tested via unit tests. I'll get to work on fixing the validation shortly.

dbutenhof (commented):
A few minor comments to start with (which I realize I've had pending for several days after a first scan, and might as well post!)

src/guidellm/data/deserializers/trace_mooncake.py
SkiHatDuckie: Renamed to "token_count"

src/guidellm/data/deserializers/trace_mooncake.py
SkiHatDuckie: I decided on switching to "sibling_token_blocks" since that maintains the relationship between the newly generated and previously generated token blocks.

src/guidellm/data/deserializers/trace_mooncake.py
SkiHatDuckie: Renamed to "max_attempts"

src/guidellm/data/deserializers/trace_mooncake.py
```diff
@@ -0,0 +1,249 @@
...
+    DatasetDeserializerFactory,
+)
+from guidellm.data.deserializers.trace_synthetic import TraceSyntheticDataArgs
+from guidellm.data.schemas import DataArgs
+from guidellm.utils.trace_io import load_trace_rows
+
+__all__ = ["TraceMooncakeDataArgs", "TraceMooncakeDatasetDeserializer"]
+
+
+@DataArgs.register("trace_mooncake")
```
sjmonson: Maybe just `mooncake` rather than `trace_mooncake`.

src/guidellm/data/deserializers/trace_mooncake.py
```diff
@@ -0,0 +1,249 @@
...
+        # The Faker.text() can only generate text of at least 5 characters.
+        num_chars = max(token_count * margin_of_safety * attempt, 5)
+        text = faker.text(max_nb_chars=num_chars)
+        token_ids = processor.encode(text)
+        if len(token_ids) >= token_count:
+            return token_ids[:token_count]
+    raise DataNotSupportedError(
+        "Failed to generate enough synthetic prompt tokens for "
+        f"{token_count} tokens after {attempt} attempts"
+    )
```
sjmonson: I'm a little concerned about throwing this here. `DataNotSupportedError` is designed to fail the run if it is thrown during deserialization. Not 100% sure here but I think it will run fail the row here rather then the dataset. I am not sure we want that behavior though. Might be better to keep trying until we generate the right "amount". That being said hopefully we never hit this on a real dataset.

src/guidellm/data/deserializers/trace_mooncake.py
```diff
@@ -0,0 +1,249 @@
...
+    max_attempts: int = 20,
+) -> list[int]:
+    """Constructs a new token block of `block_size` that does not appear in
+    `sibling_nodes`."""
+    attempt = 0
+    while attempt < max_attempts:
+        token_ids = _generate_token_ids(block_size, processor, faker)
+        if token_ids not in sibling_token_blocks:
+            return token_ids
+        attempt += 1
```
sjmonson, line 115: Did you do any tests for how often we see re-attempts?

src/guidellm/data/deserializers/trace_mooncake.py
```diff
@@ -0,0 +1,249 @@
...
+    processor: PreTrainedTokenizerBase,
+    faker: Faker,
+    max_attempts: int = 20,
+) -> list[int]:
+    """Constructs a new token block of `block_size` that does not appear in
+    `sibling_nodes`."""
+    attempt = 0
+    while attempt < max_attempts:
+        token_ids = _generate_token_ids(block_size, processor, faker)
+        if token_ids not in sibling_token_blocks:
```
sjmonson, line 113: If needed, could probably speed this up by computing a hash for each block (on creation) and checking against that instead. Probably not our biggest bottleneck though.

src/guidellm/data/deserializers/trace_mooncake.py
```diff
@@ -0,0 +1,249 @@
...
+                    _resize_to_hold_id(hash_id_table, hash_id)
+                    prev_id = None if idx == 0 else ids[idx - 1]
+                    num_tokens = _calculate_required_prompt_tokens(row, config, hash_id)
+                    sibling_token_blocks.setdefault(prev_id, [])
+                    hash_id_table[hash_id] = _create_distinct_token_block(
+                        num_tokens, sibling_token_blocks[prev_id], processor, faker
+                    )
+                    sibling_token_blocks[prev_id].append(hash_id_table[hash_id])
+            prompt = _create_prompt_from_hash_ids(ids, hash_id_table, processor)
+            prompts.append(prompt)
```
sjmonson: Don't pre-generate the whole dataset. Most of the time users only run a subset of rows and we cannot always determine the number ahead of time. Take a look at `synthetic_text` but basically you need a custom [`_BaseExamplesIterable`](https://github.com/huggingface/datasets/blob/68ac1a9b35869aadb7ae29857be46f6489d6150d/src/datasets/iterable_dataset.py#L202) which yields each row. Then a custom `IterableDataset` which wraps your iterator. Which is what you will return below.

sjmonson (changes requested):
Some nits, we also discussed merging this with "trace_synthetic" since they are mostly the same. Lets do that in a followup after this merges.

src/guidellm/data/deserializers/trace_mooncake.py
SkiHatDuckie: Currently no. I'll have to figure out a way to check that.

SkiHatDuckie: Doing some tests with a mock processor that only outputs tokens of value 0 or 1 on the conversations dataset: Exactly 1 re-attempt out of 12k rows. While real tokenizers are not gonna be total randomness, I don't see re-attempts even on large datasets being that severe of a concern for most models.

src/guidellm/data/deserializers/trace_mooncake.py
```diff
@@ -180,6 +181,123 @@ def _validate_row(row: dict, config: TraceMooncakeDataArgs) -> None:
...
+                        num_tokens,
+                        sibling_token_blocks[prev_id],
+                        self.processor,
+                        self.faker,
+                    )
+                    sibling_token_blocks[prev_id].append(hash_id_table[hash_id])
+            prompt = _create_prompt_from_hash_ids(ids, hash_id_table, self.processor)
+            self.iteration_count += 1
+            yield (
+                self.iteration_count - 1,
```
sjmonson: Not a huge thing, but you can `self.iteration_count += 1` after the yield and avoid the subtraction.

src/guidellm/data/deserializers/trace_mooncake.py
```diff
@@ -180,6 +181,123 @@ def _validate_row(row: dict, config: TraceMooncakeDataArgs) -> None:
...
+            Column(config.timestamp_column, Value("float")),
+            required_columns=[
+                Column(config.prompt_tokens_column, Value("int32")),
+                Column(config.output_tokens_column, Value("int32")),
+                Column(config.hash_ids_column, List(Value("int32"))),
+            ],
+        )
+        for row in self.trace_rows:
+            _validate_row(row, self.config)
+        self.iteration_count = 0
```
sjmonson, line 215: For `synthetic_text`, `iteration_count` is the number of times the iterable was created, not the number of rows created. Each time we run a separate sub-benchmark (E.g. multi-rate concurrency) the iterable is recreated. For rows, don't maintain that state globally; the only reason we do so in synthetic text is because we never want to generate the same prompt twice even across sub-benchmarks.

It unlikely to matter much here since you will likely only use this data with `replay` and not run multiple sub-benchmarks; but is technically incorrect. Any state data storing the current index should be contained in `__iter__`.

sjmonson (changes requested):

src/guidellm/data/deserializers/trace_mooncake.py
SkiHatDuckie: I've done some benchmarking on this and haven't been able to find a method of hashing that leads to significant boosts in performance. What *did* lead to a performance boost was increasing the `margin_of_safety` from 2 to 8 when generating a token block (roughly twice as fast thanks to fewer re-attempts).

sjmonson: This can be very model dependent since different models use different tokenizers, but I agree 8 sounds like a better default than 2.

src/guidellm/data/deserializers/trace_mooncake.py
```diff
@@ -0,0 +1,330 @@
...
+
+__all__ = ["TraceMooncakeDataArgs", "TraceMooncakeDatasetDeserializer"]
+
+
+@DataArgs.register("mooncake")
+class TraceMooncakeDataArgs(TraceSyntheticDataArgs):
+    kind: Literal["trace_mooncake"] = Field(  # type: ignore[assignment]
+        default="trace_mooncake",
+        description="Type identifier for the trace Mooncake dataset deserializer.",
+    )
```
sjmonson: `kind` needs to match registered name:

```suggestion
    kind: Literal["mooncake"] = Field(  # type: ignore[assignment]
        default="mooncake",
        description="Type identifier for the trace Mooncake dataset deserializer.",
    )
```

sjmonson (changes requested):

sjmonson (changes requested):
Still seeing failures:

```
Skipping data row due to error: <class 'guidellm.data.deserializers.trace_mooncake._TraceMooncakeExamplesIterable'> doesn't implement shard_data_sources yet.
```

Might be a good idea to write an integration test that loads a 5ish row dataset and tries to iterate through it.

SkiHatDuckie:
> Still seeing failures:
> 
> ```
> Skipping data row due to error: <class 'guidellm.data.deserializers.trace_mooncake._TraceMooncakeExamplesIterable'> doesn't implement shard_data_sources yet.
> ```
> 
> Might be a good idea to write an integration test that loads a 5ish row dataset and tries to iterate through it.

Should be fixed with the latest commit. I'm thinking of adding the integration tests during the refactoring/merging together of TraceSynthetic and TraceMooncake (separate PR), as I fear adding the tests now will lead to them being immediately obsoleted.

sjmonson (dismissed):
Needs a linting fix but otherwise looks good.

src/guidellm/data/deserializers/trace_mooncake.py
```diff
@@ -0,0 +1,348 @@
...
+    return (
+        hash_id < len(hash_id_table)
+        and hash_id >= 0
+        and hash_id_table[hash_id] is not None
+    )
+
+
+def _resize_to_hold_id(hash_id_table: list[Any], hash_id: int) -> None:
+    num_new_entries = hash_id - (len(hash_id_table) - 1)
+    hash_id_table.extend(None for _ in range(num_new_entries))
```
jaredoconnell, line 66: Since this is extending a list for every value, if the values in the dataset aren't sequential then this could be a very large number. Options are to ignore this because this is unlikely in practice, to warn for large numbers, or to make this a sparse list.

src/guidellm/data/deserializers/trace_mooncake.py
```diff
@@ -0,0 +1,348 @@
...
+    token_count: int,
+    processor: PreTrainedTokenizerBase,
+    faker: Faker,
+) -> list[int]:
+    """Generate `token_count` synthetic token ids for trace prompt construction."""
+    # Ideally, `margin_of_safety` should be set to slighty more than
+    # the average number of characters used by tokenizers to form one token.
+    margin_of_safety = 8
+    attempt = 0
+    while True:
```
jaredoconnell, line 91: It might make sense to bound the infinite loop. Especially in case there is an encoding problem. But I see the same pattern is used in synthetic.py, and it hasn't been a problem there. I can only imagine this being a problem if there's a problem with the faker or memory gets corrupted or something.

jaredoconnell (dismissed):
Looks good to me, except we could use some documentation for this format in docs/guides/datasets.md. I also added a few comments that are not blockers.

src/guidellm/data/deserializers/trace_mooncake.py
dbutenhof, line 1: > - src/guidellm/data/deserializers/trace_mooncake.py:
    1 × I001 (unsorted-imports)

`ruff check --fix src` should take care of this.

dbutenhof (dismissed):
The CI is unhappy with your imports, so you're going to need one more commit.

src/guidellm/data/deserializers/trace_mooncake.py
SkiHatDuckie: In prior versions of this PR, this was bounded followed by raising if it somehow exceeded some max number of attempts. The mock tokenizer used to verify that it threw encoded the entire string into one token.

src/guidellm/data/deserializers/trace_mooncake.py
SkiHatDuckie: When I was originally learning how traces worked, I was worried about the same thing. I'll keep a note of this (+ updating the docs) for future contributions.

dbutenhof (approved):

jaredoconnell (approved):

PR #783, opened 2026-06-09
Fix report serialization for nested paths

### LINKED ISSUE #764: SecretStr api_key crashes report serialization (no output when api_key is set)

### LINKED ISSUE #781: TypeError: Object of type PosixPath is not JSON serializable when saving benchmark report with --output-dir
## Description

When running `guidellm benchmark run` with `--output-dir` and `--outputs json`, the benchmark completes successfully but crashes during report serialization:

```
TypeError: Object of type PosixPath is not JSON serializable
```

## Root Cause

In `src/guidellm/benchmark/schemas/generative/report.py:105`, `save_file()` calls:

```python
model_dict = self.model_dump()       # mode='python' (default)
save_str = json.dumps(model_dict)     # TypeError
```

`model_dump()` defaults to `mode='python'`, which preserves `PosixPath` objects. While `BenchmarkGenerativeTextArgs` has a `@field_serializer("output_dir")` that converts `Path → str`, that serializer only activates during `model_dump_json()` or `model_dump(mode='json')`, **not** the default `model_dump()`.

The `PosixPath` originates from `--output-dir`, which click produces via `click.Path(path_type=Path)`.

## Reproduction

```bash
guidellm benchmark run \
  --profile "kind=sweep" \
  --data '{"kind": "json_file", "path": "sharegpt_flat.jsonl", "load_kwargs": {"split": "train"}}' \
  --max-seconds 60 \
  --max-requests 5000 \
  --output-dir ./benchmark_results \
  --outputs json,csv,html
```

## Stack Trace

```
File "guidellm/benchmark/outputs/serialized.py", line 82, in finalize
    return report.save_file(output_path)
File "guidellm/benchmark/schemas/generative/report.py", line 106, in save_file
    save_str = json.dumps(model_dict)
File "json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
TypeError: Object of type PosixPath is not JSON serializable
```

## Suggested Fix

Change `report.py:105` to use JSON serialization mode:

```python
model_dict = self.model_dump(mode='json')
```

This one-word change activates `@field_serializer` decorators and converts `Path → str` before `json.dumps()` sees it. The same fix should be applied to the YAML branch on line 108 as well.

**Environment:**
- guidellm version: latest (installed via pip)
- Python 3.12
- Linux

sjh9714:
## Summary

Fixes #781.

Benchmark report persistence currently dumps the report in Python mode, so nested `Path` values in benchmark args can remain as `PosixPath` objects and break JSON output or emit unsafe YAML tags. This switches report persistence to Pydantic's JSON mode so nested path values are converted before writing JSON/YAML files.

## Details

- Use `model_dump(mode="json")` before serializing benchmark reports.
- Add regression coverage for `json_file` data paths and `output_dir` path values in JSON and YAML report output.

## Test Plan

- `uv run pytest tests/unit/benchmark/test_serialized_output.py -k nested_paths`
- `uv run pytest tests/unit/benchmark/test_serialized_output.py`
- `uv run pytest tests/unit/benchmark`
- `uv run ruff check src/guidellm/benchmark/schemas/generative/report.py tests/unit/benchmark/test_serialized_output.py && uv run ruff format --check src/guidellm/benchmark/schemas/generative/report.py tests/unit/benchmark/test_serialized_output.py`
- Attempted `uv run tox -e test-unit -- tests/unit/benchmark/test_serialized_output.py -k nested_paths` (failed before test collection in this local checkout because `.tox/test-unit/bin/python` had no `pytest` module installed).

## Related Issues

- Fixes #781
- Fixes #764
- Closes #765 

---

## Use of AI

Note: I used Codex while preparing this change, reviewed the final diff, and ran the listed checks locally.

---

# git log

commit 37777334d7938acaca78cf15a229ad2586fa35e2
Author: jinhyuk9714 <jinhyuk9714@gmail.com>
Date:   Wed Jun 10 00:15:40 2026 +0900

    Fix report serialization for nested paths
    
    Assisted-by: Codex
    Signed-off-by: jinhyuk9714 <jinhyuk9714@gmail.com>

---------

Assisted-by: Codex
Signed-off-by: jinhyuk9714 <jinhyuk9714@gmail.com>

### DIFF
--- a/src/guidellm/benchmark/schemas/generative/report.py
+++ b/src/guidellm/benchmark/schemas/generative/report.py
@@ -101,7 +101,7 @@ class GenerativeBenchmarksReport(StandardBaseModel):
         )
         file_path.parent.mkdir(parents=True, exist_ok=True)
         file_type = type_ or file_path.suffix.lower()[1:]
-        model_dict = self.model_dump()
+        model_dict = self.model_dump(mode="json")
         if file_type == "json":
             save_str = json.dumps(model_dict)
         elif file_type in ["yaml", "yml"]:
--- a/tests/unit/benchmark/test_serialized_output.py
+++ b/tests/unit/benchmark/test_serialized_output.py
@@ -37,6 +37,38 @@ def minimal_report() -> GenerativeBenchmarksReport:
     return GenerativeBenchmarksReport(args=args)
 
 
+@pytest.fixture
+def file_path_report(tmp_path: Path) -> GenerativeBenchmarksReport:
+    """
+    Create a report whose benchmark args contain nested Path values.
+
+    ## WRITTEN BY AI ##
+    """
+    args = BenchmarkGenerativeTextArgs.model_validate(
+        {
+            "backend_kwargs": {
+                "kind": "openai_http",
+                "target": "http://localhost:8000/v1",
+                "model": "test-model",
+            },
+            "data": [
+                {
+                    "kind": "json_file",
+                    "path": tmp_path / "data.jsonl",
+                    "load_kwargs": {"split": "train"},
+                }
+            ],
+            "tokenizer": {"kind": "huggingface_auto", "model": "test-model"},
+            "data_column_mapper": {"kind": "generative_column_mapper"},
+            "data_preprocessors": [],
+            "data_finalizer": {"kind": "generative"},
+            "data_loader": {"kind": "pytorch"},
+            "output_dir": tmp_path / "outputs",
+        }
+    )
+    return GenerativeBenchmarksReport(args=args)
+
+
 class TestResolveFormatType:
     """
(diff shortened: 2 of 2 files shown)


sjmonson (approved):
Reading through the [documentation](https://pydantic.dev/docs/validation/latest/api/pydantic/base_model/#pydantic.BaseModel.model_dump(mode)) it seems like this should work.

sjmonson:
@mergifyio queue

PR #794, opened 2026-06-12
Infer audio encoding format from source instead of defaulting to MP3

### LINKED ISSUE #623: Avoid default MP3 transcoding for transcription benchmarks
### Problem Statement

When benchmarking transcription/translation workloads, GuideLLM currently transcodes input audio to MP3 by default through the `encode_media` preprocessing path (`encode_audio(audio_format="mp3")` default). This changes the original dataset audio format before request submission.

### Proposed Solution

Change default audio encoding behavior to avoid forced MP3 transcoding.

Suggested behavior:
1. If `audio_format` is explicitly provided by the user, use it.
2. Otherwise, infer format from source metadata/path (file suffix, URL suffix, or dataset-provided format).
3. If format cannot be inferred, default to `wav` (safe/common fallback), not `mp3`.

Optional:
- add a warning/log when fallback-to-WAV is used due to missing format metadata,
- document this behavior in benchmarking docs and CLI examples.

### Alternatives Considered

Always force WAV instead of MP3: better than MP3 for fidelity, but still ignores source format unnecessarily.

### Usage Examples

```markdown

```

### Additional Context

_No response_

jaredoconnell:
## Summary

Uses the source's audio format rather than defaulting to MP3

## Details

- If no format is provided by the user, and it cannot be inferred, it defaults to WAV and warns the user.
- Simply uses the format read from the dataset if none is provided.
- Includes a fix for a missing super call in the response handler.

## Test Plan

Make sure you install vLLM with the audio features.

You can run this with multiple datasets. Some options include:
```bash
guidellm benchmark run   --target "http://localhost:8000"   --request-type /v1/audio/transcriptions   --profile kind=synchronous   --max-requests 10   --data '{"kind": "huggingface", "source": "google/fleurs", "load_kwargs": {"name": "en_us", "split": "test"}}'   --data-column-mapper '{"column_mappings": {"audio_column": "audio"}}'   --disable-progress
```

```bash
guidellm benchmark run   --target "http://localhost:8000"   --request-type /v1/audio/transcriptions   --profile kind=synchronous   --max-requests 10   --data '{"kind": "huggingface", "source": "openslr/librispeech_asr", "load_kwargs": {"name": "clean", "split": "test"}}'   --data-column-mapper '{"column_mappings": {"audio_column": "audio"}}'   --disable-progress
```

To explicitly use MP3 like it used to, do:
```bash
guidellm benchmark run \
  --request-type /v1/audio/transcriptions \
  --profile kind=synchronous \
  --max-requests 10 \
  --data '{"kind": "huggingface", "source": "openslr/librispeech_asr", "load_kwargs": {"name": "clean", "split": "test"}}' \
  --data-column-mapper '{"column_mappings": {"audio_column": "audio"}}' \
  --data-preprocessors '{"kind": "encode_media", "audio_kwargs": {"audio_format": "mp3"}}' \
  --disable-progress
```

## Related Issues

- Resolves #623 

---

## Use of AI

---

# git log

commit 6cc6e9fa72178752632956b5b193a4107c3aee98
Date:   Thu Jun 11 15:00:08 2026 -0400

    Infer audio encoding format from source instead of defaulting to MP3
    
commit 09d77d0bd91ad68ad5424b6c3a46554cf0f20c8b
Date:   Mon Jun 15 19:13:30 2026 -0400

    Address review feedback
    
---------

### DIFF
--- a/src/guidellm/backends/vllm_python/vllm.py
+++ b/src/guidellm/backends/vllm_python/vllm.py
@@ -302,11 +302,8 @@ class VLLMPythonBackend(Backend):
                     try:
                         # Decode raw audio bytes into an array since vLLM audio models
                         # expect either raw numpy arrays or specific tensor formats
-                        audio_samples, _ = audio._decode_audio(audio_bytes)  # noqa: SLF001
-                        # torchcodec decodes audio on CPU, so .data is always
-                        # a CPU torch.Tensor. .cpu() is a no-op on CPU tensors.
-                        audio_array = audio_samples.data.cpu().numpy()
-                        multi_modal_data["audio"] = audio_array
+                        audio_data = audio.decode_audio(audio_bytes)
+                        multi_modal_data["audio"] = audio_data
                     except (ValueError, TypeError, OSError, RuntimeError) as exc:
                         raise ValueError(
                             f"Failed to decode audio from audio_column for vLLM: {exc}"
--- a/src/guidellm/backends/vllm_python/vllm.py
+++ b/src/guidellm/backends/vllm_python/vllm.py
@@ -302,7 +302,7 @@ class VLLMPythonBackend(Backend):
                     try:
                         # Decode raw audio bytes into an array since vLLM audio models
                         # expect either raw numpy arrays or specific tensor formats
-                        audio_samples = audio._decode_audio(audio_bytes)  # noqa: SLF001
+                        audio_samples, _ = audio._decode_audio(audio_bytes)  # noqa: SLF001
                         # torchcodec decodes audio on CPU, so .data is always
                         # a CPU torch.Tensor. .cpu() is a no-op on CPU tensors.
                         audio_array = audio_samples.data.cpu().numpy()
--- a/src/guidellm/utils/audio.py
+++ b/src/guidellm/utils/audio.py
@@ -32,8 +32,6 @@ _CODEC_TO_FORMAT: dict[str, str] = {
     "opus": "ogg",
 }
 
-_wav_fallback_state = {"warned": False}
-
 
 def _codec_to_format(codec: str) -> str | None:
     """
@@ -103,13 +101,7 @@ def encode_audio(
             audio_format = _codec_to_format(source_codec)
         if audio_format is None:
             audio_format = "wav"
-            if not _wav_fallback_state["warned"]:
-                _wav_fallback_state["warned"] = True
-                logger.warning(
-                    "Could not detect source audio codec; "
-                    "falling back to WAV encoding. "
-                    "Set audio_format explicitly to suppress this warning."
-                )
+            logger.debug("Falling back to WAV audio formatting")
 
     bitrate_val = (
         int(bitrate.rstrip("k")) * 1000 if bitrate.endswith("k") else int(bitrate)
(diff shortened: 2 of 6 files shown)


src/guidellm/utils/audio.py
```diff
@@ -89,8 +154,12 @@ def _decode_audio(  # noqa: C901, PLR0912
     | dict[str, Any],
     sample_rate: int | None = None,
     max_duration: float | None = None,
-) -> libs.AudioSamples:
-    """Decode audio from various input types into AudioSamples."""
+) -> tuple[libs.AudioSamples, str | None]:
+    """
+    Decode audio from various input types into AudioSamples.
+
+    :return: Tuple of (decoded samples, detected codec string or None).
```
dbutenhof, line 154: Not a new problem -- but while adding a `:return:` I probably would have added the parameters to the docstring, too.

src/guidellm/utils/audio.py
```diff
@@ -65,12 +124,18 @@ def encode_audio(
         mono=mono,
     )
 
+    # Use source file name when available, otherwise build from resolved format
+    if isinstance(audio, str | Path):
+        resolved_file_name = get_file_name(audio)
+    elif file_name is not None:
+        resolved_file_name = file_name
+    else:
+        resolved_file_name = f"audio.{format_val}"
```
dbutenhof: In the OpenAI `AudioRequestHandler`, a null `file_name` field defaults to the string `"audio_input"`, but now it'll get your constructed name. This seems to be passed into the HTTP request, and I assume it's mostly informational on the server side, which would make it mostly redundant with the `mimetype` in any messaging. But was there a deeper point here?

dbutenhof (dismissed):
A minor comment and a minor question ... but looks OK.

src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -302,7 +302,7 @@ def _build_multi_modal_data_from_columns(  # noqa: C901, PLR0912
                     try:
                         # Decode raw audio bytes into an array since vLLM audio models
                         # expect either raw numpy arrays or specific tensor formats
-                        audio_samples = audio._decode_audio(audio_bytes)  # noqa: SLF001
```
sjmonson, line 305: Back when you did the vLLM backend I meant to say something about this use of a private method. Now would probably be a good time to fix it; Maybe just add a new method to `guidellm.utils.audio` that simplifies the interface like:

```python
def decode_audio(audio: bytes):
    audio_samples, _ = _decode_audio(audio)
    return audio_samples.data.cpu().numpy()
```

src/guidellm/utils/audio.py
sjmonson: Yeah I would drop this change as its unnecessary. The filename doesn't need to match the mime type.

src/guidellm/utils/audio.py
```diff
@@ -49,8 +75,41 @@ def encode_audio(
...
+            audio_format = _codec_to_format(source_codec)
+        if audio_format is None:
+            audio_format = "wav"
+            if not _wav_fallback_state["warned"]:
+                _wav_fallback_state["warned"] = True
+                logger.warning(
+                    "Could not detect source audio codec; "
+                    "falling back to WAV encoding. "
+                    "Set audio_format explicitly to suppress this warning."
+                )
```
sjmonson:
```suggestion
```

No. If you really want to do this use [`warnings`](https://docs.python.org/3/library/warnings.html#repeated-warning-suppression-criteria) which are designed for this "only emit once" pattern.

sjmonson (changes requested):

src/guidellm/utils/audio.py
jaredoconnell: If you want I can do that. I think the return type of less self-explanatory than the parameters, which I would say are self-documenting. So I'm not sure if there's value added by doing this.

src/guidellm/utils/audio.py
jaredoconnell: I switched it to a simple debug log. How about that?

src/guidellm/utils/audio.py
jaredoconnell: Yeah, that works. Reverted. This is leftover from trying a different design.

src/guidellm/backends/vllm_python/vllm.py
jaredoconnell: Ah I shouldn't have let that slip through either. Fixed.

dbutenhof (commented):
Looks good, aside from 

> 2 files would be reformatted, 254 files already formatted

sjmonson (approved):

dbutenhof (approved):
Hopefully the CI will be happy -- if it is, let's ship it.

PR #795, opened 2026-06-16
fix(openai): surface streaming SSE error payloads as request failures (#743)

### LINKED ISSUE #743: Streaming error responses are not recognized as errors
### Bug Description

When the backend server returns an error via streaming SSE messages (HTTP 200 with `data: {"error": {...}}` followed by `data: [DONE]`), guidellm treats the generation as successful rather than failed. This happens because the initial HTTP connection succeeds (status 200), and the error is only conveyed through the stream body — which guidellm does not inspect for error payloads.

Services like vLLM use this pattern (e.g., `create_streaming_error_response`) to report errors mid-stream after the connection is already established. A typical error stream looks like:

```
data: {"error": {"message": "TimeoutError", "type": "GatewayTimeout", "code": 504}}

data: [DONE]
```

Because guidellm ignores the error payload, the reported metrics are misleading:
- **TTFT** and **ITL** are reported as `0` (no actual content was streamed)
- **Token counts** show the expected/requested numbers rather than actual output
- **Words** and **characters** are reported as `0`

### Expected Behavior

Streaming responses that contain an `"error"` key should be recognized as failures. The generation should be marked as errored/aborted, and metrics (TTFT, ITL, token counts) should reflect that no valid output was produced.

A more robust approach would follow vLLM's `ErrorResponse` / `ErrorInfo` schema to parse and surface the error details (message, type, code, param) to the caller.

**Error vs. incomplete**: One point worth discussion is whether a streaming error response should be classified as an error or as an incomplete request. From the HTTP perspective, the connection succeeded (status 200), but the stream body conveys an error. Treating it as an explicit error may be most natural when the error is the first (and only) streaming message. However, classifying it as incomplete has an advantage in cases where the server successfully generated part of the response before the error occurred — the already-streamed tokens could still be recorded rather than discarded entirely.

### Steps to Reproduce

1. Configure guidellm to use an OpenAI-compatible backend that returns streaming errors (e.g., vLLM with a gateway timeout)
2. Submit a generation request that causes a server-side timeout or other error after the stream begins
3. Observe that guidellm reports the request as successful with zero words/characters but non-zero token counts

### Operating System

Ubuntu 24.04

### Python Version

Python 3.12.3

### GuideLLM Version

0.6.0

### Installation Method

pip install guidellm[recommended]

### Installation Details

_No response_

```shell

```

### Additional Context

**vLLM reference:** vLLM defines its streaming error response format as:

```python
class ErrorInfo(OpenAIBaseModel):
    message: str
    type: str
    param: str | None = None
    code: int

class ErrorResponse(OpenAIBaseModel):
    error: ErrorInfo
```

And serializes it via `create_streaming_error_response()` which produces a JSON line with the structure `{"error": {"message": "...", "type": "...", "code": ...}}`.

**Temporary workaround**:

```diff
diff --git a/src/guidellm/backends/openai/request_handlers.py b/src/guidellm/backends/openai/request_handlers.py
index b6e9780..68e0f0b 100644
--- a/src/guidellm/backends/openai/request_handlers.py
+++ b/src/guidellm/backends/openai/request_handlers.py
@@ -10,6 +10,7 @@ extract usage metrics, and convert results into standardized GenerationResponse.
 
 from __future__ import annotations
 
+import asyncio
 import base64
 from typing import Any, Protocol, cast
 
@@ -563,6 +564,9 @@ class ChatCompletionsRequestHandler(TextCompletionsRequestHandler):
         if not (data := self.extract_line_data(line)):
             return None if data is None else 0
 
+        if "error" in data:
+            raise asyncio.CancelledError
+
         if "id" in data and self.streaming_response_id is None:
             self.streaming_response_id = data["id"]
```

Anai-Guo:
## Summary

Fixes #743. Streaming responses that convey a server-side error through the
SSE body — `data: {"error": {...}}` followed by `data: [DONE]`, returned with
HTTP 200 — were treated as successful empty generations.

Several OpenAI-compatible servers report mid-stream failures this way (e.g.
vLLM's `create_streaming_error_response`). Because the HTTP connection
succeeds and guidellm only inspected `choices`/`delta` payloads, the error
field was ignored. The request was recorded as **successful** with:

- TTFT and ITL reported as `0` (no content streamed)
- token counts showing the requested numbers rather than the actual output
- words/characters reported as `0`

## Change

- Add a small module-level helper `_check_streaming_error(data)` that raises
  `ValueError` when a parsed SSE payload carries a non-empty top-level
  `error` field, surfacing the server-provided message
  (`message` → `type` → `code` → repr).
- Call it from both `extract_line_data` implementations: the
  `TextCompletionsRequestHandler` base (shared by chat / audio / pooling) and
  the `ResponsesRequestHandler` override.

Raising `ValueError` matches the existing request-failure convention in this
backend (see `_check_tool_call_expectations`'s `error_stop` path), so the
streaming loop in `http.py` propagates it and the worker marks the request as
errored. Empty or `null` `error` fields (which some servers emit on normal
chunks) are ignored.

## Notes

- The issue raised a design question of *error vs. incomplete* classification.
  This PR implements the straightforward "treat as error" behavior, which is
  the natural default when the error is the first/only streaming message.
  Preserving already-streamed tokens for the partial-then-error case could be
  layered on top later without changing this detection point.

## Test

Adds `test_extract_line_data_streaming_error` asserting `extract_line_data`
raises `ValueError` for both dict-form and string-form error payloads. Since
chat/audio/pooling inherit the base `extract_line_data`, the shared path is
covered once.

---

# git log

commit 598d578cb5d1fd630e611ad809c173703847c0ff
Date:   Tue Jun 16 10:03:06 2026 +0000

    fix(openai): surface streaming SSE error payloads as request failures (#743)
    
    Some OpenAI-compatible servers (e.g. vLLM's create_streaming_error_response)
    return HTTP 200 and report failures through the stream body as
    `data: {"error": {...}}` followed by `data: [DONE]`. guidellm only inspected
    choices/delta payloads, so the error was silently ignored: the request was
    recorded as a successful but empty generation with TTFT/ITL reported as zero
    and the requested (not actual) token counts.
    
    Add a shared `_check_streaming_error` helper invoked from both
    `extract_line_data` implementations (the TextCompletions base shared by
    chat/audio/pooling, and the Responses override). When a parsed SSE payload
    carries a non-empty top-level `error` field it raises `ValueError` with the
    server-provided message, following the same convention as other request-level
    failures so the worker marks the request as errored. Empty/null `error`
    fields are ignored.
    
commit dd7fa5199e34e08dcff38abbf1e0dbd5b1d51980
Date:   Tue Jun 16 22:07:09 2026 +0000

    style: apply ruff format to request_handlers and test
    
commit dbe8daf40e36f759bf70b093e2b8751483ba3ccc
Date:   Thu Jun 18 06:10:28 2026 -0700

    fix: restore em-dashes mangled by Windows cp1252 encoding
    
    dbutenhof noted that the earlier commit converted several em-dashes (U+2014)
    in comments/docstrings of request_handlers.py into mojibake. Restore the
    original em-dash characters.
    
---------

### DIFF
--- a/src/guidellm/backends/openai/request_handlers.py
+++ b/src/guidellm/backends/openai/request_handlers.py
@@ -212,7 +212,7 @@ def _wrap_reasoning(reasoning_text: str | None, mode: bool | str) -> str | None:
     """Apply the configured reasoning format to reasoning text.
 
     :param reasoning_text: Raw reasoning text from the model response.
-    :param mode: Wrapping mode â€” False disables, True uses the default
+    :param mode: Wrapping mode — False disables, True uses the default
         ``<think>...</think>`` template, a string is used as a format
         template containing ``{reasoning}``.
     :return: Formatted reasoning string, or None when disabled or no text.
@@ -511,7 +511,7 @@ class TextCompletionsRequestHandler(OpenAIRequestHandler):
         :return: Tuple of input_metrics and output_metrics as UsageMetrics objects
         """
         if text is None:
-            # text not applicable (e.g. tool-call-only) â€” exclude from aggregation
+            # text not applicable (e.g. tool-call-only) — exclude from aggregation
             text_words = None
             text_chars = None
         else:
@@ -1097,7 +1097,7 @@ class AudioRequestHandler(ChatCompletionsRequestHandler):
         :return: Tuple of input_metrics and output_metrics as UsageMetrics objects
         """
         if text is None:
-            # text not applicable (e.g. tool-call-only) â€” exclude from aggregation
+            # text not applicable (e.g. tool-call-only) — exclude from aggregation
             text_words = None
             text_chars = None
         else:
@@ -1705,7 +1705,7 @@ class ResponsesRequestHandler(OpenAIRequestHandler):
         # for multimodal breakdowns, mirroring chat completions'
         # "prompt_tokens_details"/"completion_tokens_details".
         if text is None:
-            # text not applicable â€” exclude from aggregation
+            # text not applicable — exclude from aggregation
             text_words = None
             text_chars = None
         else:
--- a/src/guidellm/backends/openai/request_handlers.py
+++ b/src/guidellm/backends/openai/request_handlers.py
@@ -152,6 +152,38 @@ class OpenAIRequestHandlerFactory(RegistryMixin[type[OpenAIRequestHandler]]):
         return handler_cls()
 
 
+def _check_streaming_error(data: Any) -> None:
+    """Raise when a streaming SSE payload conveys a server-side error.
+
+    Some OpenAI-compatible servers (e.g. vLLM's
+    ``create_streaming_error_response``) return HTTP 200 and report failures
+    through the stream body as ``data: {"error": {...}}`` followed by
+    ``data: [DONE]``. Without this check the error payload is ignored, the
+    request is recorded as a successful but empty generation, and timing
+    metrics (TTFT, ITL) are reported as zero. Raising ``ValueError`` follows
+    the same convention as other request-level failures so the worker marks
+    the request as errored.
+
+    :param data: Parsed JSON payload from a single streaming line.
+    :raises ValueError: If the payload contains a non-empty ``error`` field.
+    """
+    if not isinstance(data, dict):
+        return
+    error = data.get("error")
+    if not error:
+        return
+    if isinstance(error, dict):
+        message = (
+            error.get("message")
+            or error.get("type")
+            or error.get("code")
+            or str(error)
+        )
+    else:
+        message = str(error)
+    raise ValueError(f"Streaming response returned an error: {message}")
+
+
 def _apply_tool_call_metrics(
     output_metrics: UsageMetrics,
     tool_call_count: int,
@@ -454,7 +486,9 @@ class TextCompletionsRequestHandler(OpenAIRequestHandler):
 
         line = line[len("data:") :].strip()
 
-        return json.loads(line)
+        data = json.loads(line)
+        _check_streaming_error(data)
+        return data
 
     def extract_choices_and_usage(
         self, response: dict
@@ -1432,7 +1466,9 @@ class ResponsesRequestHandler(OpenAIRequestHandler):
         if line == "data: [DONE]":
             return None
 
-        return json.loads(line[len("data:") :].strip())
+        data = json.loads(line[len("data:") :].strip())
+        _check_streaming_error(data)
+        return data
 
     @staticmethod
     def _responses_item_to_tool_call(item: dict[str, Any]) -> ToolCall:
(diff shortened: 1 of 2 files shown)


sjmonson:
Please rebase. I pushed a fix for the 429 error.

Anai-Guo:
Rebased onto latest main (picked up your tox HF-cache fix for the 429). Dropped the earlier empty re-trigger commit; both real commits retain DCO sign-off. Thanks @sjmonson!

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -215,7 +212,7 @@ def _wrap_reasoning(reasoning_text: str | None, mode: bool | str) -> str | None:
     """Apply the configured reasoning format to reasoning text.
 
     :param reasoning_text: Raw reasoning text from the model response.
-    :param mode: Wrapping mode — False disables, True uses the default
+    :param mode: Wrapping mode â€” False disables, True uses the default
```
dbutenhof: Apparently a fairly common Windows encoding error: this commit has converted what used to be a long dash (em-dash) into this `â€”`.

This is a small issue and probably won't confuse a reader, but it should be fixed. (If it's easier, you can use a single or double hyphen [`-` or `--`] rather than an em-dash.)

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -514,7 +511,7 @@ def extract_metrics(
         :return: Tuple of input_metrics and output_metrics as UsageMetrics objects
         """
         if text is None:
-            # text not applicable (e.g. tool-call-only) — exclude from aggregation
+            # text not applicable (e.g. tool-call-only) â€” exclude from aggregation
```
dbutenhof: And here ...

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -1100,7 +1097,7 @@ def extract_metrics(
...
```
dbutenhof: And here ...

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -1708,7 +1705,7 @@ def extract_metrics(
         # for multimodal breakdowns, mirroring chat completions'
         # "prompt_tokens_details"/"completion_tokens_details".
         if text is None:
-            # text not applicable — exclude from aggregation
+            # text not applicable â€” exclude from aggregation
```
dbutenhof: And here ...

dbutenhof (changes requested):
I know this is fairly trivial, but your character encoding broke some existing comments and docstrings, which should be fixed.

Anai-Guo:
Good catch @dbutenhof — fixed. Restored all four em-dashes (U+2014) that my Windows editor had mangled into `â€”`; the file now contains only the intended em-dash characters and no other non-ASCII. DCO sign-off retained.

dbutenhof (approved):
Without trying to force streaming errors, I did at least verify that this logic doesn't break successful operation with or without `stream=true`.

jaredoconnell (approved):

PR #796, opened 2026-06-16
fix(data): fail fast when no mappable columns are found (#787)

### LINKED ISSUE #787: Benchmark hangs forever (scheduler/worker deadlock) when dataset has no mappable columns — should fail fast instead
## Describe the bug

When the input dataset contains only columns that `GenerativeColumnMapper` has no candidate for (e.g. a single `messages` column), `guidellm benchmark run` prints `Setup complete, starting benchmarks...` and then hangs forever. No requests are ever sent, no error is raised, and the process never exits — in containerized/CI environments it only dies when an external timeout (e.g. Kubernetes `activeDeadlineSeconds`) kills it.

Expected behavior: setup should fail fast with a clear error like "no mappable text/image/video/audio columns found in dataset (columns: ['messages'])".

## Environment

- guidellm: 0.6.0
- Python: 3.10 (Linux)
- Backend: any OpenAI-compatible server (reproduced against `guidellm mock-server` and against vLLM)

## To Reproduce

```bash
# data.jsonl — a column name that is not in GenerativeColumnMapper's candidates
printf '%s\n' '{"messages":[{"role":"user","content":"hello"}]}' > data.jsonl

guidellm mock-server --host 127.0.0.1 --port 18999 &

timeout 120 guidellm benchmark run \
  --target http://127.0.0.1:18999 \
  --model mock \
  --request-format /v1/chat/completions \
  --backend-kwargs '{"validate_backend": false}' \
  --profile concurrent --rate 2 --max-seconds 8 \
  --data data.jsonl \
  --output-dir ./out --outputs json
# → prints "Setup complete, starting benchmarks..." then hangs until killed (exit 124)
```

The same dataset with the column renamed to `prompt` (a `text_column` candidate) completes normally, so the trigger is purely the unmappable column name, not the request format.

## What seems to happen

`GenerativeColumnMapper.datasets_mappings()` (`guidellm/data/preprocessors/mappers.py`) finds no match for any column type, returning empty mappings. Downstream, requests carry no content columns, and the scheduler/worker processes end up idle waiting on each other (confirmed with py-spy: main process and all worker processes parked in idle waits — a lost-wakeup deadlock). `--max-seconds` never fires because the benchmark clock seemingly only starts once requests flow.

## Suggested fix

Validate the resolved column mappings after `datasets_mappings()`: if no text/image/video/audio column was mapped for any dataset, raise a `ValueError` listing the dataset's actual column names and the supported candidates. This turns a silent multi-process hang into an immediate, actionable error.

Happy to provide more detail (py-spy dumps) if useful.

Anai-Guo:
## Summary

Fixes #787.

When a dataset contains only columns that `GenerativeColumnMapper` has no candidate for (e.g. a single `messages` column), `setup_data` previously just logged a warning and returned an empty mapping. Every row then produced an empty result, the scheduler started but never enqueued a request, and `guidellm benchmark run` hung forever — only killed by an external timeout.

This changes `setup_data` to **fail fast**: when no columns can be mapped it raises a clear `ValueError` listing the columns that were found, so the user knows to rename a column to a recognized name (e.g. `prompt`) or pass an explicit column mapping.

## Behavior

```
ValueError: GenerativeColumnMapper found no mappable columns in the dataset(s)
(columns: ['messages']). Requested mappings: default mappings. Without at least
one mapped column every row produces an empty result and the benchmark would
hang with no requests sent. Rename a column to a recognized name (e.g. 'prompt')
or pass an explicit column mapping.
```

## Test plan

- [x] Dataset with only an unmappable column (`messages`) now raises `ValueError` at setup instead of hanging.
- [x] Dataset with a mappable column (`prompt`) still maps normally (`text_column` resolved).
- [x] `ruff check` passes; unused `logger` import removed.

---

# git log

commit abd762617da6f1d77a52cddb455edf7340a6e613
Date:   Tue Jun 16 10:08:45 2026 +0000

    fix(data): fail fast when no mappable columns are found
    
    GenerativeColumnMapper.setup_data only logged a warning when a dataset
    had no mappable columns, then returned an empty mapping. Every row then
    produced an empty result, the scheduler started but never enqueued any
    request, and the benchmark hung forever (only killed by an external
    timeout).
    
    Raise a clear ValueError at setup time instead, listing the dataset
    columns that were found so the user knows to rename a column to a
    recognized name (e.g. 'prompt') or pass an explicit column mapping.
    
    Fixes #787
    
commit b79d0bc69bff3314be438f49bfbbbedcc905d963
Date:   Tue Jun 16 16:08:49 2026 +0000

    fix(data): normalize line endings to LF
    
commit 90d11edad815ff799d87bcff7e46fc5b7caa4f87
Date:   Tue Jun 16 19:07:07 2026 +0000

    fix(data): add trailing newline to satisfy ruff format and EOF fixer
    
commit 27f298f3d29c2285f6a53dedc197d5ea29ee5df1
Date:   Wed Jun 17 01:03:12 2026 +0000

    fix(data): keep concise no-columns error message per review
    
    Drop the verbose column listing and hang explanation; raise a short
    ValueError as requested.
    
commit afa5b268de176ad9f0551ea8b8b380102352873a
Date:   Wed Jun 17 07:09:30 2026 +0000

    style: add trailing newline to mappers.py (end-of-file-fixer + ruff format)
    
commit 41b1efedfcc33723989a5c5e7e524b2fa7ecd615
Date:   Wed Jun 17 10:03:15 2026 +0000

    fix(data): keep original warning message text when raising ValueError
    
    Addresses review: preserve the original message wording per maintainer
    request, only changing the warning into a fatal ValueError.
    
    Signed-off-by: Anai-Guo <antai12232931@anaiguo.com>
    
---------


src/guidellm/data/preprocessors/mappers.py
```diff
@@ -250,14 +249,22 @@ def setup_data(
...
+                "result and the benchmark would hang with no requests sent. "
+                "Rename a column to a recognized name (e.g. 'prompt') or pass "
+                "an explicit column mapping."
             )
 
 
 @PreprocessorRegistry.register("pooling_column_mapper")
 class PoolingColumnMapper(GenerativeColumnMapper):
-    defaults: ClassVar[dict[str, list[str]]] = {"pooling_column": ["prompt"]}
+    defaults: ClassVar[dict[str, list[str]]] = {"pooling_column": ["prompt"]}
```
dbutenhof: The last line of a file needs to be terminated.

dbutenhof (changes requested):
Thanks for catching and fixing the CRLF line termination!

Anai-Guo:
Thanks @dbutenhof! Good catch — the missing final newline is fixed in the latest commit (the file now ends with a trailing `\n`, satisfying the ruff `end-of-file-fixer`). Ready for another look whenever you have a moment.

src/guidellm/data/preprocessors/mappers.py
```diff
@@ -250,11 +249,19 @@ def setup_data(
...
+            for dataset in datasets:
+                columns = getattr(dataset, "column_names", None)
+                if columns:
+                    available_columns.extend(columns)
+            raise ValueError(
+                "GenerativeColumnMapper found no mappable columns in the "
+                f"dataset(s) (columns: {available_columns}). "
+                f"Requested mappings: {self.input_mappings or 'default mappings'}. "
+                "Without at least one mapped column every row produces an empty "
+                "result and the benchmark would hang with no requests sent. "
```
dbutenhof: I'm uncertain about this sentence. It's a long message already, and while the "Without at least one mapped column..." is true and potentially useful, I also suspect that for most readers it won't add much given the rest of the message. I'm going to approve anyway (we can always shorten it later), but I'm pointing it out to make sure everyone gives it some thought.

dbutenhof (dismissed):

src/guidellm/data/preprocessors/mappers.py
jaredoconnell: Saying that this would cause a hang does correctly describe the current behavior, but and I can imagine a future where we change that. And we can update the message at that time. We can change this in the future.

sjmonson: I would prefer keeping the original message and just converting it to a `ValueError`. We had this debate the last time this part of the code was changed and the conclusions were:

- Listing out all columns of all datasets is more confusing then helpful
- Once an error message gets this long people don't read it. Just keep it shorter and we can document the full meaning somewhere else.

sjmonson (changes requested):

src/guidellm/data/preprocessors/mappers.py
```diff
@@ -250,11 +249,19 @@ def setup_data(
...
-                "GenerativeColumnMapper found no matching columns. "
-                "Requested mappings: {}. "
-                "Every row will produce an empty result.",
-                self.input_mappings or "default mappings",
+            available_columns: list[str] = []
+            for dataset in datasets:
+                columns = getattr(dataset, "column_names", None)
+                if columns:
+                    available_columns.extend(columns)
+            raise ValueError(
```
jaredoconnell: @sjmonson are you okay with the available column list? I added that in my prior PR, and that would probably be my preference, but I recall you wanting that removed.

jaredoconnell (dismissed):
Yeah, it makes sense to switch this check from a warning to a fatal exception.

Anai-Guo:
Thanks @sjmonson @jaredoconnell @dbutenhof — addressed in the latest commit. I reverted to the original concise wording and just promote it to a `ValueError`:

```
GenerativeColumnMapper found no matching columns. Requested mappings: {...}. Every row would produce an empty result.
```

No column listing, no hang explanation — kept short as requested. Ready for another look.

jaredoconnell (dismissed):

Anai-Guo:
Thanks @sjmonson @dbutenhof @jaredoconnell. Reverted to the original message text and kept it as-is, just converting the `logger.warning` into a fatal `ValueError`. Happy to drop the available-column list too if that's the consensus.

dbutenhof (approved):

sjmonson (approved):

dbutenhof:
This failed due to Huggingface rate limiting, which has been a recurring CI issue. I re-ran the job -- I'm not sure whether mergify will continue on its own once (if) that passes...

dbutenhof:
@mergifyio queue

sjmonson:
@mergifyio queue

PR #800, opened 2026-06-16
feat: add LiteLLM backend for multi-provider benchmarking

RheagalFire:
## Summary

Adds a new `litellm` backend that routes generation requests through the LiteLLM SDK, enabling benchmarking across 100+ providers (Anthropic, Gemini, Bedrock, Groq, Cohere, Mistral, etc.) via a unified interface. Timing instrumentation matches the existing OpenAI HTTP backend so benchmark results are directly comparable.

## Details

- [x] New `LiteLLMBackend` and `LiteLLMBackendArgs` following the existing `Backend` / `BackendArgs` registration pattern
- [x] Uses `litellm.acompletion(stream=True)` with `drop_params=True` for cross-provider compatibility
- [x] Reuses `ChatCompletionsRequestHandler.format()` to build messages from `GenerationRequest.columns`
- [x] Lazy-loaded via `guidellm.extras.litellm` so the optional dep doesn't break imports when not installed
- [x] `litellm>=1.80.0,<1.87.0` added as optional dependency under `[project.optional-dependencies].litellm`
- [x] 21 unit tests covering args, registration, lifecycle, streaming dispatch, timing, and token usage
- [x] All ruff checks pass
- [x] All 392 existing backend tests still pass

## Test Plan

- Run `pytest tests/unit/backends/litellm/ -v` to verify unit tests
- Run `pytest tests/unit/backends/ -v` to verify no regressions in existing backends
- Live E2E verified with `anthropic/claude-sonnet-4-6` via Azure Foundry:
  ```
  args = LiteLLMBackendArgs(
      model="anthropic/claude-sonnet-4-6",
      api_key="...",
      api_base="...",
      max_tokens=50,
  )
  ```
  Confirmed: streaming works, TTFT signal fires, token counts captured, timing fields populated.

## Related Issues

- N/A (new feature)

---

## Use of AI

---

# git log

commit 0dba09be0a4b1fc2c0faa6df5ec442ecf37d1dc3
Author: RheagalFire <arishalam121@gmail.com>
Date:   Tue Jun 16 23:51:47 2026 +0530

    feat: add LiteLLM backend for multi-provider benchmarking
    
    Generated-by: Claude claude-opus-4-6
    Signed-off-by: RheagalFire <arishalam121@gmail.com>

commit 1bad39c7d2c8b95ce6ea51cccebe4662fdbd6f46
Author: Aarish Alam <arishalam121@gmail.com>
Date:   Wed Jul 1 20:00:26 2026 +0530

    fix: raise litellm minimum to 1.83, remove upper cap, regen lockfile
    
    - Bump minimum litellm version to >=1.83.0 (post supply-chain fix)
    - Remove <1.87.0 upper bound to allow latest releases
    - Regenerate uv.lock via tox run -e lock
    - Fix ruff formatting issues
    
    Signed-off-by: Aarish Alam <arishalam121@gmail.com>

commit f848ceb2a4abb68c61e1a5a36921514f4acf80eb
Author: Aarish Alam <arishalam121@gmail.com>
Date:   Thu Aug 6 00:28:49 2026 +0530

    refactor(litellm): use Backend-provided _args, dataclass _ChunkResult, __all__ in stub
    
    Signed-off-by: Aarish Alam <arishalam121@gmail.com>

---------

Generated-by: Claude claude-opus-4-6
Signed-off-by: RheagalFire <arishalam121@gmail.com>
Signed-off-by: Aarish Alam <arishalam121@gmail.com>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,7 +91,7 @@ recommended = ["guidellm[perf,tokenizers]"]
 # Feature Extras
 plot = ["matplotlib>=3.10.9"]
 perf = ["orjson", "msgpack", "msgspec", "uvloop"]
-litellm = ["litellm>=1.80.0,<1.87.0"]
+litellm = ["litellm>=1.83.0"]
 tokenizers = ["tiktoken", "blobfile", "mistral-common"]
 audio = [
     # Version with stable aarch64 CPU-only wheels
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,11 +86,12 @@ dependencies = [
 
 [project.optional-dependencies]
 # Meta Extras
-all = ["guidellm[perf,tokenizers,audio,vision,plot]"]
+all = ["guidellm[perf,tokenizers,audio,vision,plot,litellm]"]
 recommended = ["guidellm[perf,tokenizers]"]
 # Feature Extras
 plot = ["matplotlib>=3.10.9"]
 perf = ["orjson", "msgpack", "msgspec", "uvloop"]
+litellm = ["litellm>=1.80.0,<1.87.0"]
 tokenizers = ["tiktoken", "blobfile", "mistral-common"]
 audio = [
     # Version with stable aarch64 CPU-only wheels
--- a/src/guidellm/backends/litellm/litellm.py
+++ b/src/guidellm/backends/litellm/litellm.py
@@ -54,9 +54,7 @@ class LiteLLMBackendArgs(BackendArgs):
     )
     api_key: SecretStr | None = Field(
         default=None,
-        description=(
-            "API key for the provider (overrides provider env var)."
-        ),
+        description=("API key for the provider (overrides provider env var)."),
     )
     api_base: str | None = Field(
         default=None,
@@ -69,8 +67,7 @@ class LiteLLMBackendArgs(BackendArgs):
     extras: dict[str, Any] | None = Field(
         default=None,
         description=(
-            "Additional keyword arguments forwarded to "
-            "litellm.acompletion()."
+            "Additional keyword arguments forwarded to litellm.acompletion()."
         ),
     )
 
@@ -117,9 +114,7 @@ class LiteLLMBackend(Backend):
     async def validate(self):
         """Validate that model string is non-empty."""
         if not self._args.model:
-            raise RuntimeError(
-                "LiteLLM backend requires a non-empty model string."
-            )
+            raise RuntimeError("LiteLLM backend requires a non-empty model string.")
 
     async def available_models(self) -> list[str]:
         """Return the configured model as the only available model."""
@@ -134,8 +129,7 @@ class LiteLLMBackend(Backend):
         request: GenerationRequest,
         request_info: RequestInfo,
         history: (
-            list[tuple[GenerationRequest, GenerationResponse | None]]
-            | None
+            list[tuple[GenerationRequest, GenerationResponse | None]] | None
         ) = None,
     ) -> AsyncIterator[tuple[GenerationResponse | None, RequestInfo]]:
         """
@@ -148,7 +142,9 @@ class LiteLLMBackend(Backend):
         """
         model = self._args.model
         messages, request_args_json = self._format_request(
-            request, history, model,
+            request,
+            history,
+            model,
         )
         call_kwargs = self._build_call_kwargs()
 
@@ -169,7 +165,9 @@ class LiteLLMBackend(Backend):
 
             async for chunk in response_stream:
                 token_info = self._process_chunk(
-                    chunk, request_info, texts,
+                    chunk,
+                    request_info,
+                    texts,
                 )
                 if token_info.response_id and response_id is None:
                     response_id = token_info.response_id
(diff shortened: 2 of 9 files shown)


RheagalFire:
cc  @sjmonson

pyproject.toml
```diff
@@ -76,10 +76,11 @@ dependencies = [
 
 [project.optional-dependencies]
 # Meta Extras
-all = ["guidellm[perf,tokenizers,audio,vision]"]
+all = ["guidellm[perf,tokenizers,audio,vision,litellm]"]
 recommended = ["guidellm[perf,tokenizers]"]
 # Feature Extras
 perf = ["orjson", "msgpack", "msgspec", "uvloop"]
+litellm = ["litellm>=1.80.0,<1.87.0"]
```
sjmonson: Any reason to cap the version at `1.87.0` when the latest available is `1.89.*`? Also the minimum should be no lower then `1.83` due to that [supply chain attack](https://docs.litellm.ai/blog/security-update-march-2026) from a while back.

sjmonson (changes requested):
Just a quick issue I noticed. Also I am guessing you didn't run with pre-check locally since at minimum you will need to regen the lock file after changing dependencies. You can do that with `tox run -e lock` and plain `tox run` will run all the other CI tasks if you want to have a faster feedback loop.

RheagalFire:
@sjmonson Fixed the version bounds.

src/guidellm/backends/litellm/litellm.py
```diff
@@ -0,0 +1,324 @@
...
+        *,
+        response_id: str | None,
+        input_tokens: int | None,
+        output_tokens: int | None,
+        first_token: bool,
+    ):
+        self.response_id = response_id
+        self.input_tokens = input_tokens
+        self.output_tokens = output_tokens
+        self.first_token = first_token
```
sjmonson: Its not critical but this makes a lot more sense as a dataclass or a pydantic model.

src/guidellm/extras/litellm.pyi
```diff
@@ -0,0 +1,2 @@
+from litellm import acompletion as acompletion
+from litellm import completion as completion
```
sjmonson: So this is my bad but I found out that `__all__` is the technically correct way to export from typestubs. I have #893 to fix this on the existing extras but if you could convert this to the following that would be helpful:


```suggestion
from litellm import acompletion, completion

__all__ = ["acompletion", "completion"]
```

sjmonson (dismissed):
A few minor nits but otherwise LGTM. Needs a rebase due to lockfile changes. The easiest way I have found to do this is to run rebase and when you get to the lockfile conflict:

```sh
git restore --staged --worktree uv.lock
uv lock  # or `tox -e lock`
git add uv.lock
git rebase --continue
```

src/guidellm/backends/litellm/litellm.py
```diff
@@ -0,0 +1,324 @@
...
+            process(response)
+        await backend.process_shutdown()
+    """
+
+    def __init__(self, args: LiteLLMBackendArgs):
+        super().__init__(args)
+        self._args = args
+        self._in_process = False
+
+    async def process_startup(self):
```
dbutenhof, line 103: You should add

```python

    @property
    def info(self) -> dict[str, Any]:
        """
        Get backend configuration details.

        :return: Dictionary containing backend configuration details
        """
        return self._args.model_dump(mode="json")
```

This would support several internal usage patterns for reporting, including CSV reports (which, right now, I suspect will fail if you specify an `api_key`, as happens for `openai_http`.)

This is relegated to the implementations as the `Backend` base class doesn't have `_args` or `info`, but (note to myself and our team, not necessarily to the author) we should probably consolidate this pattern better ...

src/guidellm/backends/litellm/litellm.py
dbutenhof: A `@dataclass` is simple & ideal for a logic-free data-carrier class like this.

src/guidellm/backends/litellm/litellm.py
dbutenhof: FYI -- I'm doing that refactoring. So don't bother with the `info` method: you'll hopefully inherit it from `Backend`.

dbutenhof (approved):
I worried a moment about including the `litellm` layer in `all` since we build our container image that way -- however `litellm` 1.83 is already available in the AIPCC internal index and it's not enormous, so that shouldn't be an issue.

src/guidellm/backends/litellm/litellm.py
```diff
@@ -0,0 +1,324 @@
...
+            model="anthropic/claude-haiku-4-5",
+            api_key="sk-ant-...",
+        )
+        backend = LiteLLMBackend(args)
+        await backend.process_startup()
+        async for response, info in backend.resolve(request, info):
+            process(response)
+        await backend.process_shutdown()
+    """
+
```
dbutenhof, line 96: Add

```python
    _args: LiteLLMBackendArgs
```

(for the static type checker) and remove `self._args = args` from the `__init__` method.

dbutenhof (dismissed):
You need to resolve the `uv.lock` conflict and adapt to a new backend init convention that fixes CSV benchmark reports.

sjmonson (approved):

sjmonson:
@mergifyio queue

src/guidellm/backends/litellm/litellm.py
```diff
@@ -0,0 +1,312 @@
...
+            data=request,
+            history=history,
+            model=model,
+            stream=True,
+            max_tokens=self._args.max_tokens,
+        )
+        messages: list[dict[str, Any]] = (
+            arguments.body.get("messages", []) if arguments.body else []
+        )
+        return messages, arguments.model_dump_json()
```
jaredoconnell, line 228: Have you verified that we only need the messages from the format output? Looking it looks like ChatCompletionsRequestHandler.format() builds a full body (e.g. max_completion_tokens / ignore_eos from request.output_metrics.text_tokens, tools, etc.), but only messages are passed into litellm.acompletion(). The rest appears to end up in arguments.model_dump_json() for GenerationResponse.request_args. Can you confirm whether those body fields are intentionally not forwarded, or if request-level output token limits should also be sent to LiteLLM?

src/guidellm/backends/litellm/litellm.py
```diff
@@ -0,0 +1,312 @@
...
+            request_info.timings.first_request_iteration = iter_time
+        request_info.timings.last_request_iteration = iter_time
+        request_info.timings.request_iterations += 1
+
+        response_id = chunk.id if chunk.id else None
+
+        usage = getattr(chunk, "usage", None)
+        input_tokens = getattr(usage, "prompt_tokens", None) if usage else None
+        output_tokens = getattr(usage, "completion_tokens", None) if usage else None
+
```
jaredoconnell, line 263: We discourage the use of `getattr`. Have you looked into statically typing `chunk` so that you can avoid `getattr`? The type appears to be `ModelResponseStream`.

sjmonson:
@RheagalFire Just fyi the merge failed since this PR is failing CI when rebased. Also see above comments from @jaredoconnell.

dbutenhof (commented):
Minimal documentation on the litellm backend should be added to `docs/guides/backends.md` (a new section under Supported Backends), describing the usage and use cases. If the documentation is extensive, spin off details into a new `.md` file, with a link from "Supported Backends" as there is for the "vLLM Python Backend".

sjmonson:
@RheagalFire Hey sorry this has been stuck in review for so long. As a semi-apology if we don't hear back from you by the end of the week I can take over fixing up for merge.

PR #448, opened 2025-11-06
[Draft]remove computed_field from total_tokens

tukwila:
## Details

- [ ]

To fix: https://github.com/vllm-project/guidellm/issues/446

test cmd:
tox -e test -- --runxfail tests/unit/utils/test_encoding.py::TestMessageEncoding::test_encode_decode_generative

## Test Plan

-

## Related Issues

- Resolves #

---

- [ ] "I certify that all code in this PR is my own, except as noted below."

## Use of AI

- [ ] Includes AI-assisted code completion
- [ ] Includes code generated by an AI application
- [ ] Includes AI-generated tests (NOTE: AI written tests should have a docstring that includes `## WRITTEN BY AI ##`)

PR #845, opened 2026-06-24
Fix: Exclude computed fields during pydantic MP serialization

### LINKED ISSUE #446: serializer.deserialize(serialized) duplicate the last dict data
**Describe the bug**

when i debug xfail ut case: tests/unit/utils/test_encoding.py::TestMessageEncoding::test_encode_decode_generative[dict_serialization_msgpack-obj0]
assertion failure is:

Image

As screenshot show, the last dict data total_tokens=None is duplicated.

And the problem is serializer.deserialize(serialized) in src/guidellm/utils/encoding.py:

Image

**Expected behavior**

**Environment**
1. OS [e.g. Ubuntu 20.04]: conda 25.3.1
2. Python version [e.g. 3.12.2]: guidellm version: 0.4.0.dev0

**To Reproduce**
Exact steps to reproduce the behavior:

**Errors**
If applicable, add a full print-out of any errors or exceptions that are raised or include screenshots to help explain your problem.

**Additional context**
Add any other context about the problem here. Also include any relevant files.

SkiHatDuckie:
## Summary
See the linked issue for more details. In short, fields with the`@computed_field` decorator are included in `model_dump` (and similar) by default, but are then treated as extra fields when passed to `model_validate`, leading to duplicates appearing in the final deserialized object.

## Details
- Set `exclude_computed_fields=True` in `Serializer.to_dict_pydantic` and `.to_sequence_pydantic`
- Remove pytest.mark.xfail from `test_encode_decode_generative`

## Test Plan
`tox -e test-unit -- tests/unit/utils/test_encoding.py`

## Related Issues
- Resolves #446
  - Alternative solution to #448

---

## Use of AI

---

# git log

commit 4c59acb978d7fa3264d95d99a52873ec977e36c8
Author: SkiHatDuckie <SkiHatDuckie@gmail.com>
Date:   Wed Jun 24 15:02:04 2026 -0400

    Exclude computed fields during pydantic model dump
    
    Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

---------

Signed-off-by: SkiHatDuckie <SkiHatDuckie@gmail.com>

### DIFF
--- a/src/guidellm/utils/encoding.py
+++ b/src/guidellm/utils/encoding.py
@@ -431,7 +431,7 @@ class Serializer:
             "*PYD*": True,
             "typ": item.__class__.__name__,
             "mod": item.__class__.__module__,
-            "dat": item.model_dump(mode="python"),
+            "dat": item.model_dump(mode="python", exclude_computed_fields=True),
         }
 
     def from_dict_pydantic(self, item: dict[str, Any]) -> Any:
@@ -572,7 +572,9 @@ class Serializer:
         """
         class_name: str = obj.__class__.__name__
         class_module: str = obj.__class__.__module__
-        json_data = obj.__pydantic_serializer__.to_json(obj)
+        json_data = obj.__pydantic_serializer__.to_json(
+            obj, exclude_computed_fields=True
+        )
 
         return class_name.encode() + b"|" + class_module.encode() + b"|" + json_data
 
--- a/tests/unit/utils/test_encoding.py
+++ b/tests/unit/utils/test_encoding.py
@@ -201,7 +201,6 @@ class TestMessageEncoding:
         else:
             assert decoded == obj
 
-    @pytest.mark.xfail(reason="old and broken", run=False)
     @pytest.mark.smoke
     @pytest.mark.parametrize(
         "obj",


sjmonson (approved):
LGTM

dbutenhof (approved):
I suppose my concern would be if we're ever using this in a context where we *want* computed fields to be serialized (e.g., for a report) -- presumably that'd never be in a context where we're going to re-deserialize and encounter the problem, but this is one of those places where it's hard to get a full call dependency graph; and if Sam thinks this is safe, I'll throw on my +1 on the basis that, at least in isolation, the change looks OK.

sjmonson:
> I suppose my concern would be if we're ever using this in a context where we want computed fields to be serialized (e.g., for a report)

Had this thought at first but the title/description is misleading. This only changes the encoding code which is used to pass objects from one process to another.

dbutenhof:
> > I suppose my concern would be if we're ever using this in a context where we want computed fields to be serialized (e.g., for a report)
> 
> Had this thought at first but the title/description is misleading. This only changes the encoding code which is used to pass objects from one process to another.

Yeah, that was the usage I found -- but there was a callback protocol that broke the call chain and I wasn't 100% certain we couldn't get there from somewhere else... though it seemed improbable. I'm reassured that you're comfortable with that diagnosis.

PR #850, opened 2026-06-24
Add WebSocket round-trip metrics (Time To Last Round Trip and Avg RTT)

### LINKED ISSUE #832: Websocket backend metrics
### Problem Statement

When using the websocket backend for a task which has some sort of correspondence between input packets and output tokens (e.g realtime transcription), we might want more metrics than just the classical LLM metrics such as TTFT, TPOT, E2E etc, but might want to know things like the Round-Trip Time ([RTT](https://en.wikipedia.org/wiki/Round-trip_delay)), which is essentially the user experienced system lag.

Without exact alignment between input packets and output tokens and their timings - we cannot know the exact RTT, but we can estimate it in 3 different ways:

1. TTFT - Time between the first sent packet time and the first received token time (`T_recv[0] - T_sent[0]`). Basically the RTT of the first token, which is already an OK proxy
2. TTLT (Time To Last Token) - Time between the last sent packet time and the last received token time  (`T_recv[-1] - T_sent[-1]`). The RTT of the last token, also an OK proxy
3. Avg RTT (approximate) - The time between the average sent packet time and the average received packet time  (`T_recv.mean() - T_sent.mean()`). This is a much more stable metric, because like TPOT it averages over all tokens instead of measuring just once per request. However, this is an approximate estimation of the true RTT, because it assumes that the sent packets and received tokens match each other uniformly in time, which is never exactly true.

### Proposed Solution

When the websocket backend is used, report TTLT and Avg RTT in addition to the rest of the metrics. They are always simple to calculate with the websocket backend regardless of modality and endpoint, they provide useful information for all websocket use cases, and they will provide a unique advantage for realtime transcription benchmarking by approximating the experienced system lag.

### Alternatives Considered

We could just accept TTFT as the only proxy, but it's very different from actual system lag in actual streaming applications.

We could also try to measure the true RTT, but that requires having a dataset with accurate timestamp labels, and to use some forced alignment to align predicted tokens with input audio packets, and accommodate for prediction errors, and it requires a different solution for each future websocket modality. It's an overly complicated solution that will create a fragile benchmark that can't generalize to other datasets.

### Usage Examples

```markdown
python3 -m vllm.entrypoints.openai.api_server \
  --model mistralai/Voxtral-Mini-4B-Realtime-2602 \
  --tokenizer-mode mistral \
  --config-format mistral \
  --load-format mistral \
  --trust-remote-code \
  --compilation-config '{"cudagraph_mode":"PIECEWISE"}' \
  --tensor-parallel-size 1 \
  --max-model-len 45000 \
  --max-num-batched-tokens 8192 \
  --max-num-seqs 16 \
  --gpu-memory-utilization 0.90 \
  --host 0.0.0.0 --port 8000


guidellm benchmark \
  --target http://localhost:8000/v1 \
  --request-type realtime_transcription \
  --backend openai_realtime_ws \
  --data /workspace/custom-audio-dataset/hf_dataset \
  --profile synchronous \
  --max-requests 10 \
  --output-dir /workspace/repo/runs/2026-04-23T13-41-41 \
  --outputs json,html,csv

# TTLT & Avg RTT always reported when `backend=openai_realtime_ws`
```

### Additional Context

https://github.com/vllm-project/guidellm/pull/713
https://en.wikipedia.org/wiki/Round-trip_delay

I am interested in creating a PR for this 😄

dreamer-89:
## Summary

Adds two round-trip metrics to the realtime WebSocket backend (`openai_websocket`): **Time To Last Round Trip** and an approximate **Avg RTT**. In streaming transcription, input packets and output tokens flow as separate streams, so the existing token metrics (TTFT, ITL, TPOT) do not capture the send-to-receive lag a user actually experiences. These two estimate that lag from the send and receive timestamps and are reported automatically whenever the WebSocket backend is used. Per the issue discussion, Time To First Round Trip is omitted because it overlaps the existing TTFT, and the names avoid `TTFT`/`TTLT` to prevent collision with the token metrics.

## Details

- `schemas/info.py`: add scalar timing fields to `RequestTimings` (`last_request_sent`, `request_sent_sum`/`request_sent_count`, `token_received_sum`/`token_received_count`). Scalars rather than lists, so they stay correct under `RequestInfo.model_copy` (which shallow-copies timings).
- `backends/openai/websocket.py`: record a timestamp after each outbound frame (`session.update`, audio appends, both commits) via a small `_record_request_sent` helper, and accumulate received content-token timestamps in `_record_content_tokens`.
- `schemas/request_stats.py`: add computed properties `time_to_last_round_trip_ms` (last received token minus last sent packet) and `avg_round_trip_time_ms` (mean received minus mean sent). Both return `None` when no sends were recorded, so non-WebSocket backends are unaffected.
- `benchmark/schemas/metrics.py` and `accumulator.py`: add the two `StatusDistributionSummary` fields, compile them in `GenerativeMetrics.compile()`, and accumulate them for live progress.
- `benchmark/outputs/console.py` and `csv.py`: show both in the request-latency table and the CSV export. JSON/YAML already include them via the schema. The HTML report is left for a follow-up.
- `docs/guides/metrics.md`: document both metrics, noting they are WebSocket-only and that Avg RTT is approximate (it assumes sent packets and received tokens line up evenly in time).

## Test Plan

- `tox -e test-unit` passes, including new tests:
  - `tests/unit/backends/openai/test_realtime_ws.py`: the backend records send/receive timestamps (counts, sums, last-sent) over an in-process WebSocket stub.
  - `tests/unit/schemas/test_request_stats.py`: the two properties compute correctly and return `None` without send timings.
  - `tests/unit/benchmark/schemas/test_metrics.py`: the metrics expose schema fields and compile into distributions.
- `tox -e lint-check` and `tox -e type-check` pass.
- Manual check on CPU (no GPU): drove `OpenAIWebSocketBackend.resolve()` against a fake realtime server with ~100 ms simulated per-token lag. Time To Last Round Trip came out around 300 ms and Avg RTT around 200 ms (matching the lag), and both were `None` for a non-WebSocket request. For a live run, point a benchmark at a vLLM realtime server (e.g. Voxtral) and confirm both metrics appear in the console latency table and `benchmarks.json`.

## Related Issues

- Resolves #832

---

## Use of AI

---

# git log

commit 457d56e683b9b80cf9078daa7da1cd5d8a87f565
Author: Suraj Singh <surajrider@gmail.com>
Date:   Wed Jun 24 15:00:52 2026 -0700

    test(openai): add failing websocket round-trip metric tests (#832)
    
    Add RED unit tests for Time To Last Round Trip and Avg RTT on the openai_websocket backend (send/receive timestamp recording, computed properties, None-gating, and metric compile). They fail until the implementation lands.
    
    Assisted-by: Claude Code
    Signed-off-by: Suraj Singh <surajrider@gmail.com>

commit 81a0aee1238edff4ec1338e5df0f344e772a66db
Author: Suraj Singh <surajrider@gmail.com>
Date:   Wed Jun 24 15:00:52 2026 -0700

    feat(openai): record websocket send/receive timestamps (#832)
    
    Add scalar round-trip timing fields to RequestTimings and record per-frame sent timestamps and received content-token timestamps in the openai_websocket backend. These feed the Time To Last Round Trip and Avg RTT metrics; HTTP backends leave them unset.
    
    Assisted-by: Claude Code
    Signed-off-by: Suraj Singh <surajrider@gmail.com>

commit 94f941c55d10760306f86145d41a5eabe62c2228
Author: Suraj Singh <surajrider@gmail.com>
Date:   Wed Jun 24 15:00:52 2026 -0700

    feat(benchmark): compute and aggregate websocket round-trip metrics (#832)
    
    Add time_to_last_round_trip_ms and avg_round_trip_time_ms as computed request stats (None unless send timings exist), aggregate them in GenerativeMetrics.compile(), and accumulate them for live progress. Avg RTT is an approximation of the mean send-to-receive lag.
    
    Assisted-by: Claude Code
    Signed-off-by: Suraj Singh <surajrider@gmail.com>

commit c5efcaf71e79d0519f0df52ec0b655a075b337e1
Author: Suraj Singh <surajrider@gmail.com>
Date:   Wed Jun 24 15:00:52 2026 -0700

    feat(benchmark): show websocket round-trip metrics in console and CSV (#832)
    
    Surface Time To Last Round Trip and Avg RTT in the request latency console table and the CSV export. JSON/YAML already include them via the schema; the HTML report is left for a follow-up.
    
    Assisted-by: Claude Code
    Signed-off-by: Suraj Singh <surajrider@gmail.com>

commit da70077412fd9d1f0ddb51610f6f6c779ca584a1
Author: Suraj Singh <surajrider@gmail.com>
Date:   Wed Jun 24 15:00:53 2026 -0700

    docs(metrics): document websocket round-trip metrics (#832)
    
    Describe Time To Last Round Trip and Avg RTT in the metrics guide, noting they are websocket-only and that Avg RTT is an approximation.
    
    Assisted-by: Claude Code
    Signed-off-by: Suraj Singh <surajrider@gmail.com>

commit ee76dd2cdf8151af5394a8c75a20c2cb41936437
Author: Suraj Singh <surajrider@gmail.com>
Date:   Tue Aug 11 09:59:50 2026 -0700

    fix(outputs): hide realtime-only metrics from console
    
    Signed-off-by: Suraj Singh <surajrider@gmail.com>

---------

Assisted-by: Claude Code
Signed-off-by: Suraj Singh <surajrider@gmail.com>

### DIFF
--- a/src/guidellm/benchmark/outputs/console.py
+++ b/src/guidellm/benchmark/outputs/console.py
@@ -480,6 +480,16 @@ class GenerativeBenchmarkerConsole(GenerativeBenchmarkerOutput):
                 group="TPOT",
                 name="ms",
             )
+            columns.add_stats(
+                benchmark.metrics.time_to_last_round_trip_ms,
+                group="Last RT",
+                name="ms",
+            )
+            columns.add_stats(
+                benchmark.metrics.avg_round_trip_time_ms,
+                group="Avg RTT",
+                name="ms",
+            )
 
         headers, values = columns.get_table_data()
         self.console.print("\n")
--- a/src/guidellm/benchmark/outputs/console.py
+++ b/src/guidellm/benchmark/outputs/console.py
@@ -480,17 +480,6 @@ class GenerativeBenchmarkerConsole(GenerativeBenchmarkerOutput):
                 group="TPOT",
                 name="ms",
             )
-            columns.add_stats(
-                benchmark.metrics.time_to_last_round_trip_ms,
-                group="Last RT",
-                name="ms",
-            )
-            columns.add_stats(
-                benchmark.metrics.avg_round_trip_time_ms,
-                group="Avg RTT",
-                name="ms",
-            )
-
         headers, values = columns.get_table_data()
         self.console.print("\n")
         self.console.print_table(
--- a/src/guidellm/schemas/request_stats.py
+++ b/src/guidellm/schemas/request_stats.py
@@ -164,6 +164,45 @@ class GenerativeRequestStats(StandardBaseDict):
 
     @computed_field  # type: ignore[misc]
     @property
+    def time_to_last_round_trip_ms(self) -> float | None:
+        """
+        Time from the last sent packet to the last received token in milliseconds.
+
+        Only populated by the websocket backend, which records send timestamps;
+        None for backends that do not record sends.
+
+        :return: Last round-trip latency in milliseconds, or None if unavailable
+        """
+        last_received = self.info.timings.last_token_iteration
+        last_sent = self.info.timings.last_request_sent
+        if last_received is None or last_sent is None:
+            return None
+
+        return 1000 * (last_received - last_sent)
+
+    @computed_field  # type: ignore[misc]
+    @property
+    def avg_round_trip_time_ms(self) -> float | None:
+        """
+        Approximate average round-trip time in milliseconds.
+
+        Computed as the mean of received content-token timestamps minus the mean
+        of sent-packet timestamps. This is an approximation that assumes sent
+        packets and received tokens align uniformly in time. Only populated by
+        the websocket backend; None otherwise.
+
+        :return: Average round-trip time in milliseconds, or None if unavailable
+        """
+        timings = self.info.timings
+        if timings.request_sent_count <= 0 or timings.token_received_count <= 0:
+            return None
+
+        mean_sent = timings.request_sent_sum / timings.request_sent_count
+        mean_received = timings.token_received_sum / timings.token_received_count
+        return 1000 * (mean_received - mean_sent)
+
+    @computed_field  # type: ignore[misc]
+    @property
     def time_per_output_token_ms(self) -> float | None:
         """
         Average time per output token in milliseconds including first token.
--- a/src/guidellm/schemas/request_stats.py
+++ b/src/guidellm/schemas/request_stats.py
@@ -172,6 +172,45 @@ class GenerativeRequestStats(StandardBaseDict):
 
     @computed_field  # type: ignore[misc]
     @property
+    def time_to_last_round_trip_ms(self) -> float | None:
+        """
+        Time from the last sent packet to the last received token in milliseconds.
+
+        Only populated by the websocket backend, which records send timestamps;
+        None for backends that do not record sends.
+
+        :return: Last round-trip latency in milliseconds, or None if unavailable
+        """
+        last_received = self.info.timings.last_token_iteration
+        last_sent = self.info.timings.last_request_sent
+        if last_received is None or last_sent is None:
+            return None
+
+        return 1000 * (last_received - last_sent)
+
+    @computed_field  # type: ignore[misc]
+    @property
+    def avg_round_trip_time_ms(self) -> float | None:
+        """
+        Approximate average round-trip time in milliseconds.
+
+        Computed as the mean of received content-token timestamps minus the mean
+        of sent-packet timestamps. This is an approximation that assumes sent
+        packets and received tokens align uniformly in time. Only populated by
+        the websocket backend; None otherwise.
+
+        :return: Average round-trip time in milliseconds, or None if unavailable
+        """
+        timings = self.info.timings
+        if timings.request_sent_count <= 0 or timings.token_received_count <= 0:
+            return None
+
+        mean_sent = timings.request_sent_sum / timings.request_sent_count
+        mean_received = timings.token_received_sum / timings.token_received_count
+        return 1000 * (mean_received - mean_sent)
+
+    @computed_field  # type: ignore[misc]
+    @property
     def time_per_output_token_ms(self) -> float | None:
         """
         Average time per output token in milliseconds including first token.
--- a/docs/guides/metrics.md
+++ b/docs/guides/metrics.md
@@ -71,6 +71,16 @@ These metrics provide a breakdown of the overall request statuses, helping users
 - **Definition**: The average time taken to generate each output token, including the first token.
 - **Use Case**: Provides a detailed view of the model's token generation efficiency.
 
+### Time To Last Round Trip
+
+- **Definition**: For the realtime WebSocket backend (`openai_websocket`), the time from the last sent packet to the last received token.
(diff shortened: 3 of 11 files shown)


dreamer-89:
@sjmonson @dbutenhof @jaredoconnell Request for review

dreamer-89:
@sjmonson @dbutenhof @jaredoconnell Request for review

jaredoconnell (dismissed):
I tested it and it appears to work as intended.

src/guidellm/schemas/request_stats.py
```diff
@@ -162,6 +162,45 @@ def time_to_first_token_ms(self) -> float | None:
...
+
+        return 1000 * (last_received - last_sent)
+
+    @computed_field  # type: ignore[misc]
+    @property
+    def avg_round_trip_time_ms(self) -> float | None:
+        """
+        Approximate average round-trip time in milliseconds.
+
+        Computed as the mean of received content-token timestamps minus the mean
```
dbutenhof, line 197: Mean of timestamps? Odd idea, though an efficient approximation of round-trip average as long as "sent packets and received tokens align uniformly in time" really holds true most of the time... and certainly a lot easier than tracking the actual average.

dbutenhof (dismissed):
Aside from some discomfort about the assumption of uniformity that makes your short-cut algorithm viable, I don't really have a problem with this. I'm posting with the in-line comment mostly to block our auto-merge until Sam has a chance to take a look.

src/guidellm/schemas/request_stats.py
dreamer-89: Thanks, that’s a fair concern. This relies on the uniform-alignment assumption from #832. Since packets and tokens are not one-to-one, calculating the actual average would require matching each token to its source packet. That’s why the metric is labeled approximate.

dreamer-89:
@sjmonson Request for review

dreamer-89:
@sjmonson : Request for review

src/guidellm/benchmark/outputs/console.py
```diff
@@ -480,6 +480,16 @@ def print_request_latency_table(self, report: GenerativeBenchmarksReport):
...
+            columns.add_stats(
+                benchmark.metrics.time_to_last_round_trip_ms,
+                group="Last RT",
+                name="ms",
+            )
+            columns.add_stats(
+                benchmark.metrics.avg_round_trip_time_ms,
+                group="Avg RTT",
+                name="ms",
+            )
```
sjmonson: Can you remove these from the console output for now since these metrics are only populated in realtime. In the future we will try to do some improvements to the console output and metric handlers so that workload specific metrics work better, but for now I don't like these being here when they are always 0 for non-realtime.

sjmonson (changes requested):
Apologies. I was away for a bit and am still catching up. See one nit below.

src/guidellm/benchmark/outputs/console.py
dreamer-89: Thanks @sjmonson for the review. I removed the mentioned columns from the output.

jaredoconnell (approved):

sjmonson (approved):

PR #868, opened 2026-06-27
docs: add governed endpoint benchmark example

cerebrixos:
This adds a docs example for benchmarking a governed OpenAI-compatible endpoint with GuideLLM.

Why this felt useful here:
- GuideLLM already targets OpenAI-compatible endpoints through `openai_http`
- many teams benchmark a raw model server, but deploy a more complex production path
- this example shows how to measure the user-visible latency and throughput of the full governed path without changing GuideLLMs normal backend model

The concrete example uses Tuning Engines, but the workflow applies to any OpenAI-compatible control plane or gateway.

---

# git log

commit c97fdd7e7d8bff0238f5ddf79e846f81c277ce95
Author: cerebrixos <248534569+cerebrixos@users.noreply.github.com>
Date:   Sat Jun 27 10:49:19 2026 -0400

    docs: add governed endpoint benchmark example

### DIFF
--- a/docs/examples/benchmark_tuning_engines.md
+++ b/docs/examples/benchmark_tuning_engines.md
@@ -1,0 +1,78 @@
+---
+weight: -3
+---
+
+# Benchmark a governed OpenAI-compatible endpoint
+
+GuideLLM can benchmark any OpenAI-compatible target, including a governed
+endpoint such as Tuning Engines. This is useful when you want to measure not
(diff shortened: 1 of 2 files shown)


cerebrixos:
I opened #869 as a signed-off replacement for this PR so the DCO check can pass cleanly. Closing this one to keep review on the compliant branch.

PR #874, opened 2026-06-30
Loosen torchcodec requirement

sjmonson:
## Summary

Torchcodec no longer requires a strict PyTorch version. Downstream builds of GuideLLM only have torchcodec 0.11 available so make that our minimum.

## Details

As of 0.12, `torchcodec` has switched to using the PyTorch 2.11 stable ABI (https://github.com/meta-pytorch/torchcodec/pull/1260) which means that we no longer have to lock version-to-version. Due to the downstream requirement (see above) we lock to a minimum of `torchcodec` 0.12 which requires `torch==2.11`. Technically this means that a user could end up with a broken install if they bump torch past 2.11 without bumping torchcodec but that is unlikely.

## Test Plan

Testing with any audio dataset should verify that this works.

---

## Use of AI

---

# git log

commit a5cbf0f3670dfcf383098b434a04e7abcb1b048b
Date:   Tue Jun 30 18:15:57 2026 -0400

    Bump torchcodec (for the final time)
    
---------

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2,12 +2,17 @@ version = 1
 revision = 3
 requires-python = ">=3.10.0, <4.0"
 resolution-markers = [
-    "python_full_version >= '3.12' and sys_platform != 'darwin'",
-    "python_full_version == '3.11.*' and sys_platform != 'darwin'",
-    "python_full_version >= '3.12' and sys_platform == 'darwin'",
-    "python_full_version == '3.11.*' and sys_platform == 'darwin'",
-    "python_full_version < '3.11' and sys_platform != 'darwin'",
-    "python_full_version < '3.11' and sys_platform == 'darwin'",
+    "(python_full_version >= '3.12' and platform_machine != 'arm64') or (python_full_version >= '3.12' and sys_platform != 'darwin')",
+    "(python_full_version == '3.11.*' and platform_machine != 'arm64') or (python_full_version == '3.11.*' and sys_platform != 'darwin')",
+    "python_full_version >= '3.12' and platform_machine == 'arm64' and sys_platform == 'darwin'",
+    "python_full_version == '3.11.*' and platform_machine == 'arm64' and sys_platform == 'darwin'",
+    "(python_full_version < '3.11' and platform_machine != 'arm64') or (python_full_version < '3.11' and sys_platform != 'darwin')",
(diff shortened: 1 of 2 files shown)


dbutenhof (commented):
What happened here? The ARM64 fix got backported to 0.12???

dbutenhof (approved):

sjmonson:
> What happened here? The ARM64 fix got backported to 0.12???

Just for documentation sake: Its possible to build earlier versions of torchcodec for aarch64 there we just not packages available. By locking to a minimum of 0.12 a user's local pip will resolve properly for the intended platform. `uv` is configured to lock across all platforms so it will automatically determine the version that makes the most sense for each as long as it meets the pyproject constraint.

uv.lock
```diff
@@ -591,10 +596,10 @@ wheels = [
...
-    { name = "torch", version = "2.11.0+cpu", source = { registry = "https://download.pytorch.org/whl/cpu" }, marker = "sys_platform != 'darwin'" },
-    { name = "torchcodec", version = "0.13.0", source = { registry = "https://download.pytorch.org/whl/cpu" }, marker = "sys_platform == 'darwin'" },
-    { name = "torchcodec", version = "0.13.0+cpu", source = { registry = "https://download.pytorch.org/whl/cpu" }, marker = "sys_platform != 'darwin'" },
+    { name = "torch", version = "2.11.0", source = { registry = "https://download.pytorch.org/whl/cpu" }, marker = "platform_machine == 'arm64' and sys_platform == 'darwin'" },
+    { name = "torch", version = "2.11.0+cpu", source = { registry = "https://download.pytorch.org/whl/cpu" }, marker = "platform_machine != 'arm64' or sys_platform != 'darwin'" },
+    { name = "torchcodec", version = "0.13.0", source = { registry = "https://download.pytorch.org/whl/cpu" }, marker = "platform_machine == 'arm64' and sys_platform == 'darwin'" },
+    { name = "torchcodec", version = "0.13.0+cpu", source = { registry = "https://download.pytorch.org/whl/cpu" }, marker = "platform_machine != 'arm64' or sys_platform != 'darwin'" },
```
dbutenhof, line 602: Shouldn't this be relaxed, too?

sjmonson: No and it can't be. `uv.lock` is always the exact wheel as it exists on the target index. AIPCC already has to rebuild (or ignore) the `uv.lock` as it targets PYPI and the PyTorch index.

PR #897, opened 2026-07-05
Fix explicit fill-value overwrite detection in arg parser

### LINKED ISSUE #896: Arg string parser overwrites explicit values equal to sparse fill value
## Describe the bug

`ArgStringParser` uses equality with `fill_value` to decide whether a sparse list position is unassigned. An explicitly parsed value equal to that fill value is therefore treated as an empty slot and can be silently overwritten despite `allow_overwrite=False`.

This affects the default `None` fill value and custom fill values.

## Reproduction

```python
from guidellm.utils import arg_string

result = arg_string.loads("items[0]=null,items[0]=replacement")
print(result)
```

Actual result:

```python
{"items": ["replacement"]}
```

The same problem occurs when converting an explicitly assigned null into a dictionary:

```python
arg_string.loads("items[0]=null,items[0].field=value")
```

## Expected behavior

Both inputs should raise `ArgStringParseError`, consistent with duplicate non-null list assignments and the default `allow_overwrite=False` behavior.

Sparse positions created as padding should remain assignable.

## Environment

- GuideLLM: current `main` (`948c3afd`)
- Python: 3.10.7
- OS: macOS arm64

## Proposed fix

Track explicitly assigned list paths separately from user-visible sparse fill values during decoding, with regression coverage for default and custom fill values.

harivilasp:
## Summary
Prevent the argument-string parser from treating explicitly assigned list values as sparse padding when they equal the configured fill value.

## Details
- Track list paths explicitly assigned during a single decode separately from sparse fill values.
- Preserve overwrite protection for explicit `null` values and custom fill values such as `0`.
- Preserve assignment into list positions that are genuinely sparse padding.
- Add regression tests for duplicate assignments and scalar-to-dictionary replacement.

## Test Plan
- `tox -e test-unit -- tests/unit/utils/test_arg_string.py -q` - 53 passed
- `tox -q` - `lint-check` passed; `type-check` failed in `.tox/type-check/lib/python3.14/site-packages/numpy/__init__.pyi:737` with `Type statement is only supported in Python 3.12 and greater`

## Related Issues
- Resolves #896

---

# git log

commit eacde439ba11111645441f3a07b0d73b168e2ffe
Author: harivilasp <harivilasp@gmail.com>
Date:   Sun Jul 5 14:02:05 2026 -0700

    Fix fill-value overwrite detection in arg parser
    
    Track explicit list assignments separately from sparse fill values so null and custom fill values retain normal overwrite protection.
    
    Generated-by: Codex GPT-5
    Signed-off-by: harivilasp <harivilasp@gmail.com>

---------

Generated-by: Codex GPT-5
Signed-off-by: harivilasp <harivilasp@gmail.com>

### DIFF
--- a/src/guidellm/utils/arg_string.py
+++ b/src/guidellm/utils/arg_string.py
@@ -27,6 +27,8 @@ __all__ = [
     "loads",
 ]
 
+_ListValuePaths = set[tuple[tuple[str, int | None], ...]]
+
 
 class ArgStringParseError(Exception):
     """Exception raised for errors during argument string parsing."""
@@ -109,6 +111,7 @@ class ArgStringParser:
             # {'items': [{'weight': 10, 'count': 5}]}
         """
         result: dict[str, Any] = {}
+        assigned_list_values: _ListValuePaths = set()
 
         if not s or not s.strip():
             return result
@@ -145,7 +148,12 @@ class ArgStringParser:
             parsed_value = yaml.safe_load(value)
 
             # Set the value in the nested structure
-            self._value_set_nested(result, segments, parsed_value)
+            self._value_set_nested(
+                result,
+                segments,
+                parsed_value,
+                assigned_list_values=assigned_list_values,
+            )
 
         return result
 
@@ -250,6 +258,8 @@ class ArgStringParser:
         name: str,
         index: int,
         value: Any,
+        *,
+        assigned: bool = False,
     ) -> None:
         """
         Set a value in a list at the given index within a dictionary.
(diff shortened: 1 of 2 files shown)


jaredoconnell (approved):
It looks good to me, blocking an overwrite when none is allowed.

jaredoconnell:
It appears that CI is failing due to a handshake error that appears to be unrelated to your changes.

harivilasp:
@jaredoconnell  do you mind checking why it is not getting merged "Cannot update this protected ref." all checks are passed now

dbutenhof (approved):
This seems to work -- even if the detailed error message isn't getting through the CLI. (Not a problem with this PR, since the "Invalid key=value expression" error for `value:x` is also getting lost, and we only get a generic "Input should be a valid dictionary or object".)

dbutenhof:
> @jaredoconnell do you mind checking why it is not getting merged "Cannot update this protected ref." all checks are passed now

FYI -- that was just because our CI requires two core group approvals to trigger automatic merging. 😁

PR #906, opened 2026-07-06
Support ShareGPT datasets

jaredoconnell:
## Summary

Adds support for ShareGPT datasets. This is a format commonly used for tool call datasets.

These changes were pulled from my prior PR.

This change highlights some limitations of the current design, and as such I've put it in draft in case we decide to refactor the code. The current column mapper design results in two column mapping steps, basically.

## Details

- Allows alternative preprocessing so GuideLLM can read it.

## Test Plan

- Run the examples added in the docs.

---

## Use of AI

---

# git log

commit 613dc9a5839f461f3781bda9aca11e5c446ae887
Date:   Mon Jul 6 18:05:25 2026 -0400

    Support ShareGPT datasets
    
---------

### DIFF
--- a/src/guidellm/data/preprocessors/mappers.py
+++ b/src/guidellm/data/preprocessors/mappers.py
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
+import json
 import re
 from collections import defaultdict
 from typing import Any, ClassVar, Literal, TypeAlias, cast
 
-from datasets import Dataset, IterableDataset
+from datasets import Dataset, DatasetDict, IterableDataset, IterableDatasetDict
 from pydantic import Field
 
 from guidellm.data.preprocessors.preprocessor import (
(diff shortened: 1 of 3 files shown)


src/guidellm/data/preprocessors/mappers.py
```diff
@@ -29,6 +30,86 @@
...
+    """Unwrap a DatasetDict/IterableDatasetDict into a single split.
+
+    Prefers the ``"train"`` split if available, otherwise picks the first split.
+    Returns the input unchanged if it is already a single Dataset/IterableDataset.
+
+    :param dataset: The dataset or dataset dict to unwrap.
+    :return: A single Dataset or IterableDataset.
+    """
+    if isinstance(dataset, DatasetDict | IterableDatasetDict):
+        if "train" in dataset:
```
dbutenhof, line 45: This is great, although I suspect we should do this earlier when we first load the dataset. E.g., if you don't specify `load_kwargs.split=<name>` on `--data`, and the dataset *has* splits, apply this unwrapping right off rather than deferring it to here.

src/guidellm/data/preprocessors/mappers.py
```diff
@@ -29,6 +30,86 @@
...
+        return None
+
+    candidate = dataset_columns[0]
+    sample = next(iter(dataset))
+    value = sample[candidate]
+    if not isinstance(value, str):
+        return None
+
+    try:
+        parsed = json.loads(value)
```
dbutenhof, line 70: Three separate methods using `json.loads` -- can we avoid the duplication by consolidating some of these paths??

dbutenhof (changes requested):
I've seen the unhandled errors when a dataset has splits we don't process -- it's ugly. I like the idea of assuming a default split, and "train" is certainly commin -- though I'm not sure all split datasets necessarily have a "train" split, and we should handle that case gracefully as long as we're trying to be graceful.

1. I think we should handle this default unwrapping earlier, when we load the data -- basically, in the absence of `load_kwargs.split=<name>` (when we receive a split dataset) we should resolve it right there.
2. If "train" is not in the split keys, we could use an arbitrary split that *does* exist or perhaps we should instead raise an error exception advising the user that they need to specify a split for that dataset.

src/guidellm/data/preprocessors/mappers.py
```diff
@@ -29,6 +30,86 @@
...
+    """Check if a dataset has a single string column containing JSON dicts.
+
+    :param dataset: The dataset to inspect.
+    :param dataset_columns: The column names present in the dataset.
+    :return: The wrapper column name if detected, or None.
+    """
+    if len(dataset_columns) != 1:
+        return None
+
+    candidate = dataset_columns[0]
```
SkiHatDuckie, line 63: While this will work for ShareGPT, there's no guarantee that the column(s!) containing JSON-dicts in other dataset formats will be in the first column.

src/guidellm/data/preprocessors/mappers.py
```diff
@@ -196,34 +287,72 @@ def datasets_mappings(
...
 
-                column_pattern = cls.column_name_pattern.format(
-                    name="|".join(re.escape(n) for n in filtered_names)
-                )
-                # Find the first matching column name
-                base_match = re.search(column_pattern, dataset_columns_str, re.M | re.I)
-                if not base_match:
-                    continue
+            # Fallback: if no matches found, try JSON unwrapping
+            if not matched:
```
SkiHatDuckie, line 295: This won't work as intended if some, but not all of the target columns are found before checking wrappers.

src/guidellm/data/preprocessors/mappers.py
```diff
@@ -163,31 +247,38 @@ def datasets_mappings(
...
         Resolve column mappings across one or more datasets.
 
         For each dataset, matches actual column names against the requested
         mapping names (or :attr:`defaults`) using regex patterns that account
         for pluralisation and turn suffixes (e.g. ``prompt-0``, ``prompt-1``).
 
+        When a dataset has no direct column matches but contains a single
+        JSON-string column, the inner keys of that JSON are used as virtual
+        column names and matching is retried.
+
```
SkiHatDuckie, line 261: If we wish to standardize the JSON unwrapper/detector for use across GuideLLM, we'll have to make as few assumptions about the location(s) of target columns as possible.

A future dataset might have multiple columns with JSON-wrapped dicts, for example.

SkiHatDuckie (commented):
I haven't ran anything yet, but I did take a look at the code involved in detecting/unwrapping JSON dictionary columns. Regardless of what we choose to do with the column mapper, it'll help to make the unwrapper as general-purpose as possible, so that it may hopefully be useful regardless of the situation it's called in.

Edit: In other words, I think the JSON unwrapper has potential to be it's own utility function(s).

PR #912, opened 2026-07-07
Correct the CSV report format

### LINKED ISSUE #764: SecretStr api_key crashes report serialization (no output when api_key is set)

dbutenhof:
## Summary

Once again fix `api_key` serialization (this time for CSV).

## Details

The root problem stems from attempts to serialize the Pydantic `SecretStr`.

This was resolved for JSON format by serializing `BenchmarkConfig` using `model_dump(mode="json")`, but the CSV serialization breaks up the context very differently, leading to solutions that are either very specific or rather sweeping.

It appears the simplest and least invasive fix is to call `model_dump(mode="json")` on the `OpenAIHTTPBackend.info` method. This is the only place we use `SecretStr`, and feeds the JSON we serialize to CSV. `info` is only intended for "informational" serialization, so the masked password we receive is expected and correct.

We could benefit in the future from some refactoring. For example, it's odd that we convert the `Backend` to a `dict` when creating `BenchmarkConfig`, but then reconstruct the `Backend` when we call the benchmarker `run` method for each strategy.

## Test Plan

- [x] test suite
- [x] manually run CSV output with `--backend kind=openai_http,api_key=<value>`

## Related Issues

- Resolves #764 

---

## Use of AI

---

# git log

commit e99d9a3792d9e5836048dae59a04c920419d75c8
Date:   Tue Jul 7 14:08:50 2026 -0400

    Correct the CSV report format
    
    The root problem stems from attempts to serialize the Pydantic `SecretStr`.
    
    This was resolved for JSON format by serializing `BenchmarkConfig`
    using `model_dump(mode="json")`, but the CSV serialization breaks up the
    context very differently, leading to solutions that are either very specific
    or rather sweeping.
    
    It appears the simplest and least invasive fix is to call
    `model_dump(mode="json")` on the `OpenAIHTTPBackend.info` method, which is the
    only place we use `SecretStr`.
    
    We could benefit in the future from some refactoring. For example, it's odd
    that we convert the `Backend` to a `dict` when creating `BenchmarkConfig`,
    but then reconstruct the `Backend` when we call the benchmarker `run` method
    for each strategy.
    
commit ba2aa233d8de95f7b58880b9b4c2f056e83edadf
Date:   Tue Jul 7 15:40:27 2026 -0400

    Generalize the change
    
commit 33ed889dfd3628969d59ab464def308c9c51e430
Date:   Tue Jul 7 16:09:14 2026 -0400

    Yeah, that confused ruff
    
    Rather than annotate every args field access, we need to set `_args` as the
    proper type to support static type checking. (For some reason I keep thinking
    I'm writing in Python, when I'm actually writing in ruff...)
    
commit 3b8e599ccd4a2a83cd72cfbc4261ca4282bc76fd
Date:   Tue Jul 7 16:44:26 2026 -0400

    Use the type hint, Luke!
    
---------

### DIFF
--- a/src/guidellm/backends/backend.py
+++ b/src/guidellm/backends/backend.py
@@ -9,7 +9,7 @@ provide a standard interface for distributed execution across worker processes.
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import ClassVar
+from typing import Any, ClassVar
 
 from pydantic import Field
 
@@ -125,6 +125,14 @@ class Backend(
         :param type_: The backend type identifier
         """
         self.kind = args.kind
+        self._args = args
+
+    @property
+    def info(self) -> dict[str, Any]:
+        """
+        Return a snapshot of backend configuration for logging or debugging.
+        """
+        return self._args.model_dump(mode="json")
 
     @property
     def processes_limit(self) -> int | None:
--- a/src/guidellm/backends/backend.py
+++ b/src/guidellm/backends/backend.py
@@ -122,17 +122,17 @@ class Backend(
         """
         Initialize a backend instance.
 
-        :param type_: The backend type identifier
+        :param args: The backend arguments object
         """
         self.kind = args.kind
-        self._args = args
 
     @property
     def info(self) -> dict[str, Any]:
         """
         Return a snapshot of backend configuration for logging or debugging.
         """
-        return self._args.model_dump(mode="json")
+        args = getattr(self, "_args", None)
+        return args.model_dump(mode="json") if args else {}
 
     @property
     def processes_limit(self) -> int | None:
--- a/src/guidellm/backends/backend.py
+++ b/src/guidellm/backends/backend.py
@@ -96,6 +96,8 @@ class Backend(
         backend = Backend.create(args)
     """
 
+    _args: BackendArgs
+
     @classmethod
     def create(cls, args: BackendArgs) -> Backend:
         """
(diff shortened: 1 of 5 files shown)


sjmonson (commented):
Seems like a reasonable fix to me but should it get applied to the other backends as well for constancy?

jaredoconnell (dismissed):
Looks good. If there are other places this should get applied, as Sam mentioned, let's apply it to those cases, too.

dbutenhof:
> Seems like a reasonable fix to me but should it get applied to the other backends as well for constancy?

Fair ... I kept it simple since we only have the "complicated" container object inside HTTP. (And then I read the LiteLLM PR, which will have the same issue.) If I was going a step further I might be strongly tempted to factor `_args` and `info` down into `Backend` instead of duplicating this. I resisted the temptation. 😆

sjmonson (dismissed):

dbutenhof:
> > Seems like a reasonable fix to me but should it get applied to the other backends as well for constancy?
> 

"Amusingly" I discovered that the web socket backend was storing its `api_key` as a `str`; so I fixed that along the way...

This was stupidly complicated by static type checking, since I can't just store `self._args` in the `Backend` constructor like I could in Python. 😆

sjmonson:
I really don't like `getattr` unless its absolutely necessary. So maybe just move the implementation back to each backend?

dbutenhof:
> I really don't like `getattr` unless its absolutely necessary. So maybe just move the implementation back to each backend?

😞 -- this is already a mess just to work around static type checking and make it behave like Python. I'm not a fan of routine `getattr` either, though I prefer it over duplicating code.

Sigh.

dbutenhof:
> > I really don't like `getattr` unless its absolutely necessary. So maybe just move the implementation back to each backend?
> 
> 😞 -- this is already a mess just to work around static type checking and make it behave like Python. I'm not a fan of routine `getattr` either, though I prefer it over duplicating code.

Oh, wait -- there's a better way to deal with the type checker. Foolish of me ...

sjmonson (approved):

jaredoconnell (approved):

PR #915, opened 2026-07-08
Add example: benchmarking with local tokenizer and custom JSONL dataset

miglis:
## Summary

benchmarking with local tokenizer and custom JSONL dataset

Adds docs/examples/custom-jsonl-dataset.md, showing how to run GuideLLM against an OpenAI-compatible endpoint using only local tokenizer files (tokenizer.json, tokenizer_config.json, special_tokens_map.json) and a custom JSONL prompt dataset.

---

## Use of AI

---

# git log

commit 166f012ab61e37cb59373125bc9677008ebcf627
Author: Michael Stucki <michael.stucki99@gmail.com>
Date:   Wed Jul 8 13:39:29 2026 +0200

    docs: add guide for benchmarking with local tokenizer and custom JSONL dataset
    
    Adds docs/examples/custom-jsonl-dataset.md, showing how to run
    GuideLLM against an OpenAI-compatible endpoint using only local tokenizer
    files (tokenizer.json, tokenizer_config.json, special_tokens_map.json)
    and a custom JSONL prompt dataset.
    
    Signed-off-by: Michael Stucki <michael.stucki99@gmail.com>

commit b065f945dc19d7edb1af015476a2b48adebe9be0
Author: Michael Stucki <michael.stucki99@gmail.com>
Date:   Thu Jul 9 15:30:50 2026 +0200

    docs: fix --data-loader value to defaults
    
    To keep example so readers understand the options exists.
    
    Signed-off-by: Michael Stucki <michael.stucki99@gmail.com>

commit cd6507a58ff6ff16446a153a21c345113e708886
Author: Michael Stucki <michael.stucki99@gmail.com>
Date:   Thu Jul 9 15:54:17 2026 +0200

    docs: replace obsolete env var with an example for json output file
    
    Signed-off-by: Michael Stucki <michael.stucki99@gmail.com>

commit c5b3c2caa405bcd7c86557b008d0f68ac00945ab
Author: Michael Stucki <michael.stucki99@gmail.com>
Date:   Thu Jul 9 16:02:51 2026 +0200

    chore: trigger pipeline
    
    Signed-off-by: Michael Stucki <michael.stucki99@gmail.com>

commit e2e7c63ca3bb1c981bdcd800be68a9df400c5dba
Author: Michael Stucki <michael.stucki99@gmail.com>
Date:   Thu Jul 9 16:03:26 2026 +0200

    chore: trigger pipeline
    
    Signed-off-by: Michael Stucki <michael.stucki99@gmail.com>

commit ed1b8ec75bae0fd368b3a142193f8a994f21f14e
Author: Michael Stucki <michael.stucki99@gmail.com>
Date:   Fri Jul 10 11:23:46 2026 +0200

    docs: clarify to copy necessary files if using model X
    
    Signed-off-by: Michael Stucki <michael.stucki99@gmail.com>

commit ebf524bbe50f406647ccd6f66b2ce3b6b86be7ee
Author: Michael Stucki <michael.stucki99@gmail.com>
Date:   Fri Jul 10 11:31:17 2026 +0200

    docs: state that this example is based on local restrictions and step 3 can be ignored
    
    Signed-off-by: Michael Stucki <michael.stucki99@gmail.com>

commit aaa9cc90c3c259621915fab75b0f03a610f2a331
Author: Michael Stucki <michael.stucki99@gmail.com>
Date:   Fri Jul 10 11:34:29 2026 +0200

    docs: explicitly mention --data-loader-defaults
    
    Signed-off-by: Michael Stucki <michael.stucki99@gmail.com>

commit 0c26bd2d72dda1e0c87c11435339f86a1eae8c3a
Author: Michael Stucki <michael.stucki99@gmail.com>
Date:   Sun Jul 12 20:36:06 2026 +0200

    docs: remove all docker related parts
    
    Replaced with commands to run guidellm manually.
    
    Signed-off-by: Michael Stucki <michael.stucki99@gmail.com>

commit 160c9055c202dd088b220699bf5506c137a9c430
Author: Michael Stucki <michael.stucki99@gmail.com>
Date:   Sun Jul 12 20:42:16 2026 +0200

    docs: remove defaults
    
    Signed-off-by: Michael Stucki <michael.stucki99@gmail.com>

commit d2d57070b48c5de8de88d29e3d96b39fc77de92b
Author: Michael Stucki <michael.stucki99@gmail.com>
Date:   Sun Jul 12 20:46:12 2026 +0200

    docs: clarify tokenizer file list is model-specific, not universal
    
    Merge the "copy only these three files" statement with the Mistral-family
    caveat into one coherent paragraph, removing the contradiction between them.
    
    Signed-off-by: Michael Stucki <michael.stucki99@gmail.com>

commit b6661cba90b4af195838df94a4aaeb3aca28da79
Author: Michael Stucki <michael.stucki99@gmail.com>
Date:   Sun Jul 12 20:58:42 2026 +0200

    docs: fix markdown formatting per mdformat
    
    Signed-off-by: Michael Stucki <michael.stucki99@gmail.com>

---------

Signed-off-by: Michael Stucki <michael.stucki99@gmail.com>

### DIFF
--- a/docs/examples/custom-jsonl-dataset.md
+++ b/docs/examples/custom-jsonl-dataset.md
@@ -60,9 +60,7 @@ guidellm run \
   --backend kind=openai_http,target=http://<PREDICTOR_URL> \
   --data '{"kind":"json_file","path":"/home/<USERNAME>/mistral/long_translation_prompts.jsonl","load_kwargs":{"split":"train"}}' \
   --tokenizer '{"kind":"huggingface_auto","model":"/home/<USERNAME>/mistral"}' \
-  --data-column-mapper '{"kind":"generative_column_mapper","column_mappings":{"text_column":"prompt"}}' \
   --profile kind=concurrent,streams=100 \
-  --data-loader kind=pytorch,samples=-1 \
   --output kind=json,path=/home/<USERNAME>/mistral/benchmarks.json
 ```
 
@@ -73,9 +71,7 @@ guidellm run \
 | `--backend kind=openai_http,target=...` | Points GuideLLM at your OpenAI-compatible predictor endpoint. |
 | `--data kind=json_file,path=...` | Loads prompts from the local JSONL file. |
 | `--tokenizer kind=huggingface_auto,model=/home/<USERNAME>/mistral` | Loads the tokenizer from the local directory, the 3 files from step 1, instead of downloading a model. |
-| `--data-column-mapper` | Maps the JSONL `prompt` field to the `text_column` GuideLLM expects internally. Optional here since `prompt` is already a default mapping. Shown explicitly for clarity. |
 | `--profile kind=concurrent,streams=100` | Simulates 100 concurrent "users" hitting the endpoint at once. |
-| `--data-loader kind=pytorch,samples=1000` | Reads 1000 lines or samples from the JSONL file for the run. |
 | `--output kind=json,path=/home/<USERNAME>/mistral/benchmarks.json` | Output file type and path. |
 
 ______________________________________________________________________
--- a/docs/examples/custom-jsonl-dataset.md
+++ b/docs/examples/custom-jsonl-dataset.md
@@ -1,0 +1,100 @@
+# GuideLLM Benchmark Testing with a Local Tokenizer and Custom Dataset
+
+Benchmark an already-deployed OpenAI-compatible model endpoint using GuideLLM with a local tokenizer and a custom JSONL prompt dataset, without needing a local copy of the model itself.
+
+## Getting Started
+
+### 1. Prepare the Tokenizer Files
+
+The model is already deployed and running, for example on GPU, served via vLLM or OpenShift, behind an OpenAI-compatible endpoint. GuideLLM never loads or runs the model, it only talks to it over HTTP. What GuideLLM does need locally is the tokenizer, since it uses it to compute token counts for its metrics such as prompt tokens, output tokens and throughput.
+
+From the model's repo, for example the HuggingFace Hub or wherever the model artifacts live, copy only these three files:
+
+```bash
+tokenizer.json
+tokenizer_config.json
+special_tokens_map.json
+```
+
+Place them in a local directory, e.g. `/home/<USERNAME>/mistral`. This directory will be mounted into the container and referenced as the tokenizer path.
+
+### 2. Prepare a Custom JSONL Dataset
+
+In the same directory, add your JSONL prompt file(s). Each line is a JSON object with a `prompt` field, for example:
+
+```json
+{"prompt": "Translate the following text in full to German, French and Spanish. Provide a fluent translation of the entire text without summarizing or omitting any sentences. Respond with the translation in the same format as the original text, including line breaks and whitespace. Text: ..."}
+```
+
+You can split your data into multiple files by length or category if that's useful for your benchmarking scenario, for example:
+
+```bash
+short_translation_prompts.jsonl
+medium_translation_prompts.jsonl
+long_translation_prompts.jsonl
+```
+
+After both steps, your local directory should look like this:
+
+```bash
+ls /home/<USERNAME>/mistral
+long_translation_prompts.jsonl  medium_translation_prompts.jsonl  short_translation_prompts.jsonl
+special_tokens_map.json  tokenizer_config.json  tokenizer.json
+```
+
+### 3. Start the GuideLLM Container
+
+Mount the local directory into the container and drop into a shell:
+
+```bash
+sudo docker run --rm -ti \
+  -e GUIDELLM_OUTPUT_DIR="/data/mistral" \
+  -e USER=guidellm \
+  --volume /home/<USERNAME>/mistral:/data/mistral \
+  --entrypoint bash \
+  guidellm:stable
+```
+
+Verify the install and mounted files inside the container:
+
+```bash
+(app-root) bash-5.2$ guidellm --version
+guidellm version: 0.7.0
+
+(app-root) bash-5.2$ ls /data/mistral/
+long_translation_prompts.jsonl  medium_translation_prompts.jsonl  short_translation_prompts.jsonl
+special_tokens_map.json  tokenizer_config.json  tokenizer.json
+```
+
+______________________________________________________________________
+
+## 4. Running the Benchmark
+
+```bash
+guidellm run \
+  --backend kind=openai_http,target=http://<PREDICTOR_URL> \
+  --data '{"kind":"json_file","path":"/data/mistral/long_translation_prompts.jsonl","load_kwargs":{"split":"train"}}' \
+  --tokenizer '{"kind":"huggingface_auto","model":"/data/mistral"}' \
+  --data-column-mapper '{"kind":"generative_column_mapper","column_mappings":{"text_column":"prompt"}}' \
+  --profile kind=concurrent,streams=100 \
+  --data-loader kind=pytorch,samples=1000
+```
+
+### What Each Argument Does
+
+| Argument | Purpose |
+|---|---|
+| `--backend kind=openai_http,target=...` | Points GuideLLM at your OpenAI-compatible predictor endpoint. |
+| `--data kind=json_file,path=...` | Loads prompts from the local JSONL file. |
+| `--tokenizer kind=huggingface_auto,model=/data/mistral` | Loads the tokenizer from the local directory, the 3 files from step 1, instead of downloading a model. |
+| `--data-column-mapper` | Maps the JSONL `prompt` field to the `text_column` GuideLLM expects internally. |
+| `--profile kind=concurrent,streams=100` | Simulates 100 concurrent "users" hitting the endpoint at once. |
+| `--data-loader kind=pytorch,samples=1000` | Reads 1000 lines or samples from the JSONL file for the run. |
+
+______________________________________________________________________
+
+## 5. Notes
+
+- The tokenizer path (`/data/mistral` in this example) only needs the 3 tokenizer files, no model weights required.
+- The dataset file path and the tokenizer path can point to different directories or models if you want to mix and match, though in this example they are the same folder for convenience.
+- Swap `--profile kind=concurrent,streams=N` for other GuideLLM profiles, such as `synchronous` or `throughput`, depending on the load pattern you want to test.
--- a/docs/examples/custom-jsonl-dataset.md
+++ b/docs/examples/custom-jsonl-dataset.md
@@ -77,7 +77,7 @@ guidellm run \
   --tokenizer '{"kind":"huggingface_auto","model":"/data/mistral"}' \
   --data-column-mapper '{"kind":"generative_column_mapper","column_mappings":{"text_column":"prompt"}}' \
   --profile kind=concurrent,streams=100 \
-  --data-loader kind=pytorch,samples=1000
+  --data-loader kind=pytorch,samples=-1
 ```
 
 ### What Each Argument Does
--- a/docs/examples/custom-jsonl-dataset.md
+++ b/docs/examples/custom-jsonl-dataset.md
@@ -9,8 +9,8 @@ Benchmark an already-deployed OpenAI-compatible model endpoint using GuideLLM wi
 
 The model is already deployed and running, for example on GPU, served via vLLM or OpenShift, behind an OpenAI-compatible endpoint. GuideLLM never loads or runs the model, it only talks to it over HTTP. What GuideLLM does need locally is the tokenizer, since it uses it to compute token counts for its metrics such as prompt tokens, output tokens and throughput.
 
-From the model's repo, for example the HuggingFace Hub or wherever the model artifacts live, copy only these three files.
-This example uses a Mistral-family model; other model architectures may require additional or different files, so treat this list as a starting point rather than a universal one:
+From the model's repo, for example the HuggingFace Hub or wherever the model artifacts live, copy the tokenizer files required by your specific model.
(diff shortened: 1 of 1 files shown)


docs/examples/custom-jsonl-dataset.md
```diff
@@ -0,0 +1,100 @@
...
+long_translation_prompts.jsonl  medium_translation_prompts.jsonl  short_translation_prompts.jsonl
+special_tokens_map.json  tokenizer_config.json  tokenizer.json
+```
+
+______________________________________________________________________
+
+## 4. Running the Benchmark
+
+```bash
+guidellm run \
```
dbutenhof, line 59: It feels a bit reductive to cast your example in terms of running a GuideLLM *container*, but then simply running `guidellm` manually inside the container. I think a full example of configuring and using the GuideLLM container would be valuable (we should probably write one); but in this case it'd be simpler and more concise to avoid the extra steps (which aren't directly relevant to the "running with local tokenizer & dataset") and simply install/run GuideLLM locally.

(With that said -- I'm commenting, but not formally objecting. It's not "wrong", and it's not "useless" -- just a few small steps more complicated than necessary. 😆)

docs/examples/custom-jsonl-dataset.md
```diff
@@ -0,0 +1,100 @@
...
+  --backend kind=openai_http,target=http://<PREDICTOR_URL> \
+  --data '{"kind":"json_file","path":"/data/mistral/long_translation_prompts.jsonl","load_kwargs":{"split":"train"}}' \
+  --tokenizer '{"kind":"huggingface_auto","model":"/data/mistral"}' \
+  --data-column-mapper '{"kind":"generative_column_mapper","column_mappings":{"text_column":"prompt"}}' \
+  --profile kind=concurrent,streams=100 \
+  --data-loader kind=pytorch,samples=1000
```
dbutenhof: It's important to note here that you'll get an error if the number of samples you request exceeds the size of your dataset. Your example has *one*, so `1000` samples will prove problematic. For this example I'd suggest sticking with the default (-1, which is "all") and omitting `--data-loader` entirely.

docs/examples/custom-jsonl-dataset.md
```diff
@@ -0,0 +1,100 @@
...
+______________________________________________________________________
+
+## 4. Running the Benchmark
+
+```bash
+guidellm run \
+  --backend kind=openai_http,target=http://<PREDICTOR_URL> \
+  --data '{"kind":"json_file","path":"/data/mistral/long_translation_prompts.jsonl","load_kwargs":{"split":"train"}}' \
+  --tokenizer '{"kind":"huggingface_auto","model":"/data/mistral"}' \
+  --data-column-mapper '{"kind":"generative_column_mapper","column_mappings":{"text_column":"prompt"}}' \
```
dbutenhof: This is again over-complicating things. `"prompt"` is one of the default mappings, so this is redundant and you'll get the same result if you simply omit `--data-column-mapper` entirely, which would simplify the example and focus more on what you're trying to demonstrate.

docs/examples/custom-jsonl-dataset.md
```diff
@@ -0,0 +1,100 @@
...
+### 1. Prepare the Tokenizer Files
+
+The model is already deployed and running, for example on GPU, served via vLLM or OpenShift, behind an OpenAI-compatible endpoint. GuideLLM never loads or runs the model, it only talks to it over HTTP. What GuideLLM does need locally is the tokenizer, since it uses it to compute token counts for its metrics such as prompt tokens, output tokens and throughput.
+
+From the model's repo, for example the HuggingFace Hub or wherever the model artifacts live, copy only these three files:
+
+```bash
+tokenizer.json
+tokenizer_config.json
+special_tokens_map.json
```
dbutenhof, line 16: Hmm. I've never tried doing this before, but some older models require additional files you haven't mentioned. I think we need to be very careful about narrowing these instructions too much. Specifics are great ... but not if they can be misleading. Tying it down to specific models or perhaps even "most modern models" might work, but as a generic list without constraining the model, some people attempting to follow this may have problems.

dbutenhof (commented):
There's some overlap with the `practice_on_vllm_simulator.md` example -- for example it talks about downloading a tokenizer to run locally, although it doesn't specify the particular files.

Generally, more concrete examples are good. My comments are generally in areas where you're either too abstract or unnecessarily over-specified (like using unnecessary options to specify defaults or values inconsistent with your example)...

docs/examples/custom-jsonl-dataset.md
```diff
@@ -0,0 +1,100 @@
...
+special_tokens_map.json  tokenizer_config.json  tokenizer.json
+```
+
+### 3. Start the GuideLLM Container
+
+Mount the local directory into the container and drop into a shell:
+
+```bash
+sudo docker run --rm -ti \
+  -e GUIDELLM_OUTPUT_DIR="/data/mistral" \
```
sjmonson: This environment variable no longer exists. Currently each output handles its own full path: `--output kind=json,path=/data/mistral/benchmarks.json`. I might add it back as a default fallback (though under a different name).

sjmonson (dismissed):

docs/examples/custom-jsonl-dataset.md
miglis: Fair point from a pure GuideLLM-usage perspective. But we run everything containerized here as an internal policy, so I wanted the example to reflect a realistic setup rather than a bare-metal install. 😆

docs/examples/custom-jsonl-dataset.md
miglis: I wanted to keep `--data-loader` explicit in the example so readers understand the option exists, so I'll fix the value to `samples=-1 `instead of removing it entirely.

docs/examples/custom-jsonl-dataset.md
miglis: It's redundant with the default, true. But we kept running into cases where it wasn't obvious _which_ column names are auto-detected, so we wanted this example to show the mapping explicitly rather than relying on implicit defaults. Could add a one-line comment noting it's technically optional here, if that helps clarify intent for other readers?

docs/examples/custom-jsonl-dataset.md
miglis: Thanks for catching that will be removed and replaced with `--output kind=json,path=/data/mistral/benchmarks.json`.

docs/examples/custom-jsonl-dataset.md
miglis: In our case this level of detail is actually intentional: we run an older model where these files need to be explicitly staged, and more generally this matters a lot in air-gapped environments, where you can't just let the tokenizer auto-download at runtime and need to know exactly which files to pre-stage. I think it's worth keeping that context visible for other readers in a similar setup.

dbutenhof: > we run an older model where these files need to be explicitly staged

Absolutely. My point was that this specific list assumes a certain class of model, and none of that is specified. This is cast as a "universal" list, and it's not. You can close that gap by specifying enough context to avoid risking misinformation here, or by generalizing the list. E.g., your directory name implies you're using a Mistral AI model -- tying this list to that use case explicitly might be enough.

docs/examples/custom-jsonl-dataset.md
dbutenhof: I'm not saying "context specific" examples are bad; but you've set this up as "here's how to benchmark with a local dataset and tokenizer" and you're throwing in additional context beyond that scope.

Again, this isn't bad necessarily, but I also don't want an unprepared reader to misunderstand exactly what's necessary and assumed.

Even a direct statement, perhaps in the introduction paragraph, that this example "happens" to show running GuideLLM in a local container but that choice isn't a requirement of the example.

docs/examples/custom-jsonl-dataset.md
dbutenhof: I'd prefer to keep it simple and not include "explicit defaults". If you're really determined, though, at least mention on your explanation bullet points which are redundantly specifying defaults and unnecessary to the example.

docs/examples/custom-jsonl-dataset.md
jaredoconnell: I agree. In this case, you can specify that "In the case of X model, you can copy these files".

docs/examples/custom-jsonl-dataset.md
```diff
@@ -2,13 +2,16 @@
 
 Benchmark an already-deployed OpenAI-compatible model endpoint using GuideLLM with a local tokenizer and a custom JSONL prompt dataset, without needing a local copy of the model itself.
 
+This example runs GuideLLM from a container, reflecting our internal workflow, but that's not a requirement. A local `pip install`-based setup works the same way and skips step 3.
```
dbutenhof: The context of "our" is going to get lost completely once this PR is merged. It's not "our" (GuideLLM's) internal workflow, but that's how the words will be interpreted in the future. Instead, try something more like "reflecting one particular workflow", or (probably better), just remove that phrase since it doesn't add anything to the sentence. The example just happens to run GuideLLM from a container, but that's not necessary to follow the example.

docs/examples/custom-jsonl-dataset.md
```diff
@@ -2,13 +2,16 @@
...
+
 ## Getting Started
 
 ### 1. Prepare the Tokenizer Files
 
 The model is already deployed and running, for example on GPU, served via vLLM or OpenShift, behind an OpenAI-compatible endpoint. GuideLLM never loads or runs the model, it only talks to it over HTTP. What GuideLLM does need locally is the tokenizer, since it uses it to compute token counts for its metrics such as prompt tokens, output tokens and throughput.
 
-From the model's repo, for example the HuggingFace Hub or wherever the model artifacts live, copy only these three files:
+From the model's repo, for example the HuggingFace Hub or wherever the model artifacts live, copy only these three files.
```
dbutenhof: I like the addition, but it follows your clear statement that you must "copy only these three files". You need to merge these sentences into a coherent sentence (or paragraph). For example, explain that you need to download the specific files required by your particular model's tokenizer and that, for the Mistral family model used in the example, that means these three files. You can even mention that this list is typical of many model tokenizers, but won't apply to all models.

docs/examples/custom-jsonl-dataset.md
```diff
@@ -0,0 +1,104 @@
...
+### What Each Argument Does
+
+| Argument | Purpose |
+|---|---|
+| `--backend kind=openai_http,target=...` | Points GuideLLM at your OpenAI-compatible predictor endpoint. |
+| `--data kind=json_file,path=...` | Loads prompts from the local JSONL file. |
+| `--tokenizer kind=huggingface_auto,model=/data/mistral` | Loads the tokenizer from the local directory, the 3 files from step 1, instead of downloading a model. |
+| `--data-column-mapper` | Maps the JSONL `prompt` field to the `text_column` GuideLLM expects internally. Optional here since `prompt` is already a default mapping. Shown explicitly for clarity. |
+| `--profile kind=concurrent,streams=100` | Simulates 100 concurrent "users" hitting the endpoint at once. |
+| `--data-loader kind=pytorch,samples=1000` | Reads 1000 lines or samples from the JSONL file for the run. |
```
dbutenhof: You changed the actual command to `samples=-1`, but forgot to update this. You also want to add that `-1` (load all samples in the dataset) is the default, and you're just trying to demonstrate the use of `--data-loader`.

dbutenhof (changes requested):
This is improving -- sorry, but still a few more problems.

dbutenhof:
By the way, you need to fix your file formatting:

```console
run markdown formatter...................................................Failed
- hook id: mdformat
- files were modified by this hook
```

So, when you're making further changes, `git add` the updated file to your local git index, then

```console
$ pip install pre-commit
$ pre-commit install
$ pre-commit run
```

It will give you a message like the CI's message above ... only in this case, the modified files are in your local branch. Add the changes to your git index, then you can re-run `pre-commit run` to verify that there are no remaining problems before committing the changes.

docs/examples/custom-jsonl-dataset.md
```diff
@@ -0,0 +1,83 @@
...
+
+Benchmark an already-deployed OpenAI-compatible model endpoint using GuideLLM with a local tokenizer and a custom JSONL prompt dataset, without needing a local copy of the model itself. This example runs GuideLLM via a local `pip install`.
+
+## Getting Started
+
+### 1. Prepare the Tokenizer Files
+
+The model is already deployed and running, for example on GPU, served via vLLM or OpenShift, behind an OpenAI-compatible endpoint. GuideLLM never loads or runs the model, it only talks to it over HTTP. What GuideLLM does need locally is the tokenizer, since it uses it to compute token counts for its metrics such as prompt tokens, output tokens and throughput.
+
+From the model's repo, for example the HuggingFace Hub or wherever the model artifacts live, copy the tokenizer files required by your specific model. This example uses a Mistral-family model, which needs only these three files. A pattern that's typical of many tokenizers, but not universal, so treat it as a starting point rather than a fixed requirement for every model:
```
dbutenhof, line 11: The final sentence fragment is a bit distracting. How about "This is a pattern" ... or even more specifically, "The dependency on these three files is a pattern".

dbutenhof (approved):
One somewhat awkward sentence could be clarified.

And while I'm basically happy about removing the extra steps of the containerized deployment and the unnecessary default options, I also feel just a little bad that I pushed you away from what you were trying to do. (Just a little: but... "sorry".)

SkiHatDuckie (approved):
I've got no complaints.

dbutenhof:
@mergifyio queue

dbutenhof:
@mergifyio queue

dbutenhof:
@mergifyio refresh

PR #916, opened 2026-07-08
Add a setting to configure the default working dir

sjmonson:
## Summary

The Containerfiles are still using the `GUIDELLM_OUTPUT_DIR` environment variable which no longer exists. There is still a valid use-case for having a separate default directory from the current one so this PR adds back that functionality. Individual outputs will still override the path in its entirety.

---

## Use of AI

---

# git log

commit be587bc4b2152ef5b14cf3bedf4360a1fc821060
Date:   Wed Jul 8 13:45:52 2026 -0400

    Add a setting to configure the default working dir
    
    Previously it was possible set the output directory for all outputs with
    a single environment variable. This was removed with the v0.7.0 refactor
    as now each output manages its own path. However, its still useful to
    set a default output directory for special environments such as
    containers where we may want to redirect all outputs to a volume by
    default.
    
commit dd4ed2fbd97b0ce6631ddfcca2b3f7d95a302bfb
Date:   Wed Jul 8 14:07:09 2026 -0400

    Fix WD setting in Containerfiles
    
commit 48b7cae59bb6f91c2c0569866c7c57733d7a190d
Date:   Thu Jul 9 14:48:59 2026 -0400

    Rename WD control to default_results_dir
    
---------

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -68,9 +68,9 @@ RUN --mount=type=cache,sharing=locked,target=/var/cache/dnf \
 USER 1001:0
 
 # Add guidellm bin to PATH
-# Argument defaults can be set with GUIDELLM_<ARG>
+# Change the default save directory for results
 ENV HOME="/home/guidellm" \
-    GUIDELLM_OUTPUT_DIR="/results"
+    GUIDELLM__WORKING_DIRECTORY="/results"
 
 # Create the user home dir
 WORKDIR $HOME
--- a/Containerfile
+++ b/Containerfile
@@ -70,7 +70,7 @@ USER 1001:0
 # Add guidellm bin to PATH
 # Change the default save directory for results
 ENV HOME="/home/guidellm" \
-    GUIDELLM__WORKING_DIRECTORY="/results"
+    GUIDELLM__DEFAULT_RESULTS_DIR="/results"
 
 # Create the user home dir
 WORKDIR $HOME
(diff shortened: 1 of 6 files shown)


Containerfile
```diff
@@ -68,9 +68,9 @@ RUN --mount=type=cache,sharing=locked,target=/var/cache/dnf \
 USER 1001:0
 
 # Add guidellm bin to PATH
-# Argument defaults can be set with GUIDELLM_<ARG>
+# Change the default save directory for results
 ENV HOME="/home/guidellm" \
-    GUIDELLM_OUTPUT_DIR="/results"
+    GUIDELLM__WORKING_DIRECTORY="/results"
```
dbutenhof: I think the Containerfile context suggests that "working directory" is not the best name choice -- as contrasted with the actual containerized GuideLLM "working directory" set below with `WORKDIR`.

"Output" (or maybe "results") directory was a more accurate representation of what it actually does.

jaredoconnell: Yeah, looking it the use of it, output directory is how it works and how the user cares about it.

sjmonson: I didn't want to call it `GUIDELLM__OUTPUT_DIR` because that makes it seem like its the same as the previous env which it is not.

jaredoconnell (dismissed):

Containerfile
sjmonson: Made it `GUIDELLM__DEAFAULT_RESULTS_DIR`. Longer name seems fine since its more for packagers then users.

dbutenhof (approved):

jaredoconnell (approved):

PR #935, opened 2026-07-17
Subagent Conversation Support

### LINKED ISSUE #966: DAG support for agentic conversation flows
Support building and scheduling a directed acyclic graph to allow following the branching and joining of conversations in a general agentic workflow.

jaredoconnell:
## Summary

This PR includes the initial implementation of the DAG, and basic synthetic data support to utilize that.

## Details

- Implements a DAG that uses nodes and edges. The edges specify dependency requirements for history.
- Improves the conversation script to color code it and show the node's parents, and a full history path.
- It's designed in a way to hopefully easily support trace dataset formats in the future.
- The input format from the data deserializers is not the flattened design used in existing formats. This is to match both how the DAG needs it, and how typical trace dataset formats actually nest subagents.

## Test Plan

- I would recommend starting by running a benchmark using the examples given in the markdown to create simple subagents.

## Related Issues

TBD

---

## Use of AI

---

# git log

commit 04fcfd864c6a1313d94cb35d00c1030b77c98801
Date:   Wed Jul 8 11:42:13 2026 -0400

    Initial DAG implementation
    
commit 92d996ccfc4ac0c08780d2f19c53ec2103fdaa7f
Date:   Mon Jul 13 19:49:05 2026 -0400

    Integrate DAG into full synthetic data pipeline
    
commit 383dcb25d9b22d8e5f557bf90e27b9d33cd28e76
Date:   Wed Jul 15 17:50:15 2026 -0400

    Address review comments
    
commit 134719ea5fbe9a7b68a9d3c89ead9511a5c7de2f
Date:   Thu Jul 16 18:18:05 2026 -0400

    Implemented requeue delay to DAG, and fixed formatting
    
    Assisted-by: Cursor AI Claude Opus 4.6

commit d87197cd5efdb5e0252cf6cfd5fbcca3bae1ed78
Date:   Fri Jul 17 00:30:57 2026 -0400

    Redesign input format from dataset deserializers for subagent conversations
    
commit bf0701f7e678ed771d2536d007f0b91b8c7e5d41
Date:   Fri Jul 17 01:03:47 2026 -0400

    Make code compliant with all CI checks
    
commit b5a8604bfed58b35d3d3fde7ba59d4b1dc141990
Date:   Thu Jul 23 19:23:55 2026 -0400

    Added way to make conversation extraction script less verbose
    
    Generated-by: Cursor AI Grok 4.5

commit 069de56deefbfd5056aeea7976c6df7d5aa0dce0
Date:   Thu Jul 23 20:11:02 2026 -0400

    Allow setting the merge turn for synthetic subagent branches
    
    Generated-by: Cursor AI Grok 4.5

commit ae8fc5153c9474a4c0ba190adeea4644558740bf
Date:   Thu Jul 23 20:23:53 2026 -0400

    Clarify synthetic data sampling behavior
    
    Generated-by: Cursor AI Grok 4.5

commit 92e6c5c09ce547ce2d700d050d7b1bdf903ff2df
Date:   Thu Jul 23 22:35:47 2026 -0400

    Address review comment
    
    Assisted-by: Cursor AI

commit c744a83bff14f7974b1c5fadeb5729cfe3c25fc8
Date:   Thu Jul 23 22:41:46 2026 -0400

    Address review comment. Switch root_node_ids to computed field
    
    Generated-by: Cursor AI

commit 4f2440329c8ec100940c97566541471743b6df55
Date:   Fri Jul 24 16:04:04 2026 -0400

    Remove sorting from DAG
    
    This makes the DAG function order based on the DAG node insertion order.
    
    Generated-by: Cursor AI Grok 4.5

commit eee656febf4f35f724b6567b588dfb26396f8867
Date:   Fri Jul 24 17:26:07 2026 -0400

    Address review feedback
    
commit 5c13cb8acc32656bf235df7bc5d481a3db4c7fe7
Date:   Fri Jul 24 17:59:43 2026 -0400

    Refactor history building
    
    Generated-by: Cursor AI Grok 4.5

commit 752fc3ec51e21e36b26a3a2f4ae4cf33d07172fc
Date:   Tue Jul 28 14:15:20 2026 -0400

    Fix linter errors
    
    Generated-by: Cursor AI

commit 75ec0f4928d7d2c8dd3b79786e83daf245dc2031
Date:   Tue Jul 28 18:52:53 2026 -0400

    Address review comments
    
    Generated-by: Cursor AI Grok 4.5

commit 3dbeed70902b8cb10df8f57f2fb158e3999cd69a
Date:   Wed Jul 29 15:29:58 2026 -0400

    Replace turn index with history length
    
    Generated-by: Cursor AI Grok 4.5

commit f4e8da31ff8a3a1b496d580e78d288ac685e7862
Date:   Wed Jul 29 17:42:43 2026 -0400

    Fix multiturn synthetic sampling bug, and add first prompt settings
    
    This will allow us to better mock real world scenarios.
    
    Generated-by: Cursor AI Grok 4.5

commit 43d57e7755a75aef461fd28ce3129b9b4b0b55a4
Date:   Wed Jul 29 18:00:56 2026 -0400

    Address review comments
    
commit b46f1d4ecc80e9e11e650a9cde40a200a1d61b03
Date:   Thu Jul 30 10:58:14 2026 -0400

    Add turn_index in addition to history_len
    
    Generated-by: Cursor AI Grok 4.5

commit 44c09d6e55c5e2f4512541d15740359cec887110
Date:   Thu Jul 30 11:41:41 2026 -0400

    Remove script accidentally added in prior commit
    
commit ee61429706395c1cdd77338e5584ea5e4c75753b
Date:   Thu Jul 30 15:16:11 2026 -0400

    Reduce duplication for the data pathway
    
    With the changes to accept subagent setups, it got a little duplicated.
    
    Generated-by: Cursor AI Grok 4.5

commit f0ff27dc0c006488dc1f43ff20bf97fcfd0d25e0
Date:   Thu Jul 30 16:14:29 2026 -0400

    Address review feedback and fix linter error
    
commit 5b54ffd05ffa40c4a1e6cc3dac6a6fe7b407c542
Date:   Thu Jul 30 16:51:44 2026 -0400

    Fix unit tests, and make E2E tests work with port conflict
    
    Generated-by: Cursor AI Grok 4.5

commit 2c520ec0181f11ae3c81dcc88002305aebb2016d
Date:   Fri Jul 31 23:34:26 2026 -0400

    Address review comments and remove dead code
    
    Assisted-by: Cursor AI Grok 4.5

---------

Assisted-by: Cursor AI
Assisted-by: Cursor AI Claude Opus 4.6
Assisted-by: Cursor AI Grok 4.5
Generated-by: Cursor AI
Generated-by: Cursor AI Grok 4.5


docs/guides/multiturn.md
```diff
@@ -114,6 +114,47 @@ For this configuration:
...
 
+#### Sub-Agent Branches
+
+GuideLLM supports simulating multi-agent workloads where an orchestrator spawns parallel sub-agents during a conversation. Use the `branches` parameter to specify sub-agent branches that fork from the main conversation at a specific turn and merge back at the next turn.
+
+Each branch:
+
+- **Spawns** at `at_turn` with fresh context (the sub-agent does not see the main conversation history)
+- **Runs** for `turns` turns independently
+- **Merges** back at `at_turn + 1`, where the main conversation receives the sub-agent's final output
```
dbutenhof: So no support for background subagents that run in parallel across multiple orchestrator turns?

scripts/extract_conversation.py
```diff
@@ -73,30 +236,98 @@ def extract_conversations(
         start_index = 0
 
     for ri, req in enumerate(reqs):
-        _print_request(req, turn_number=start_index + ri + 1, total=total)
+        _print_request(
```
dbutenhof, line 245: It'd be cool if there was another CLI option to skip the actual *text* of the prompt and response -- because for synthetic data it's mildly amusing the first time and then tremendously boring. But I like the header info. (Maybe tool info, too, so possibly being able to disable USER/RESPONSE/TOOL detail separately would be valuable?)

E.g.,

```console
 Turn 3 of 13    (prompt_tokens: 264 | output_tokens: 128)
    agent: twoer | node: branch_1_0 | parents: ['main_2'] | graph: fed94ac6-4797-40ab-9df1-5c5ef9bb8f56
    history: main_0 (default) -> main_1 (default) -> main_2 (default) -> branch_1_0 (twoer)
 Turn 4 of 13    (prompt_tokens: 658 | output_tokens: 128)
    agent: default | node: main_1 | parents: ['main_0'] | graph: fed94ac6-4797-40ab-9df1-5c5ef9bb8f56
    history: main_0 (default) -> main_1 (default)
```

I did a simple dump using `jq -r '.benchmarks.[].requests.successful.[].info|"agent \(.agent_id) turn \(.turn_index): rid \(.request_id), cid \(.conversation_id), nid \(.node_id)"' < benchmarks.json` which is less elegant, but usefully concise.

scripts/extract_conversation.py
```diff
@@ -20,11 +20,169 @@
...
+from collections.abc import Callable
 from pathlib import Path
 from typing import Any
 
 import yaml
 
+# ---------------------------------------------------------------------------
+# ANSI color helpers — disabled when stdout is not a terminal
+# ---------------------------------------------------------------------------
+_USE_COLOR = sys.stdout.isatty()
```
dbutenhof, line 37: Somehow my gnome terminal isn't showing color. `sys.stdout.isatty()` returns `True`, and literal color output like bash `printf "\e[1;31mThis is Bold Red Text\e[0m\n"` works fine, so it's something in here, not in the terminal settings. 😦

src/guidellm/data/deserializers/synthetic.py
```diff
@@ -64,6 +70,52 @@ class SyntheticTextPrefixBucketConfig(BaseModel):
...
+        gt=0,
+    )
+    output_tokens: int | None = Field(
+        description=(
+            "Override output token count for branch turns. "
+            "If None, uses the main chain's output_tokens."
+        ),
+        default=None,
+        gt=0,
+    )
```
dbutenhof: Probably overkill, so forget I even said it ... but no min/max/stddev? 😁

src/guidellm/data/deserializers/synthetic.py
```diff
@@ -266,6 +328,24 @@ def check_prefix_options(self) -> SyntheticTextDataArgs:
...
+        if isinstance(v, str):
+            try:
+                v = json.loads(v)
+            except (json.JSONDecodeError, ValueError) as err:
+                raise ValueError(
+                    f"branches must be a JSON list of BranchSpec objects, got {v!r}"
+                ) from err
+        if not isinstance(v, list):
+            raise ValueError(f"branches must be a list, got {type(v)}")
+        return v
```
dbutenhof: In theory, it seems like this would allow specifying the branch list as a serialized JSON string -- but I don't think our parser would actually let that through anyway as it splits on `,` and requires `<str>=<value>` in each segment. This should be pointless if you specify a full serialized JSON data spec or something like

```
--data "kind=synthetic_text,prompt_tokens=256,output_tokens=128,turns=10,branches[0].at_turn=2,branches[0].turns=3"
```

src/guidellm/scheduler/dag.py
```diff
@@ -0,0 +1,378 @@
...
+                                    (
+                                        last_completed.request,
+                                        last_completed.response,
+                                    ),
+                                )
+                            )
+
+            current_id = full_edge.source_node_id if full_edge is not None else None
+
+        chain_reversed.reverse()
```
SkiHatDuckie, line 383: So is the chain being reversed *here* or was the chain already reversed and now we're unreversing it?

src/guidellm/scheduler/dag.py
```diff
@@ -0,0 +1,378 @@
...
+
+        return result
+
+    def get_remaining_node_ids(self) -> list[str]:
+        """
+        Get all node IDs that haven't been completed yet.
+
+        :return: Sorted list of incomplete node IDs.
+        """
+        return sorted(nid for nid in self._graph.nodes if nid not in self._completed)
```
SkiHatDuckie: What is the purpose of sorting the list in alphanumeric order?

src/guidellm/scheduler/dag.py
```diff
@@ -0,0 +1,378 @@
...
+            has_any_history = True
+            full_chain = self._walk_back_full(full_edge.source_node_id)
+
+        # Collect last entries (sorted by source_node_id for determinism)
+        for edge in sorted(incoming, key=lambda e: e.source_node_id):
+            if edge.history_context == "last":
+                has_any_history = True
+                completed = self._completed.get(edge.source_node_id)
+                if completed is not None:
+                    last_entries.append(
```
SkiHatDuckie: Why are we storing the source node ID in `last_entries` if it's never used afterwards?

src/guidellm/scheduler/dag.py
```diff
@@ -0,0 +1,378 @@
...
+
+            # Only collect last parents at nodes where the walk CONTINUES
+            # (has a full parent). At the stopping node, last parents are
+            # the node's own context, not part of downstream history.
+            if full_edge is not None:
+                for edge in sorted(current_incoming, key=lambda e: e.source_node_id):
+                    if edge.history_context == "last":
+                        last_completed = self._completed.get(edge.source_node_id)
+                        if last_completed is not None:
+                            interleaved_last.append(
```
SkiHatDuckie: Same question as on line 263: What is the purpose of storing the source node ID with the (request, response) pair if it's never accessed and thrown out before returning the walk back?

src/guidellm/scheduler/dag.py
```diff
@@ -0,0 +1,378 @@
...
+        full_edge = self._find_full_parent_edge(incoming)
+        if full_edge is not None:
+            has_any_history = True
+            full_chain = self._walk_back_full(full_edge.source_node_id)
+
+        # Collect last entries (sorted by source_node_id for determinism)
+        for edge in sorted(incoming, key=lambda e: e.source_node_id):
+            if edge.history_context == "last":
+                has_any_history = True
+                completed = self._completed.get(edge.source_node_id)
```
SkiHatDuckie: So if an edge with a "last" history context is found, but has already been completed, it's *not* added to the walk back? How does that work? Is the (request, response) pair no longer available once the source node is complete?

src/guidellm/scheduler/dag.py
```diff
@@ -0,0 +1,378 @@
...
+        delay = self._graph.nodes[node_id].settings.requeue_delay or 0.0
+        unlock_at = time.time() + delay
+
+        newly_ready: list[str] = []
+        for edge in self._outgoing_edges[node_id]:
+            child_id = edge.target_node_id
+            self._remaining_parents[child_id] -= 1
+            if (
+                self._remaining_parents[child_id] == 0
+                and child_id not in self._completed
```
SkiHatDuckie, line 199: Can a child node really be completed before it's remaining parents is 0? `claim_node` prevents a child from even being claimed until all remaining parents is 0. This second condition seems impossible to be false unless I've missed something.

src/guidellm/scheduler/dag.py
```diff
@@ -0,0 +1,378 @@
...
+        :return: The conversation graph being executed.
+        """
+        return self._graph
+
+    @property
+    def is_complete(self) -> bool:
+        """
+        :return: True if all nodes have completed (or the graph was aborted).
+        """
+        return self._aborted or len(self._completed) == len(self._graph.nodes)
```
SkiHatDuckie: I mean, yeah technically the nodes are done being processed when the thing aborts, but there's already an `is_aborted` property. Would it make sense to remove the `self._aborted` check from `is_complete`?

So then, the value of `is_complete` clearly signifies a successful completion of execution, and if someone needs to check for aborts OR completion, they can call:

```python
if state.is_aborted or state.is_complete:
    ...
```

src/guidellm/scheduler/dag.py
```diff
@@ -0,0 +1,378 @@
...
+    "DAGExecutionState",
+]
+
+
+_RequestT = TypeVar("_RequestT")
+_ResponseT = TypeVar("_ResponseT")
+
+
+@dataclass(frozen=True)
+class CompletedNodeData(Generic[_RequestT, _ResponseT]):
```
SkiHatDuckie, line 36: I wonder if a dataclass with a broader purpose (i.e. `NodeData`, with fields not frozen) would make sense to have? You reference (request, response) pairings a lot in just this file alone.

SkiHatDuckie (commented):
Some comments on dag.py. Many of these are questions that will also help me better understand what the DAG is doing and why.

There will be multiple batches of comment/requests as I go through this, but one of them will be dedicated to seeing how well this interface would work with trace replays like WEKA.

dbutenhof (commented):
I still haven't actually been able to finish this, but before I head out to hike I might as well submit my initial comments from Friday and this morning. 😁

src/guidellm/scheduler/schemas/conversation.py
```diff
@@ -0,0 +1,329 @@
...
+        ),
+    )
+    request_infos: dict[str, RequestInfo] = Field(
+        default_factory=dict,
+        description=(
+            "Per-node RequestInfo, populated by the scheduler coordinator "
+            "before dispatch. Not part of the dataset -- created at runtime."
+        ),
+    )
+
```
SkiHatDuckie: Pydantic has a decorator for computed values:
```suggestion
    request_infos: dict[str, RequestInfo] = Field(
        default_factory=dict,
        description=(
            "Per-node RequestInfo, populated by the scheduler coordinator "
            "before dispatch. Not part of the dataset -- created at runtime."
        ),
    )

    @computed_field  # type: ignore[prop-decorator]
    @property
    def root_node_ids(self) -> list[str]:
        "Node IDs with no incoming edges."
        node_ids = set(self.nodes.keys())
        incoming: set[str] = {edge.target_node_id for edge in self.edges}
        return sorted(nid for nid in node_ids if nid not in incoming)
```

src/guidellm/scheduler/schemas/conversation.py
```diff
@@ -0,0 +1,329 @@
...
+        Kahn's algorithm, and derives ``root_node_ids``.
+
+        :raises ValueError: If edges reference missing nodes, a node has
+            multiple ``full`` incoming edges, or the graph contains a cycle.
+        """
+        node_ids = set(self.nodes.keys())
+        self._validate_edge_endpoints(node_ids)
+        in_degree, children = self._build_adjacency(node_ids)
+        self._check_acyclicity(node_ids, in_degree, children)
+        self._derive_root_node_ids(node_ids)
```
SkiHatDuckie: If prior suggestion committed:
```suggestion
```

src/guidellm/scheduler/schemas/conversation.py
```diff
@@ -0,0 +1,329 @@
...
+        if visited_count != len(node_ids):
+            raise ValueError(
+                "ConversationGraph contains a cycle; "
+                f"{len(node_ids) - visited_count} nodes are unreachable "
+                f"via topological sort"
+            )
+
+    def _derive_root_node_ids(self, node_ids: set[str]) -> None:
+        incoming: set[str] = {edge.target_node_id for edge in self.edges}
+        self.root_node_ids = sorted(nid for nid in node_ids if nid not in incoming)
```
SkiHatDuckie: If prior suggestion committed:
```suggestion
```

src/guidellm/scheduler/schemas/conversation.py
```diff
@@ -0,0 +1,329 @@
+from __future__ import annotations
+
+import uuid
+from collections import deque
+from collections.abc import Iterable
+from typing import Generic, Literal
+
+from pydantic import Field, model_validator
```
SkiHatDuckie: If prior suggestion committed:
```suggestion
from pydantic import Field, computed_field, model_validator
```

SkiHatDuckie (changes requested):
Treat these as one suggestion. I can't find a good way of suggesting changes across different lines without making a large diff.

src/guidellm/scheduler/dag.py
```diff
@@ -0,0 +1,334 @@
...
+        :return: Sorted list of node IDs ready for execution.
+        """
+        if self._aborted:
+            return []
+
+        ready = []
+        for nid in self._graph.nodes:
+            if nid not in self._completed and self._remaining_parents[nid] == 0:
+                ready.append(nid)
+        return sorted(ready)
```
dbutenhof: What value is there to *sorting* lists of node IDs? Do they have some canonical order that's meaningful to preserve? If this is just for "reproducability" (e.g., for testing), it seems like a big cost.

src/guidellm/scheduler/dag.py
```diff
@@ -0,0 +1,334 @@
...
+        if full_edge is not None:
+            has_any_history = True
+            full_chain = self._walk_back_full(full_edge.source_node_id)
+
+        # Collect last entries (sorted by source_node_id for determinism)
+        for edge in sorted(incoming, key=lambda e: e.source_node_id):
+            ctx = edge.history_context
+            if isinstance(ctx, HistoryContext):
+                ctx = ctx.value
+            if ctx == HistoryContext.LAST.value:
```
dbutenhof: This looks odd. `history_context` is typed to `HistoryContext`, so it shouldn't be anything else; and why compare the *value* instead of just `if edge.history_context is HistoryContext.LAST`??

If this is really due to weaknesses in the IPC, and we're not doing `HistoryContext(value)` on the receiving side, then the `history_context` signature ought to be `HistoryContext | str` to be clear. (Or, I suppose, just use the ENUM as a prototype str generator and use `HistoryContext.<name>.value` everywhere ... which is ugly but still ­-- barely -- beats inline literal strings.)

Also, you do this in a lot of places. If this is really what you want, I'd suggest something like

```python
    def matches(self, value: HistoryContext | str) -> bool:
        if isinstance(value, HistoryContext):
            value = value.value
        return self.value == value
```

Or even

```python
    def matches(self, value: HistoryContext | str) -> bool:
        return self is HistoryContext(value)
```

src/guidellm/scheduler/dag.py
dbutenhof: I believe we're following the chain *backwards* in history, appending to the list (i.e., reversed order) and then reversing here. I almost wonder if it'd be easier/quicker to use a deque and appendleft, then just `list()` it.

src/guidellm/scheduler/schemas.py
```diff
@@ -72,6 +79,241 @@
...
+        self, node_ids: set[str]
+    ) -> tuple[dict[str, int], dict[str, list[str]]]:
+        in_degree = dict.fromkeys(node_ids, 0)
+        children: dict[str, list[str]] = {nid: [] for nid in node_ids}
+        full_parent_count = dict.fromkeys(node_ids, 0)
+
+        for edge in self.edges:
+            in_degree[edge.target_node_id] += 1
+            children[edge.source_node_id].append(edge.target_node_id)
+            if edge.history_context == HistoryContext.FULL:
```
dbutenhof: Enums are canonically compared by identity (`is`) rather than equality (`==`) since they're by definition singletons.

Alternatively, you could use the standard `StrEnum` base type (`class HistoryContext(StrEnum)`) in which case you have the ENUM names but they actually *act* like strings (and are compared like strings) ... which, come to think of it, might also solve your IPC marshalling issue.

dbutenhof (commented):
Some additional comments / questions on the first (DAG) commit. (Whew.)

src/guidellm/scheduler/worker_group.py
```diff
@@ -565,47 +564,56 @@ def requests_generator(
...
 
-            def _turn_iter(
-                requests_chain: Iterable[tuple[RequestT, RequestSettings]],
-            ) -> Generator[RequestDataT[RequestT], None, None]:
-                nonlocal count, stop_queueing
-                # NOTE: This allows users to correlate requests in post-processing
-                conv_id = str(uuid.uuid4())
-                for i, (request, setting) in enumerate(requests_chain):
+            for graph in requests:
+                from guidellm.scheduler.dag import DAGExecutionState
```
dbutenhof: Do we really want to do local imports? I don't see any obvious need to do this here -- it's not an optional extra.

src/guidellm/scheduler/schemas/backend.py
dbutenhof, line 1: Love that you broke this out into a separate file ... hate that makes it so hard to review in GitHub. 😦

dbutenhof (commented):
OK, I started going through by commit, thinking it'd be easier, and then found out you removed the ENUMs and reorganized everything, so that exercise proved pointless. AH, well: I'll resume/restart tomorrow.

src/guidellm/scheduler/worker_group.py
jaredoconnell: We definitely don't. Fixed in follow up.

src/guidellm/scheduler/schemas.py
jaredoconnell: Good point. I'll keep that in mind if we ever decide to switch back to enums.

src/guidellm/scheduler/schemas/backend.py
jaredoconnell: Yeah. Sam suggested I split it off. It was done in the second commit, I believe. Most of it is unchanged.

src/guidellm/scheduler/dag.py
jaredoconnell: I don't love Python's implementation of enums. For now it's switched back to Literals.

scripts/extract_conversation.py
jaredoconnell: Good idea. I added a CLI option to display just the metadata, just the roles, or everything it can show.

docs/guides/multiturn.md
jaredoconnell: Sure. That's easy to add. Done.

scripts/extract_conversation.py
jaredoconnell: Based on our conversation in Slack, this may be an environment issue because it's working now.

src/guidellm/data/deserializers/synthetic.py
jaredoconnell: Yeah, this is likely overkill. That said, looking into this made me realize that the size of the prompts is sampled per conversation, rather than once per message. I wonder if we want to keep that behavior.

src/guidellm/data/deserializers/synthetic.py
jaredoconnell: I removed this function.

src/guidellm/scheduler/schemas/conversation.py
jaredoconnell: Switched to a computed field locally.

src/guidellm/scheduler/dag.py
jaredoconnell: I definitely appreciate more generic pathways. I'm not sure if it would help enough to offset the other complicating factors.

src/guidellm/scheduler/dag.py
jaredoconnell: Sure. This design is a good separation of concerns.

src/guidellm/scheduler/dag.py
jaredoconnell: I removed the sorting. The order is now based on the order of the insertion into the DAG.

src/guidellm/scheduler/dag.py
jaredoconnell: Assuming we don't introduce a bug, it should be impossible. This check is defensive in case the state gets corrupt.

src/guidellm/scheduler/dag.py
jaredoconnell: In this case it's looping through every "last" entry, but it's not doing so recursively because of the "last" behavior. "last" nodes are basically just ones you get the history of non-recursively because you only get the last node's history.
But it can only add completed ones, as the other ones do not have the necessary information. But because we should never get to non-completed nodes if the DAG is used correctly, I switched it to raise an exception if the completed data is not available.

src/guidellm/scheduler/dag.py
jaredoconnell: I'm not sure. It's removed, and I refactored this code a bit.

src/guidellm/scheduler/dag.py
jaredoconnell: It's removed now. I refactored the code.

src/guidellm/scheduler/dag.py
jaredoconnell: The return type is a list, so it's likely cheaper to reverse an array-backed data type than to use one that's efficient for prepending and later convert it to a list.

src/guidellm/scheduler/dag.py
jaredoconnell: I removed the sorting. The ordering is now based on DAG node insertion order.

src/guidellm/scheduler/worker.py
```diff
@@ -349,147 +325,231 @@ def _task_done(task: asyncio.Task[ProcessRequestT[RequestT, ResponseT]]):
...
+                if ready:
+                    node_id = ready[0]
+                    state.claim_node(node_id)
+                    return state, node_id
+
+            earliest: float | None = None
+            for state in self.turns_queue:
+                next_at = state.next_delayed_ready_at()
+                if next_at is not None and (earliest is None or next_at < earliest):
+                    earliest = next_at
```
sjmonson: `get_ready_nodes` and `next_delayed_ready_at` are super similar functions especially considering you discard all of the ready node ids except the first one. Why not just have a single `state.next_node_ready_at() -> tuple[str, float] | None` to return the next node id and ready time and return -inf if the node is ready now?

src/guidellm/scheduler/worker.py
```diff
@@ -349,147 +325,231 @@ def _task_done(task: asyncio.Task[ProcessRequestT[RequestT, ResponseT]]):
...
+                except asyncio.TimeoutError:
+                    continue
+            else:
+                graph = await self.messaging.get()
 
-        Retrieves request from messaging queue, applies timing strategy, processes
-        through backend, and publishes status updates throughout the lifecycle.
+            state = DAGExecutionState(graph)
+            self.turns_queue.append(state)
+            self.graph_request_infos[graph.graph_id] = dict(graph.request_infos)
```
sjmonson: Not quite sure why you need a separate dictionary for `request_info`s. Since `DAGExecutionState` holds a reference to the graph why not just add a get property so that you can `state.request_infos(node_id)`? Or `state.request_infos[node_id]` if you want to be fancy.

src/guidellm/scheduler/worker.py
```diff
@@ -349,147 +325,231 @@ def _task_done(task: asyncio.Task[ProcessRequestT[RequestT, ResponseT]]):
...
+        )
+        if effective_target_start != target_start:
+            request_info.timings.targeted_start = effective_target_start
+            self._send_update("pending", None, request, request_info)
+        await self._schedule_request(request, request_info, effective_target_start)
+
+        history_pairs = state.assemble_history(node_id)
+        history: HistoryT[RequestT, ResponseT] | None = (
+            history_pairs if history_pairs else None
+        )
```
sjmonson:
```suggestion
        history: HistoryT[RequestT, ResponseT] | None = (
            state.assemble_history(node_id) or None
        )
```

? Or do `history or None` in the backend call below. Could also change the backend interface to remove `None` and avoid having to do this conversion.

src/guidellm/scheduler/worker.py
```diff
@@ -349,147 +325,231 @@ def _task_done(task: asyncio.Task[ProcessRequestT[RequestT, ResponseT]]):
...
-
-            effective_target_start = await self.strategy.resolve_dequeued_target_start(
-                self.worker_index,
-                target_start,
-                request_info.settings,
+            state, node_id = await self._get_next_ready_node()
+            request, request_info = self._prepare_node(state, node_id, target_start)
+            response = await self._execute_node(
+                state, node_id, request, request_info, target_start
             )
```
sjmonson: Because you split this up `_execute_node` needs to yield responses not return them. As is right now, if the request is cancelled partway you will not receive the final response from the backend here since `_execute_node` never returns.

sjmonson (changes requested):
Some initial notes on the worker code. Overall looks good, it was pretty easy to follow the code. Nits are a few inefficiencies and one bug.

src/guidellm/scheduler/worker_group.py
```diff
@@ -565,47 +565,57 @@ def requests_generator(
...
-                    request_id = self._find_request_id(request)
+                    request_id = self._find_request_id(node.request)
                     request_info: RequestInfo = RequestInfo(
                         request_id=request_id,
-                        conversation_id=conv_id,
+                        conversation_id=graph.graph_id,
+                        node_id=node_id,
+                        agent_id=node.agent_id,
+                        parent_node_ids=incoming_map[node_id],
                         turn_index=i,
```
sjmonson: `turn_index` should be the number of requests since initial, so branches increment turn index separately. Maybe it would be better to calculate this later from history length instead of here though.

src/guidellm/scheduler/worker.py
jaredoconnell: Switched to the function you suggested.

src/guidellm/scheduler/worker.py
jaredoconnell: I removed the unnecessary processing, and all that's left is the call to `assemble_history`. Everything else was redundant.

src/guidellm/scheduler/worker.py
jaredoconnell: The proposed fix is now pushed.

src/guidellm/scheduler/worker.py
jaredoconnell: Good point. Changed to get the request the request infos from the DAGExecutionState.

src/guidellm/scheduler/worker.py
```diff
@@ -349,147 +322,229 @@ def _task_done(task: asyncio.Task[ProcessRequestT[RequestT, ResponseT]]):
...
+                if nxt is not None and nxt[1] <= now:
+                    node_id, _ = nxt
+                    state.claim_node(node_id)
+                    return state, node_id
+
+            earliest: float | None = None
+            for state in self.turns_queue:
+                nxt = state.next_node_ready_at()
+                if nxt is not None and (earliest is None or nxt[1] < earliest):
+                    earliest = nxt[1]
```
sjmonson: Maybe I am missing something but why do these need to be separate loops?

```suggestion
            now = time.time()
            earliest: float | None = None
            for state in self.turns_queue:
                nxt = state.next_node_ready_at()
                if nxt is not None:
                    if nxt[1] <= now:
                        node_id, _ = nxt
                        state.claim_node(node_id)
                        return state, node_id
                    if earliest is None or nxt[1] < earliest:
                        earliest = nxt[1]
```

src/guidellm/data/finalizers/generative.py
```diff
@@ -78,7 +87,65 @@ def __call__(
...
+        :return: A generative conversation graph.
+        :raises ValueError: If the payload is missing or invalid.
+        """
+        raw_values = item.get("conversation_turns_column") or []
+        if not raw_values:
+            raise ValueError("conversation_turns_column is empty")
+
+        raw = raw_values[0]
+        if isinstance(raw, str):
+            graph_data = ConversationGraphData.model_validate(json.loads(raw))
```
sjmonson:
```suggestion
            graph_data = ConversationGraphData.model_validate_json(raw)
```

sjmonson: Actually does this condition even make sense? Why would this ever be a string?

src/guidellm/data/finalizers/generative.py
```diff
@@ -78,7 +87,65 @@ def __call__(
...
+            request, settings = self.finalize_turn(dict(turn.columns))
+            # Prefer explicit turn-level scheduling fields over column values.
+            if turn.relative_timestamp is not None:
+                settings = settings.model_copy(
+                    update={"relative_timestamp": turn.relative_timestamp}
+                )
+            if turn.requeue_delay is not None:
+                settings = settings.model_copy(
+                    update={"requeue_delay": turn.requeue_delay}
+                )
```
sjmonson: I am very confused by what is going on here. Why does `ConversationTurnData` have a duplicate of every request setting field? And why is this code coping the model instead of just assigning the value?

src/guidellm/data/finalizers/generative.py
sjmonson, line 1: I guess maybe the whole finalizer is a hack until #972 but I don't like how we have different code paths for column-wise vs single column datasets. Ideally at this point we are either all one or the other.

src/guidellm/data/deserializers/synthetic.py
```diff
@@ -98,6 +170,8 @@ class SyntheticTextDataArgs(DataArgs):
     output_tokens: int | None = Field(
         description=(
             "The average number of text tokens generated for outputs. "
+            "For multiturn, one value is sampled per conversation and applied "
+            "to every turn (including branch turns that do not override it). "
```
sjmonson: This is a bug, it should be fixed not codified.

sjmonson: Maybe fixed in a separate PR but please remove this and any other reference to "one value is sampled per conversation" that is not how this should work.

sjmonson (changes requested):
Few more comments

src/guidellm/data/deserializers/synthetic.py
```diff
@@ -98,18 +104,28 @@ class BranchSpec(StandardBaseModel):
         description="Agent identity for branch nodes.",
         default="worker",
     )
+    merge_after: int = Field(
+        description=(
+            "How many main-chain turns after at_turn the branch merges back. "
```
dbutenhof: "main-chain"? You also used this (unhyphenated) for `at_turn`, but I think the "Main conversation" notation used in the docstring is more clear.

(Although "main conversation" still sounds a bit awkward when we're talking about a graph of conversations, I don't really have a better concise label ... 🤔)

dbutenhof (commented):
Quick skim through changes (all I've got brain left for right now); one minor new comment, and I believe I've resolved all my previous comments.

src/guidellm/data/deserializers/synthetic.py
jaredoconnell: I ended up fixing it, and while I was at it I switched to a way to separately specify the distribution for the first prompts, since in real world scenarios, you often include a big payload in the first message.

src/guidellm/data/finalizers/generative.py
jaredoconnell: This is a hard problem to solve without redoing everything. Because the existing pathways always assume it's flattened, but flattened data is very inefficient for multiturn conversations with subagents. But switching away from the existing flattened format as an option would break a lot of the advantages that the current column mapper provide.

So my thought is the dual code path format until we refactor the entire column mapper and preprocessor code sections.

src/guidellm/data/deserializers/synthetic.py
jaredoconnell: I switched it to `main (parent) conversation`.

src/guidellm/data/finalizers/generative.py
jaredoconnell: According to the AI, yes it can be a string.
> Yes, it is a string whenever the column comes from the synthetic branched dataset (or any HF/large_string JSON column). The condition exists because that payload is serialized for storage, not because the finalizer invents a string type.

I applied your suggestion locally.

src/guidellm/data/finalizers/generative.py
jaredoconnell: I'll look into this more tomorrow.

src/guidellm/scheduler/worker_group.py
jaredoconnell: I replaced `turn_index` with `history_len`, if that's okay. I think that's a better metric for the new meaning, and turn_index would be misleading in nonlinear situations.

src/guidellm/schemas/info.py
```diff
@@ -169,9 +169,14 @@ class RequestInfo(StandardBaseModel):
             "Identifier for the conversation this request is part of, if applicable."
         ),
     )
-    turn_index: int = Field(
+    history_len: int = Field(
```
sjmonson, line 172: You can have a history_len field too if you want but turn_index should still exist and start from the first turn regardless of history,

src/guidellm/data/finalizers/generative.py
jaredoconnell: I simplified this to reduce duplication.

src/guidellm/data/finalizers/generative.py
jaredoconnell: So the latest design for the new format is that there's one new column: "conversation_turns_column". It stores JSON data that follows the schema defined in `src/guidellm/data/schemas/conversation_graph_data.py`.

SkiHatDuckie:
Could you give me a rundown of which parts of the DAG the trace replay datasets would be interfacing with? I'm also looking through the code myself so that I can start writing up a plan for the WEKA trace sub-agents while this is still in review.

src/guidellm/scheduler/worker.py
jaredoconnell: Merged.

src/guidellm/schemas/info.py
jaredoconnell: Let me know what you think of the latest version of turn_index. Rather than the total history length up to that point, it's instead just the longest path to a parentless history node or new context history node. `last` can only count up to 1, but in most cases won't count, since in real world use cases the the "full" path is 1 or larger.

jaredoconnell:
> Could you give me a rundown of which parts of the DAG the trace replay datasets would be interfacing with? I'm also looking through the code myself so that I can start writing up a plan for the WEKA trace sub-agents while this is still in review.

Take a look at `src/guidellm/data/schemas/conversation_graph_data.py`

That file is where the schema is defined for the new `conversation_turns_column`. It's just a new column that accepts the JSON serialized version of the types in conversation_graph_data.py, and therefore allows nested data. That one is recursive, as you can see that `ConversationTurnData` also has a columns field.

src/guidellm/data/finalizers/conversation_graph.py
```diff
@@ -0,0 +1,229 @@
...
+    Scheduling columns are lifted onto ``turn.settings`` so ``columns`` stay
+    request-content only.
+
+    :param items: Mapper output — either one graph payload item or one dict of
+        columns per logical turn.
+    :return: A conversation graph data object (possibly with an empty turns list).
+    :raises ValueError: If ``conversation_turns_column`` is present but empty.
+    """
+    for item in items:
+        raw_values = item.get("conversation_turns_column") or []
```
SkiHatDuckie: The default is unnecessary; both `None` and `[]` are going to be caught by `if not raw_values` on the following line.
```suggestion
        raw_values = item.get("conversation_turns_column")
```

src/guidellm/data/finalizers/conversation_graph.py
```diff
@@ -0,0 +1,229 @@
...
+) -> bool:
+    """Return True when this logical turn should become tool-call + injection."""
+    injection_id = f"{turn.node_id}_injection"
+    if injection_id in existing_ids:
+        return False
+
+    turn_type_values = turn.columns.get("turn_type_column", [])
+    explicit_type = (
+        turn_type_values[0] if turn_type_values and turn_type_values[0] else None
+    )
```
SkiHatDuckie, line 127: The `and turn_types_values[0]` doesn't seem to be necessary given that whatever `False`-equivalent value it may be won't change the outcome of the rest of the function.
```suggestion
    explicit_type = turn_type_values[0] if turn_type_values else None
```

src/guidellm/scheduler/schemas/conversation.py
```diff
@@ -0,0 +1,329 @@
...
+        """
+        Node IDs with no incoming edges, derived from ``nodes`` and ``edges``.
+
+        :return: Root node IDs in ``nodes`` insertion order.
+        """
+        incoming = {edge.target_node_id for edge in self.edges}
+        return [nid for nid in self.nodes if nid not in incoming]
+
+    @classmethod
+    def from_linear_chain(
```
SkiHatDuckie: The test coverage for this function is a bit spotty (nothing that raises the ValueError, nor enters the second for loop)

SkiHatDuckie (approved):
Some minor things I found. Otherwise, I've completed my review.

src/guidellm/data/finalizers/conversation_graph.py
jaredoconnell: Good point. Implemented locally.

src/guidellm/data/finalizers/conversation_graph.py
jaredoconnell: You're not wrong. I'm not sure if this micro-optimization is worth it, given that in the future I may need to differentiate between "" and `False`.

src/guidellm/scheduler/schemas/conversation.py
jaredoconnell: I can actually remove this. It's only used in test cases.

dbutenhof (commented):
Looks good. I think that Sam's open conversations will prevent auto-merge, but somehow I feel reluctant to add a second approval "prematurely". 😁

sjmonson (approved):
What is the worse that can happen 😁

dbutenhof:
> What is the worse that can happen 😁

💣 🔥 💥  ???

PR #936, opened 2026-07-17
Handle repeated `--output` specifications consistently across run and from-file

### LINKED ISSUE #933: Inconsistent and limited handling of `--output` lists
### Problem Statement

With the registry type CLI refactoring in 0.7.x, we allow repeating `--output kind=<name>,...`. While the specific intent and benefit is to allow specifying multiple output types like `--output kind=json --output kind=csv`, there's nothing to prevent specifying two "json" outputs, for example, possibly with different file names. Like `--output kind=json,path=json1.json --output kind=json,path=json2.json`. While it's not clear there's much real value in doing this, there's an interesting inconsistency between `guidellm run` and `guidellm benchmark from-file`...

In the former case, we resolve the outputs in `resolve_output_formats`, which *resolves* each output specification but then stores them in a dict by format key: so that we will *resolve* two separate JSON Output instances but the second will overwrite the first in the dict and we'll never actually generate `json1.json`.

In the latter case, we *resolve* the output specifications into another list, but then store the results of `output.finalize(report)` into a dict by output type: which means we'll *generate* both `json1.json` and `json2.json`, but the console output will only confirm creation of `json2.json`.

This becomes potentially more interesting with the new "plot" output type, where the output specification's path specifies the image format and the specification supports an image resolution. So `--output kind=plot,path=plot.pdf,dpi=100 --output kind=plot,path=plot.png,dep=300` would be expected to create a 100dpi PDF plot and a 300dpi PNG plot. `from-file` will actually do this, but will only report the second; while `run` will only generate the second and silently ignore the first.

### Proposed Solution

We should resolve, generate, and report from a list in both cases. This seems most reasonable/obvious, and with the addition of the plot output also provides potential real value to end users.

### Alternatives Considered

The alternative would be to diagnose and fail on repeated output specifications of the same kind; rather than silently ignore one for `run` and generating but failing to report one for `from-file`.

Making this work "as likely expected" is not much more complicated, and seems more reasonable.

### Usage Examples

```markdown
$ guidellm run --backend kind=openai_http,target=http://localhost:8000 --output kind=plot,dpi=72,path=plot.png --output kind=plot,dpi=300,path=plot.pdf
```

### Additional Context

_No response_

Pragadeesh122:
## Summary

`--output` may be specified multiple times, but two specs of the *same kind* were handled inconsistently and lossily. Both entrypoints collected outputs/results in a `dict` keyed by output `kind`, which cannot hold two outputs of the same kind:

- `guidellm run` overwrote at resolve time, so a repeated kind (e.g. `--output kind=json,path=a.json --output kind=json,path=b.json`) was **never generated** — only the last survived.
- `guidellm benchmark from-file` generated both files but **reported only the last**.

This is most visible with the new `plot` output from the issue, e.g. `--output kind=plot,dpi=72,path=plot.png --output kind=plot,dpi=300,path=plot.pdf`.

This change resolves, generates, and reports outputs from an ordered list in both paths, so every `--output` is generated and reported, consistently.

## Details

- `resolve_output_formats` now returns an ordered `list[GenerativeBenchmarkerOutput]` (one entry per spec, repeated kinds preserved) instead of a `dict[str, GenerativeBenchmarkerOutput]` keyed by kind.
- `benchmark_generative_text` (run) and `reimport_benchmarks_report` (from-file) finalize and report over that list, collecting an ordered `list[tuple[str, Any]]` of `(kind, result)`.
- Both entrypoints' second return value changes from `dict[str, Any]` to `list[tuple[str, Any]]`. This is a public-API shape change, but unavoidable: a dict keyed by kind cannot represent duplicate kinds. The only in-repo consumers of the return value are tests (updated here); the two CLI callers discard it, and user-facing reporting happens inside the functions.
- Scope kept tight: only `src/guidellm/benchmark/entrypoints.py` and its tests; no new public types.

## Test Plan

- New `tests/unit/benchmark/test_entrypoints.py`: `resolve_output_formats` preserves duplicate kinds as an ordered list.
- Updated `tests/unit/entrypoints/test_benchmark_from_file_entrypoint.py` to the list-shaped results and added a regression test: two same-kind outputs with different paths are both generated **and** both reported, in order.
- `tox -e lint-check` and `tox -e type-check` pass.
- `tox -e test-unit` passes for these changes (the only failure observed locally was an unrelated, pre-existing `MaxDurationConstraint` timing test that passes in isolation).
- Manual CLI smoke: `guidellm benchmark from-file benchmarks.json --output kind=json,path=a.json --output kind=json,path=b.json` → both `a.json` and `b.json` are created and both are reported in the console summary.

## Related Issues

- Fixes #933

---

# git log

commit bfffc67b13f7a8be30753a6450edf6e7ecb1cf7a
Date:   Fri Jul 17 11:21:15 2026 -0500

    fix: handle repeated --output specs consistently (#933)
    
    Both `guidellm run` and `benchmark from-file` collected outputs in a dict
    keyed by output kind, so repeating a kind (two json specs, two plot specs,
    etc.) overwrote earlier entries: run never generated the duplicate, and
    from-file generated both files but reported only the last.
    
    Resolve, generate, and report outputs from an ordered list in both paths
    so every --output is generated and reported.
    
---------

### DIFF
--- a/src/guidellm/benchmark/entrypoints.py
+++ b/src/guidellm/benchmark/entrypoints.py
@@ -291,21 +291,22 @@ async def resolve_profile(
 async def resolve_output_formats(
     outputs: list[BenchmarkOutputArgs],
     console: Console | None = None,
-) -> dict[str, GenerativeBenchmarkerOutput]:
+) -> list[GenerativeBenchmarkerOutput]:
     """
     Resolve output format specifications into configured output handler instances.
 
     :param outputs: List of BenchmarkOutputArgs specifying output kind and path
     :param console: Console instance for progress reporting, or None
-    :return: Dictionary mapping format names to configured output handler instances
+    :return: Ordered list of configured output handler instances, one per input
+        specification (repeated kinds are preserved as separate entries)
     """
     console_step = (
         console.print_update_step(title="Resolving output formats") if console else None
     )
 
-    resolved: dict[str, GenerativeBenchmarkerOutput] = {}
+    resolved: list[GenerativeBenchmarkerOutput] = []
     for output_arg in outputs:
-        resolved[output_arg.kind] = GenerativeBenchmarkerOutput.resolve(output_arg)
+        resolved.append(GenerativeBenchmarkerOutput.resolve(output_arg))
 
     if console_step:
         console_step.finish(
@@ -458,7 +459,7 @@ async def benchmark_generative_text(
     progress: GenerativeConsoleBenchmarkerProgress | None = None,
     console: Console | None = None,
     **constraints: str | ConstraintInitializer | Any,
-) -> tuple[GenerativeBenchmarksReport, dict[str, Any]]:
+) -> tuple[GenerativeBenchmarksReport, list[tuple[str, Any]]]:
     """
     Execute a comprehensive generative text benchmarking workflow.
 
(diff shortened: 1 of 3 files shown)


dbutenhof (approved):

jaredoconnell (approved):
This makes sense. I tested it and I can confirm that it works as intended.

PR #938, opened 2026-07-19
Fix misleading CLI help text for benchmark command group

### LINKED ISSUE #937: CLI: 'benchmark' command help text is misleading
## Description

The `guidellm benchmark --help` output is misleading:

```
$ guidellm benchmark --help
Usage: guidellm benchmark [OPTIONS] COMMAND [ARGS]...

  Load a previously saved benchmark report.

Options:
  --help  Show this message and exit.
from-file  Load a saved benchmark report and optionally re-export to...
```

Two issues:

1. **The help text says "Load a previously saved benchmark report"** — this describes the `from-file` subcommand, not the benchmark group itself. Users looking to run a benchmark will naturally type `guidellm benchmark` and find no way to actually run one.

2. **The docstring on the function is correct but unused** — `benchmark/__init__.py` has `help="Load a previously saved benchmark report."` passed to `@click.group()`, but the function docstring says `"Benchmark commands for performance testing generative models."` The `help=` parameter overrides the docstring in Click.

### Suggested Fix

In `src/guidellm/cli/benchmark/__init__.py`:

```python
@click.group(
    help="Load a previously saved benchmark report.",  # <- should describe the group, not from_file
    cls=DefaultGroupHandler,
    default="run",
)
```

Should be something like:

```python
@click.group(
    help="Benchmark commands for performance testing generative models.",
    cls=DefaultGroupHandler,
    default="run",
)
```

Or better yet, point users to `guidellm run` since that's the actual command for running benchmarks.

### Environment

- guidellm installed via pip (latest as of 2026-07-18)
- Fedora 44

bennyturns:
## Summary

- Fix `benchmark` group help text: was describing the `from-file` subcommand ("Load a previously saved benchmark report") instead of the group itself. Now correctly describes the group and explains the `run` command fallback behavior.
- Fix grammar in `--label` help: "Define a labels" → "Define a label"
- Fix missing space before `[repeatable]` tags in `--label` and `--override` help text
- Add missing help text for `from-file --output` option

### Before
```
$ guidellm benchmark --help
  Load a previously saved benchmark report.
```

### After
```
$ guidellm benchmark --help
  Benchmark commands for performance testing generative models.
  Unrecognized options are forwarded to the 'run' command, so
  'guidellm benchmark <run-options>' works as a shorthand for
  'guidellm run <run-options>'.
```

Fixes #937

## Test plan
- [ ] Run `guidellm benchmark --help` and verify updated description
- [ ] Run `guidellm run --help` and verify `--label` and `--override` help text
- [ ] Run `guidellm benchmark from-file --help` and verify `--output` has help text

---

# git log

commit 63b181b053a57abfa8adc6ca1d82e5a32c8a9516
Author: bennyturns <bturner@redhat.com>
Date:   Sun Jul 19 14:02:57 2026 -0400

    Fix misleading CLI help text for benchmark command group
    
    - Fix benchmark group help: was describing the from-file subcommand
      instead of the group itself
    - Fix grammar in --label help: "Define a labels" -> "Define a label"
    - Fix missing space before [repeatable] tags in --label and --override
    - Add missing help text for from-file --output option
    
    Fixes #937
    
    Signed-off-by: bennyturns <bturner@redhat.com>

---------

Signed-off-by: bennyturns <bturner@redhat.com>

### DIFF
--- a/src/guidellm/cli/benchmark/__init__.py
+++ b/src/guidellm/cli/benchmark/__init__.py
@@ -12,7 +12,10 @@ __all__ = ["benchmark"]
 
 
 @click.group(
-    help="Load a previously saved benchmark report.",
+    help=(
+        "Load and display previously saved benchmark reports. "
+        "Use 'guidellm benchmark from-file <path>' to re-export saved results."
+    ),
     cls=DefaultGroupHandler,
     default="run",
 )
--- a/src/guidellm/cli/benchmark/from_file.py
+++ b/src/guidellm/cli/benchmark/from_file.py
@@ -32,6 +32,7 @@ __all__ = ["from_file"]
     registry=BenchmarkOutputArgs,
     multiple=True,
     default=[{"kind": "console"}, {"kind": "json"}, {"kind": "html"}, {"kind": "csv"}],
+    help="Output formats for the report (default: console, json, html, csv).",
 )
 def from_file(path, outputs):
     asyncio.run(reimport_benchmarks_report(path, outputs))
--- a/src/guidellm/cli/run.py
+++ b/src/guidellm/cli/run.py
@@ -62,9 +62,9 @@ __all__ = [
     multiple=True,
     callback=cli_tools.parse_kv_str,
     help=(
-        "Define a labels in key-value pair for the run. "
-        "Example: `--label timestamp=1999-09-12@12:00:00 --label env=staging`"
-        "  [repeatable]"
+        "Define a label as a key-value pair for the run. "
+        "Example: `--label timestamp=1999-09-12@12:00:00 --label env=staging` "
+        " [repeatable]"
     ),
 )
 @registry_options_from_model(model=BenchmarkArgs, group_key="spec")
@@ -77,8 +77,8 @@ __all__ = [
     help=(
         "Define overrides for each sub-benchmark. "
         "Currently this only supports `profile.streams` or `profile.rate`. "
-        "Example: `--profile kind=concurrent --override 'profile.streams' 1,2,4,8,16`"
-        "  [repeatable]"
+        "Example: `--profile kind=concurrent --override 'profile.streams' 1,2,4,8,16` "
+        " [repeatable]"
     ),
 )
 @click.option(


src/guidellm/cli/benchmark/__init__.py
```diff
@@ -12,7 +12,12 @@
 
 
 @click.group(
-    help="Load a previously saved benchmark report.",
+    help=(
+        "Load and display previously saved benchmark reports. "
+        "Use 'guidellm benchmark from-file <path>' to re-export saved results. "
+        "Note: unrecognized options are forwarded to the 'run' command "
+        "for backward compatibility."
```
dbutenhof: I believe this is incorrect. Prior to v0.7.0, the `run` command was a subcommand of `guidellm benchmark`, and the use of `DefaultGroupHandler` with `default="run"` allowed `guidellm benchmark` to work as `guidellm benchmark run`. However, in 0.7.0, `run` is no longer a subcommand of `benchmark` and `guidellm benchmark` just returns the group help message and exits.

I think we should have removed the `DefaultGroupHandler` and `default` for 0.7.0: although it doesn't appear to be causing any problem, it also doesn't do anything useful, and we shouldn't document it. (Since `from-file` is the only `benchmark` subcommand, it might make sense to define *that* as the default here.)

For this PR, I think you should just remove the "Note", which is misleading.

dbutenhof (changes requested):
Some nice cleanups -- thanks. One problem -- you've documented something that used to work but no longer can. We should do something about that; but for now let's just not add the misleading documentation.

You also need to *sign* your commits before our CI will accept them. (Look at the failing DCO check.)

src/guidellm/cli/benchmark/__init__.py
bennyturns: OK I will! I am happy to help clean anything else up if you want to point me at it.

bennyturns: OK I removed the note line, thanks for helping me get it right.

dbutenhof (approved):

SkiHatDuckie (approved):

dbutenhof:
@mergifyio queue

PR #947, opened 2026-07-22
bug(openai): support structured chat content metadata for OpenAI backend endpoint

### LINKED ISSUE #940: Support structured chat content for google/translategemma-12b-it
## Problem

Models like `google/translategemma-12b-it` require structured content objects with additional metadata fields in `/v1/chat/completions` requests. TranslateGemma specifically expects `source_lang_code` and `target_lang_code` in each content item:

```json
{
  "role": "user",
  "content": [{
    "type": "text",
    "text": "The quick brown fox jumps over the lazy dog.",
    "source_lang_code": "en",
    "target_lang_code": "es"
  }]
}
```

GuideLLM's chat handler currently only constructs:

```json
{
  "type": "text",
  "text": "..."
}
```

This causes requests to `/v1/chat/completions` to fail with errors like `'dict object' has no attribute ...` (HTTP 400) because the model's chat template expects the language metadata fields.

## Current Workaround

Using `/v1/completions` endpoint works for raw throughput benchmarking but does not exercise the model's intended translation workflow.

## Expected Behavior

GuideLLM should support passing additional metadata fields in structured chat content objects, either via:

1. A custom dataset where each prompt can include the required structured object, with GuideLLM preserving extra fields like `source_lang_code` and `target_lang_code` when constructing the chat request.
2. A CLI option or config to specify extra fields to include in content objects (e.g. `--content-fields '{"source_lang_code": "en", "target_lang_code": "es"}'`).

## Reproduction

```bash
guidellm benchmark \
    --target 'http://localhost:8000' \
    --model google/translategemma-12b-it \
    --data='{"prompt_tokens":1000,"output_tokens":1000}' \
    --rate-type concurrent \
    --rate 1 \
    --max-seconds 60 \
    --request-format /v1/chat/completions
```

This fails because the generated chat messages are missing the required language metadata fields.

Prasannajaga:
## Summary

GuideLLM right now doesn't support structured chat content objects for models like `google/translategemma-12b-it` that require additional metadata fields (such as `source_lang_code` and `target_lang_code`) in `/v1/chat/completions` requests. Without these fields, Jinja chat templates on inference servers fail with HTTP 400 errors (`'dict object' has no attribute ...`).

This PR details the design and implementation for supporting extra metadata fields directly in chat content objects as mentioned in issue #940. It provides:
1. **CLI / Backend Configuration**: A generic `--append-payloads` option (and backend argument `content_extra_fields`) to specify extra fields to inject into chat completion requests for any model or endpoint.
2. **Custom Dataset Support**: Preservation of dictionary prompt objects from custom dataset rows in `text_column` so dynamic, per-prompt metadata can be passed.

---

### Applied Changes

* #### [MODIFY] [http.py](file:///data/inference/guidellm/src/guidellm/backends/openai/http.py)
  * Added `content_extra_fields: dict[str, Any] | None` to `OpenAIHTTPBackendArgs`.
  * Forwarded `content_extra_fields` from backend configuration to `request_handler.format(...)` in `_prepare_resolve_request`.

* #### [MODIFY] [request_handlers.py](file:///data/inference/guidellm/src/guidellm/backends/openai/request_handlers.py)
  * Updated `ChatCompletionsRequestHandler.format()` and `_format_prompts()` to inject `content_extra_fields` into text content objects.
  * Preserved custom dictionary fields when dataset rows contain dictionary items in `text_column`.

* #### [MODIFY] [run.py](file:///data/inference/guidellm/src/guidellm/cli/run.py)
  * Added `--append-payloads` CLI option to append key-value fields to structured content objects in chat completion requests.

* #### [MODIFY] [generative.py](file:///data/inference/guidellm/src/guidellm/data/finalizers/generative.py)
  * Safely extracted string content from dictionary prompt objects in `GenerativeRequestFinalizer.finalize_turn()` to ensure word and character metric calculation works for dictionary inputs.

* #### [NEW] Unit Tests
  * `tests/unit/backends/openai/test_request_handlers.py`: Added tests verifying `content_extra_fields` injection and dictionary item preservation in chat completions format.
  * `tests/unit/backends/openai/test_http.py`: Added tests verifying backend argument handling and request resolution.
  * `tests/unit/data/test_finalizers.py`: Added tests for `GenerativeRequestFinalizer` with dictionary prompt items.

### Alternatives Considered:

*(Empty)*

---

## Test Verified

- Verified with `tox -e test-unit`, `tox -e lint-check`, and `tox -e type-check`. All tests passed.

---

## Related Issues

- Resolves #940

---

## Use of AI

---

# git log

commit 7fabc22c84a953ff5187af27a90d2f8e379ffc13
Author: prasanna <prasannajaga9@gmail.com>
Date:   Thu Jul 23 00:44:32 2026 +0530

    feat(openai): support structured chat content metadata for ChatCompletions (#940)
    
    Models like `google/translategemma-12b-it` require structured text content
    objects with extra metadata fields (e.g., `source_lang_code` and
    `target_lang_code`) in `/v1/chat/completions` requests. Without these fields,
    server-side Jinja chat templates fail with HTTP 400 errors.
    
    This change adds support for appending custom metadata fields into structured
    chat content objects:
    
    - Added `content_extra_fields` to `OpenAIHTTPBackendArgs` and forwarded it to
      `ChatCompletionsRequestHandler._format_prompts()`.
    - Added the generic `--append-payloads` CLI option to pass key-value metadata.
    - Enabled preservation of dictionary prompt objects from custom dataset rows in
      `text_column`.
    - Updated `GenerativeRequestFinalizer.finalize_turn()` to safely extract string
      content from dictionary prompts when computing character and word usage metrics.
    - Added comprehensive unit test coverage under `tests/unit/backends/openai/` and
      `tests/unit/data/`.
    
    Resolves #940
    
    Signed-off-by: prasanna <prasannajaga9@gmail.com>

commit ccd9a479a5bf547c31c683eba555197432043681
Author: prasanna <prasannajaga9@gmail.com>
Date:   Mon Jul 27 21:46:39 2026 +0530

    configure text content fields through extras
    
    Replace --append-payloads and structured dataset prompt metadata with
    backend extras.content. Apply the configured fields to generated text
    content for Chat Completions and Responses, including conversation history.
    
    Co-Authored-By: Prasanna <123716600+Prasannajaga@users.noreply.github.com>
    Signed-off-by: prasanna <prasannajaga9@gmail.com>

commit 4fd5f21650c8d041b785e7d962a56f43ed32b086
Author: prasanna <prasannajaga9@gmail.com>
Date:   Tue Jul 28 09:11:24 2026 +0530

    updated the reviewed changes
    
    Signed-off-by: prasanna <prasannajaga9@gmail.com>

commit cec92212fefd219e2d17cac5dd7799fb38e02a4a
Author: prasanna <prasannajaga9@gmail.com>
Date:   Wed Jul 29 12:24:05 2026 +0530

    reviewed changes removing the helper function
    
    Signed-off-by: prasanna <prasannajaga9@gmail.com>

---------

Co-Authored-By: Prasanna <123716600+Prasannajaga@users.noreply.github.com>
Signed-off-by: prasanna <prasannajaga9@gmail.com>


src/guidellm/cli/run.py
```diff
@@ -68,6 +85,15 @@
...
 @registry_options_from_model(model=BenchmarkArgs, group_key="spec")
+@click.option(
+    "--append-payloads",
+    callback=_parse_append_payloads,
+    help=(
+        "Append key-value fields to structured content objects in chat completion "
+        "requests. Values override matching metadata from custom datasets. "
+        'Example: `--append-payloads \'{"key":"value"}\'`'
+    ),
+)
```
dbutenhof: You've already defined `append_payloads` as a backend args parameter, so it should be passed as part of `--backend '{"kind":"openai_http","request_format":"/v1/chat/completions","append_payloads":{"source_lang_id":"en","target_lang_id":"es"}}'

We don't need another global CLI option for this: so this option and the helper function should be removed.

src/guidellm/backends/openai/http.py
```diff
@@ -208,6 +223,24 @@ def validate_server_history(self):
...
 
+    @model_validator(mode="after")
+    def validate_append_payloads(self):
+        """Validate append payloads are safe and scoped to chat completions."""
+        if not self.append_payloads:
+            return self
+        if self.request_format != "/v1/chat/completions":
+            raise ValueError(
+                "append_payloads is only supported with the "
+                "'/v1/chat/completions' request format"
```
dbutenhof: Only chat completions? Not, for example, responses?

src/guidellm/cli/run.py
```diff
@@ -106,6 +132,12 @@ def run(**kwargs):  # noqa: C901, PLR0915
     disable_console_interactive = (
         kwargs.pop("disable_console_interactive", False) or disable_console
     )
+    append_payloads = kwargs.pop("append_payloads", None)
+    if append_payloads is not None:
+        spec = kwargs.setdefault("spec", {})
+        backend = spec.setdefault("backend", {})
+        backend.setdefault("kind", "openai_http")
+        backend["append_payloads"] = append_payloads
```
dbutenhof: Remove this, too -- no new handling is required at the CLI level.

dbutenhof (changes requested):
I'm still evaluating some of the changes, but I'm going to put out a few high level comments first.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -706,17 +706,56 @@ def _ensure_tool_format(tool: dict[str, Any]) -> dict[str, Any]:
             return {"type": tool.get("type", "function"), "function": fn}
         return tool
 
+    @staticmethod
```
dbutenhof: Ah, I see what I was missing in my first pass.

This is a simple solution for a constructed dataset like the example in your documentation -- but it requires a single dataset field with a JSON structure containing the exact fields needed for the translategemma model: text, source_lang_code, target_lang_code. Maybe that's a first step, although I'm not thrilled about the special-case interpretation code here in the request handler. (We have some work underway to handle JSON columns in datasets more generally, at the data/mapper layers.)

My bigger concern here is that you're counting on a fairly specialized JSON format to supply the content type/text/language-keys. There are, for example, several huggingface datasets with names that seem targeted to google/translategemma models and none of them have this format. We'll need to be able to map from separate *columns* to these content keys.

The reason GuideLLM supports column mapping (`--data-column-mapper`) is that dataset keys vary. Again, in the few translategemma datasets on huggingface, one use "src" and "tgt" for the from-/to- language keys, while another uses "language" and "target_language"; we need to allow for such variations.

I think it's fine to have a "first level" PR that just adds a backend parameter like your "append_payloads" to provide static data for those fields, and we (or you) can add full dataset support with column mapping later -- but this is a special case "halfway hack" that I'd rather avoid.

src/guidellm/backends/openai/http.py
```diff
@@ -139,6 +139,21 @@ class OpenAIHTTPBackendArgs(BackendArgs):
         default=None,
         description="Additional parameters to include in generation requests.",
     )
+    append_payloads: dict[str, Any] | None = Field(
```
dbutenhof: "Append payloads" sounds far too generic. Maybe something like "extra_content" would be better. In fact a much cleaner option would be to put this under the existing "extras" parameter. Right now that uses the `GenerationRequestArguments` object, which allows setting headers, params, body, and adding files ... perhaps a `content` field with your new semantics would fit nicely in that model, with k/v pairs that would be added to the 'content' in each applicable request.

```console
guidellm run --backend '{
    "kind":"openai_http",
    "target":"<url>",
    "extras": {
        "content": {
            "metadata": {"category": "support"},
            "priority": 1
        }
    }
```

(I already have some concerns with the fact that a lot of OpenAI HTTP-specific classes & fields are defined globally in `schemas` as if they were universal, but we haven't gotten around to worrying about that and this isn't the right context. For now, nobody else would use this anyway.)

dbutenhof (changes requested):

src/guidellm/cli/run.py
Prasannajaga: curious to know though we have this inside , isnt it better if we give the global CLI option here for the better UX ?

src/guidellm/backends/openai/http.py
Prasannajaga: sorry I missed out! added /responses as well

src/guidellm/cli/run.py
Prasannajaga: removed

src/guidellm/backends/openai/http.py
Prasannajaga: makes complete sense! ill update the implementation to use extras.content!

src/guidellm/backends/openai/request_handlers.py
Prasannajaga: Thanks for explaining this! I want to make sure I'm following the right approach.

I was a bit confused about how you'd prefer to structure this 
*  should I go ahead and address the full dataset column mapping (--data-column-mapper) as part of this PR ? or 
* would you prefer keeping this PR strictly focused on static backend payload overrides (extras.content) and leaving --append-fields & dataset column mapping for a separate PR ?

Let me know how you'd like me to proceed!

dbutenhof: > I was a bit confused about how you'd prefer to structure this

Yeah, I have mixed thoughts on this one.

> * should I go ahead and address the full dataset column mapping (--data-column-mapper) as part of this PR ? or

I'd rather keep this relatively simple and skip a "full solution" that involves new dataset column mapping. So I'd say that's a "no".

> * would you prefer keeping this PR strictly focused on static backend payload overrides (extras.content) and leaving --append-fields & dataset column mapping for a separate PR ?

1. The minimal change to support the TranslateGemma model is introducing the source/target language codes, and the `extras.content` model does that relatively straightforwardly.
2. Your JSON dataset prompt field decoding works for a custom-made JSONL data source, but isn't compatible with the TranslateGemma datasets I found on Huggingface; and while nested JSON fields seem to be increasingly common in some other areas (our team is dealing with this in a more general way for trace replay use cases), I worry that the way it's inserted here is perhaps too special case to be supportable.

> Let me know how you'd like me to proceed!

I think my preference at this point would be to stick with the `extras.content` static injection for this PR, and then expand it later with real dataset column mapping support.

src/guidellm/cli/run.py
dbutenhof: For one thing, it's not a "global CLI option" because it only applies to the chat completions and responses request APIs on the openai_http backend. One of the motives behind the "kind" discriminated models was to better collect related parameters (e.g., the HTTP backends have `api_key` and `target`, but the in-process backends don't).

The main advantage of breaking it out here is that with multi-level options it can be tricky to specify the discriminated option without going to quoted serialized JSON, which is awkward and error-prone to type, whereas the broken-out version can be flatter and easier to use in the simplified k=v syntax. But we'd rather not add global CLI syntax for options that aren't global ... it's messier, harder to document, and harder to support.

And this actually works, without serialized JSON ... 😁 

```console
$ guidellm run --backend kind=openai_http,extras.content.source_lang_code=en,extras.content.target_lang_code=es
```

src/guidellm/backends/openai/request_handlers.py
Prasannajaga: Thanks for the clarification. As per our discussion, I have made the required changes: removed --append-payloads and the custom dataset mappings, and kept only extras.content.

src/guidellm/cli/run.py
Prasannajaga: this make sense and must be fun to be working around with you @dbutenhof, you are like GPT but much cleaner version lol, I love the vibes

src/guidellm/schemas/request.py
```diff
@@ -72,6 +72,16 @@ class GenerationRequestArguments(StandardBaseDict):
         default=None,
         description="Files to include in the request, if applicable.",
     )
+    # used this explicitly here so backend ``extras.content`` is typed, validated,
+    # and included in generated schemas instead of existing only as an allowed
+    # arbitrary field on StandardBaseDict.
```
jaredoconnell: Isn't that the case for all of the other fields, too? What value does this comment provide?

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -193,6 +193,21 @@ def _check_streaming_error(data: Any) -> None:
...
+    """Extract content-object fields from generation request extras.
+
+    :param extras: Additional generation request arguments.
+    :return: Fields to merge into generated text content objects.
+    """
+    if isinstance(extras, GenerationRequestArguments):
+        return extras.content
+    if isinstance(extras, dict):
+        return extras.get("content")
+    return None
```
jaredoconnell: Why support both of these?

Prasannajaga: in lot of existing places in test cases we passing the dict as args in the `handler.format()` for the cases it fails I added this for handle those cases 

these are existing places we passed as dict instead of valid type 
test_tool_choice_none_when_expects_false
test_tool_choice_preserved_when_expects_true
test_auto_tool_choice_preserved_when_expects_true

I will update here and raise this again

src/guidellm/schemas/request.py
Prasannajaga: since this is differ from the normal payloads I thought adding this here, I will remove this

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -193,6 +193,17 @@ def _check_streaming_error(data: Any) -> None:
...
 
+def _get_content_extras(
+    extras: GenerationRequestArguments | None,
+) -> dict[str, Any] | None:
+    """Extract content-object fields from generation request extras.
+
+    :param extras: Additional generation request arguments.
+    :return: Fields to merge into generated text content objects.
+    """
+    return extras.content if extras is not None else None
```
jaredoconnell: I talked with my teammates, and let's inline this. It's only used in 4 or 5 places.

Prasannajaga: done!

jaredoconnell (approved):

dbutenhof (approved):
Sometimes I wish Python had Javascript's `?.` operator!

PR #948, opened 2026-07-22
Create a prototype test plan

dbutenhof:
## Summary

0.7.0 broke several things, and we were sloppy about testing. 0.7.1 fixed JSON output with an `api_key`, but not CSV output. We need to be more disciplined about pre-release testing.

## Details

I've considered trying to build a formal manual test plan for some time; a few recent issues brought that back to mind, and I thought I'd make a try to start something.

I began with a few bullet items and thought I'd send that out for review, when the idea struck me to feed my bullet list and some general instructions into Cursor and see what it came up with.

This is the result. I haven't even reviewed it in great detail, but at a glance most of this seems plausible, if not complete. At least, it seems like a place to start; and I like the idea of "enshrining" this in our documentation. Although, ideally, we'll migrate this into our CI, extending or replacing our e2e tests.

## Test Plan

- [ ] yes, we should test the test plan!

## Related Issues

Related to all issues, closes none!

---

## Use of AI

---

# git log

commit 9ff28786cbe8ac303be75099c2cef4920e94699a
Date:   Wed Jul 22 16:37:02 2026 -0400

    Create a prototype test plan
    
---------

### DIFF
--- a/docs/developer/release-test-plan.md
+++ b/docs/developer/release-test-plan.md
@@ -1,0 +1,480 @@
+---
+weight: 0
+---
+
+# Release Test Plan
+
+Manual checklist for verifying a GuideLLM release candidate. The goal is reasonable confidence that featured paths work end-to-end and that known regression hotspots have not resurfaced.
+
+This plan **complements** automated gates (`tox`, unit/integration/e2e). It does not replace them. Prefer short, constrained runs (`max_requests` / short `max_duration`) unless a section says otherwise.
+
+**Out of scope:** WEKA / OTEL trace replay (still under active development).
+
+## Prerequisites
+
+Install a release candidate (or local checkout) with the extras you will exercise:
+
+```bash
+# From PyPI / wheel under test
+pip install "guidellm[recommended,vision,audio,plot]"
+
+# Or from a local checkout
+uv sync --extra recommended --extra vision --extra audio --extra plot
+```
+
+Commands below use `guidellm …`. From a local checkout, prefer `uv run guidellm …`.
+
+| Need                                     | Notes                                                                           |
+| ---------------------------------------- | ------------------------------------------------------------------------------- |
+| OpenAI-compatible HTTP server            | Most `openai_http` sections (chat, ASR, embeddings, VLM)                        |
+| vLLM (or compatible) with `/v1/realtime` | Websocket audio transcription                                                   |
+| Hugging Face Hub access                  | HF dataset sections (cache warmed helps)                                        |
+| In-process vLLM install                  | `vllm_python` / gated `vllm_offline`                                            |
+| Results workspace                        | Use a dedicated temp directory; set `GUIDELLM__DEFAULT_RESULTS_DIR` where noted |
+
+Suggested servers / models (adjust to what you have available):
+
+| Role         | Example                                         |
+| ------------ | ----------------------------------------------- |
+| Chat / text  | Any chat-completions model on `:8000`           |
+| VLM          | `vllm serve Qwen/Qwen3-VL-2B-Instruct`          |
+| ASR (HTTP)   | `vllm serve openai/whisper-small`               |
+| Embeddings   | `vllm serve BAAI/bge-small-en-v1.5 --port 8000` |
+| In-line vLLM | Small model e.g. `Qwen/Qwen3-0.6B`              |
+
+## Baseline automated gate
+
+Run before the manual matrix (from a clone of the release tag / candidate):
+
+```bash
+tox -e lint-check
+tox -e type-check
+tox -e tests
+# Optional but recommended when infrastructure allows:
+# docker build . -f tests/e2e/vllm-sim.Dockerfile -o type=local,dest=./
+# tox -e test-e2e
+```
+
+**Pass:** all selected tox environments exit 0.
+
+______________________________________________________________________
+
+## Cross-cutting checks
+
+### 1. `api_key` / `SecretStr` (regression hotspot)
+
+`api_key` is stored as Pydantic `SecretStr`. This has broken in several forms: key not forwarded on the wire, or raw secret leaking into reports.
+
+**Setup:** Any HTTP endpoint that records request headers (mock server, proxy, or `guidellm mock-server` plus a packet/header inspector). An authenticated backend is **not** required—only that the `Authorization: Bearer …` header is present when `api_key` is set.
+
+**CLI path:**
+
+```bash
+export RESULTS_DIR="$(mktemp -d)"
+guidellm run \
+  --backend kind=openai_http,target=http://localhost:8000,api_key=sk-release-test-secret-do-not-leak \
+  --profile kind=synchronous \
+  --constraint kind=max_requests,count=5 \
+  --data kind=synthetic_text,prompt_tokens=64,output_tokens=16 \
+  --output kind=json,path="${RESULTS_DIR}/api_key.json" \
+  --output kind=csv,path="${RESULTS_DIR}/api_key.csv" \
+  --output kind=html,path="${RESULTS_DIR}/api_key.html" \
+  --output kind=console
+```
+
+**Env path (also verify):**
+
+```bash
+GUIDELLM__SPEC__BACKEND__API_KEY=sk-env-secret-do-not-leak \
+guidellm run \
+  --backend kind=openai_http,target=http://localhost:8000 \
+  --profile kind=synchronous \
+  --constraint kind=max_requests,count=5 \
+  --data kind=synthetic_text,prompt_tokens=64,output_tokens=16 \
+  --output kind=json,path="${RESULTS_DIR}/api_key_env.json"
+```
+
+**Pass criteria:**
+
+- Outbound HTTP includes `Authorization: Bearer sk-release-test-secret-do-not-leak` (CLI) / env secret (env path).
+- JSON / CSV / HTML / console output **do not** contain the raw secret string (value must be masked or absent).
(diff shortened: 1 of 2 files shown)


docs/developer/release-test-plan.md
```diff
@@ -0,0 +1,480 @@
...
+  --constraint kind=max_requests,count=10 \
+  --data kind=synthetic_text,prompt_tokens=128,output_tokens=32 \
+  --output kind=console \
+  --output kind=json,path="${OUT}/run.json" \
+  --output kind=csv,path="${OUT}/run.csv" \
+  --output kind=html,path="${OUT}/run.html" \
+  --output kind=plot,path="${OUT}/run.png",dpi=100
+```
+
+**B. `guidellm benchmark from-file`**
```
sjmonson, line 146: Is this `guidellm export` now?

docs/developer/release-test-plan.md
```diff
@@ -0,0 +1,480 @@
...
+
+______________________________________________________________________
+
+## Known gaps (context for testers)
+
+- `kind=plot` is implemented but not yet documented in the outputs guide; always pass an explicit `path=` for release runs.
+- Websocket realtime ASR has little user documentation; rely on this checklist and code/tests.
+- WEKA / OTEL trace replay is intentionally excluded until implementation lands.
+- `vllm_offline` may be absent from the candidate—record N/A rather than Fail.
+- Hugging Face multimodal paths are thinly covered by automated e2e; this manual matrix is the primary confidence gate for Hub datasets.
```
sjmonson, line 480: These don't seem to helpful:

> - `kind=plot` is implemented but not yet documented in the outputs guide; always pass an explicit `path=` for release runs.
> - Websocket realtime ASR has little user documentation; rely on this checklist and code/tests.

Shouldn't we just fix the missing documentation?

> - WEKA / OTEL trace replay is intentionally excluded until implementation lands.
> - `vllm_offline` may be absent from the candidate—record N/A rather than Fail.

Is it really a gap that we don't test things that don't exist yet?

> - Hugging Face multimodal paths are thinly covered by automated e2e; this manual matrix is the primary confidence gate for Hub datasets.

This is not really saying anything.

docs/developer/release-test-plan.md
```diff
@@ -0,0 +1,480 @@
...
+| 9   | HF text                                            |                            |        |                        |
+| 10  | HF audio (HTTP transcription)                      |                            |        |                        |
+| 11  | HF image                                           |                            |        |                        |
+| 12  | HF video                                           |                            |        |                        |
+| 13  | Websocket realtime transcription                   |                            |        |                        |
+| 14  | Embeddings API                                     |                            |        |                        |
+| 15  | `vllm_python` (normal)                             |                            |        |                        |
+| 16  | `vllm_offline` (batch)                             |                            |        | Skip if not in release |
+| 17  | `trace_synthetic` (optional)                       |                            |        |                        |
```
sjmonson, line 467: Why optional?

docs/developer/release-test-plan.md
```diff
@@ -0,0 +1,480 @@
...
+## Prerequisites
+
+Install a release candidate (or local checkout) with the extras you will exercise:
+
+```bash
+# From PyPI / wheel under test
+pip install "guidellm[recommended,vision,audio,plot]"
+
+# Or from a local checkout
+uv sync --extra recommended --extra vision --extra audio --extra plot
```
sjmonson, line 22: Why not just `all`?

sjmonson (changes requested):
Overall I think this is a good idea, though maybe I would label it as "release smoke test plan" since it does not go into performance testing. I didn't verify any of the examples but they look fine. Some minor nits below:

docs/developer/release-test-plan.md
dbutenhof: These were generated by Cursor, and I only glanced through quickly before posting to give everyone a chance at it. But, yeah; it's clearly made some questionable choices and I probably shouldn't have been so quick to post. 😆

docs/developer/release-test-plan.md
dbutenhof: Yeah -- this went up before Jared made that change.

docs/developer/release-test-plan.md
dbutenhof: Great question. I infer that Cursor thought of this as a clean subset of Mooncake replay that didn't add much on its own (note that the `trace_synthetic` test above is "Optional prelude") -- but that seems questionable logic to me since Mooncake relies on the hash-id sequences which are missing here.

docs/developer/release-test-plan.md
dbutenhof: Yeah -- Cursor, interestingly, analyzed outstanding PRs rather than just the current branch, which surprised me, and anticipated with a lot of additional test cases we can't yet perform. I corrected it in a few cases, or just edited the output; but in retrospect I should probably have just deleted this whole section.

docs/developer/release-test-plan.md
```diff
@@ -0,0 +1,480 @@
...
+| 17  | `trace_synthetic` (optional)                       |                            |        |                        |
+| 18  | Mooncake replay                                    |                            |        |                        |
+
+**Release candidate:** \_\_\_\_\_\_\_\_\_\_\_\_\_\_\_ **Date:** \_\_\_\_\_\_\_\_\_\_\_\_\_\_\_ **Overall verdict:** Pass / Fail
+
+______________________________________________________________________
+
+## Known gaps (context for testers)
+
+- `kind=plot` is implemented but not yet documented in the outputs guide; always pass an explicit `path=` for release runs.
```
dbutenhof, line 476: This is intriguing, since Cursor has just casually diagnosed a hole in the documentation that's worth another PR.

docs/developer/release-test-plan.md
```diff
@@ -0,0 +1,480 @@
...
+```
+
+**Pass criteria:**
+
+| Kind      | Expect                                                 |
+| --------- | ------------------------------------------------------ |
+| `console` | Metadata / info / stats tables print without traceback |
+| `json`    | Valid JSON; contains benchmarks and metrics            |
+| `csv`     | Non-empty summary rows                                 |
+| `html`    | Opens in a browser; tables/charts render               |
```
SkiHatDuckie, line 164: The HTML output for me has consistently returned an empty file (I'm assuming this is due to the output type not being maintained). I also don't seem to be the only one who has experienced this: #941 . Unless the state of the HTML option changes, this is a bit misleading.

docs/developer/release-test-plan.md
```diff
@@ -0,0 +1,480 @@
...
+### 18. Mooncake
+
+Create a minimal Mooncake-format trace (see [Trace File Formats](../guides/trace_replay.md)):
+
+```bash
+export MOONCAKE="$(mktemp -d)/mooncake.jsonl"
+cat > "${MOONCAKE}" <<'EOF'
+{"timestamp": 0.0, "input_length": 10, "output_length": 5, "hash_ids": [0]}
+{"timestamp": 0.2, "input_length": 20, "output_length": 8, "hash_ids": [0, 1]}
+{"timestamp": 0.4, "input_length": 15, "output_length": 6, "hash_ids": [2]}
```
SkiHatDuckie, line 426: The hash IDs here are malformed. With the Mooncake default block size, you'd need at least 513 input tokens to get 2 hash IDs. You could either set the `hash_id_block_size` to something like 15, or remove either the 0 or 1 in the second row. Both should work.

Edit: Or increase the input_length to something between 512 and 1024.

docs/developer/release-test-plan.md
```diff
@@ -0,0 +1,480 @@
...
+EOF
+
+guidellm run \
+  --backend kind=openai_http,target=http://localhost:8000 \
+  --data kind=mooncake,path="${MOONCAKE}" \
+  --profile kind=replay,time_scale=1.0 \
+  --output kind=json,path=results/mooncake.json
+```
+
+Optional column overrides: `hash_ids_column`, `hash_id_block_size` (default `512`), plus the shared `timestamp_column` / `prompt_tokens_column` / `output_tokens_column`.
```
SkiHatDuckie, line 436: `--pedantic`, but `hash_id_block_size` isn't a *column* override.

docs/developer/release-test-plan.md
SkiHatDuckie: I noticed that as well. This might also lead to having to regularly read through and update this doc to keep it up to date.

SkiHatDuckie (commented):
Some things I found. Main gripe is with the "xyz is missing/in active development" notes, as this doc will have to be regularly rechecked to see if notes need to be added or deleted.

docs/developer/release-test-plan.md
dbutenhof: Interesting. I've always seen it generate a file with data; but until Sam's recent fix it just didn't *render*. The fix does require a new HTML file generated with the GuideLLM fix since it's really just locking to a specific github.io URL. But I just generated a new html from main, and it renders just fine.

On the other hand, the description wording here loosely suggests that something will magically open the HTML in a browser, while obviously you need to do that manually.

docs/developer/release-test-plan.md
dbutenhof: Thanks ... I posted this without trying to proof-read in depth, although it's possible I might not have noticed that detail anyway. 😆

docs/developer/release-test-plan.md
dbutenhof: I don't think I like Cursor's term "override" anyway here -- thanks for pointing that out.

PR #970, opened 2026-07-27
Prevent .env files from tainting tests

jaredoconnell:
## Summary

This is a follow up to #751. I discovered that .env files were not addressed by my prior fix.

## Details

- Just instructs the tests to ignore the .env file.

## Test Plan

Just run the tests normally and with a .env file that includes values that cause problems with the tests, like `GUIDELLM__LOGGING__CONSOLE_LOG_LEVEL=DEBUG`

## Related Issues

- Relates to #751 

---

## Use of AI

---

# git log

commit 0acbbb4f67dbf9bb27982a44f77981f1f81b1202
Date:   Mon Jul 27 17:46:08 2026 -0400

    Prevent .env files from tainting tests
    
---------

### DIFF
--- a/tests/unit/test_settings.py
+++ b/tests/unit/test_settings.py
@@ -22,7 +22,7 @@ def test_default_settings(mocker):
         {k: v for k, v in os.environ.items() if not k.startswith("GUIDELLM__")},
         clear=True,
     )
-    settings = Settings()
+    settings = Settings(_env_file=None)
     assert settings.logging == LoggingSettings()
     assert settings.report_generation.source.startswith(BASE_URL)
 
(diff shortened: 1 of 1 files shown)


dbutenhof (approved):

sjmonson:
Maybe we should just disable `.env` support entirely? Since its only for the old settings object it doesn't provide much value to users.

jaredoconnell:
> Maybe we should just disable `.env` support entirely? Since its only for the old settings object it doesn't provide much value to users.

I find it useful for always setting the two settings that I need; one to make it so multiprocessing works on mac, and the other to enable debug logging.

sjmonson (approved):

PR #978, opened 2026-07-31
Update click to 8.4

dbutenhof:
## Details

A transitive dependency collision in the AIPCC build environment caused us to bind transformers < 5.0, and failing to resolve the transformers CVEs for which we released 0.7.2.

Basically, the crux was huggingface_hub: we have bound upstream to 1.16.0, while the AIPCC index jumps from 0.38 to 1.16.4. 1.16.4 added an explicit dependency on click 8.4, which caused us to bind 0.38 and an equally old transformers.

We resolve this by bumping click to 8.4. In the AIPCC build base image, this allows huggingface_hub 1.23 and transformers 5.14.1, which does not have the CVEs at issue.

## Test Plan

- I ran a container build using (essentially) the AIPCC Containerfile, on the 3.5 builder base image (but copying in my local source and using `pip install ".[all]"`), and verified that it binds transformers 5.14.1.
- A local `trivy` scan shows no CRITICAL/HIGH CVEs on the resulting image.

## Related Issues

N/A

---

## Use of AI

---

# git log

commit fdb89ab0ce2a942ad34364428840cac91ac80b14
Date:   Fri Jul 31 14:18:03 2026 -0400

    Update click to 8.4
    
    A transitive dependency collision in the AIPCC build environment caused us to
    bind transformers < 5.0, and failing to resolve the transformers CVEs for
    which we released 0.7.2.
    
    Basically, the crux was huggingface_hub: we have bound upstream to 1.16.0,
    while the AIPCC index jumps from 0.38 to 1.16.4. 1.16.4 added an explicit
    dependency on click 8.4, which caused us to bind 0.38 and an equally old
    transformers.
    
    We resolve this by bumping click to 8.4. In the AIPCC build base image, this
    binds to transformers 5.14.1, which does not have the CVEs at issue.
    
    I ran a container build using (essentially) the AIPCC Containerfile, on the
    3.5 builder base image (but copying in my local source and using
    `pip install ".[all]"`), and verified that it binds transformers 5.14.1.
    
    A local `trivy` scan shows no CRITICAL/HIGH CVEs on the resulting image.
    
---------

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,7 +59,7 @@ keywords = [
     "vllm",
 ]
 dependencies = [
-    "click~=8.3.0",
+    "click~=8.4.0",
     "culsans~=0.10.0",
     "datasets>=4.1.0",
     "eval_type_backport",
--- a/src/guidellm/utils/default_group.py
+++ b/src/guidellm/utils/default_group.py
@@ -95,7 +95,7 @@ class DefaultCommandFormatter(click.HelpFormatter):
     def __getattr__(self, attr):
         return getattr(self.formatter, attr)
 
-    def write_dl(self, rows: cabc.Sequence[tuple[str, str]], *args, **kwargs):
+    def write_dl(self, rows: cabc.Iterable[tuple[str, str]], *args, **kwargs):
         rows_: list[tuple[str, str]] = []
         for cmd_name, help_msg in rows:
             if cmd_name == self.group.default_cmd_name:
--- a/tests/unit/cli/test_preprocess_dataset.py
+++ b/tests/unit/cli/test_preprocess_dataset.py
@@ -112,7 +112,7 @@ def test_preprocess_dataset_rejects_legacy_config_flag():
     )
 
     assert result.exit_code != 0
-    assert "No such option: --config" in result.output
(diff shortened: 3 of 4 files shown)


SkiHatDuckie (approved):
Where could I find the AIPCC (or equivalent) Containerfile? Otherwise not much to see in the diffs.

jaredoconnell (approved):
I did a basic test and it works for me.

sjmonson (approved):

dbutenhof:
> Where could I find the AIPCC (or equivalent) Containerfile? Otherwise not much to see in the diffs.

I won't put the internal link here, but I can share it if you're curious. (You can access the internal wheels and look at the repo as long as you're in the Red Hat domain, but you can't see the container images because they're on a protected repo.) The main thing is that it uses `pip install guidellm[all]` from the built-in AIPCC baseimage internal index, and pulls all dependencies from that index -- which was the cause of the transitive dependency mess since it didn't have the huggingface_hub version our upstream container uses (which was compatible both with transformers 5.5 and click 8.3). The APICC index has a very old huggingface_hub (that's not compatible with transformers 5) and a newer huggingface_hub that requires click 8.4.

PR #982, opened 2026-08-03
fix(openai): allow opting out of ignore_eos for end-of-turn-sensitive models (#979)

### LINKED ISSUE #979: [Bug] gpt-oss benchmarks silently error out: unconditional ignore_eos is incompatible with Harmony chat completions
**Describe the bug**

Benchmarking `openai/gpt-oss-120b` on vLLM over `/v1/chat/completions` fails most of its requests, and the failure is easy to misread as server saturation rather than a client/server incompatibility.

GuideLLM sets `ignore_eos: true` unconditionally on chat completions (`backends/openai/request_handlers.py`). For gpt-oss, `<|return|>` is a Harmony stop token; `ignore_eos` suppresses stopping on it, generation continues past the end of the turn, and vLLM's Harmony parser then rejects the trailing tokens:

```
ValueError('Streaming response returned an error: Unexpected token 200002 while expecting start token 200006')
```

Measured on 1x H200, vLLM 0.18.0, sweep profile, ISL/OSL 1024/1024:

| sweep point | total | successful | errored |
|---|---|---|---|
| synchronous | 40 | 9 | 31 |
| throughput | 644 | 184 | 397 |
| constant (8 rates) | 858 | 280 | 542 |

**971 of 1542 requests (65%) failed**, all with that message. The same config at OSL 128 errored 0 times — the model rarely finishes a turn within 128 tokens, so it never emits the terminator. The error rate therefore scales with output length.

**Why this matters beyond the errors**

The report still produced a goodput number. Percentiles are computed over successful requests only, so a load level where two of three requests fail looks like a clean pass. Anyone benchmarking gpt-oss with a `sweep` profile gets a plausible-looking number derived from the ~35% of requests that survived.

**Expected behavior**

Either:

1. Don't send `ignore_eos` when the target model uses a structured response format (Harmony/gpt-oss). There's already precedent for exactly this in the same file — `ignore_eos` is dropped for structured output, with the comment that it "would force generation past that point". Harmony is the same class of problem.
2. Use `min_tokens` instead of `ignore_eos` (the direction of #533), which pins output length without suppressing the turn terminator.
3. At minimum, make it configurable so a benchmark can opt out per model.

`/v1/completions` is unaffected — no Harmony parsing on that path — so `--backend kind=openai_http,...` pointed at text completions is the current workaround.

**Environment**

1. OS: Ubuntu 22.04
2. Python: 3.12
3. GuideLLM: 0.7.2
4. vLLM: 0.18.0 and 0.26.0 (both affected)

**To Reproduce**

```bash
vllm serve openai/gpt-oss-120b --max-model-len 32768 --tensor-parallel-size 1

guidellm run \
  --backend kind=openai_http,target=http://127.0.0.1:8000,model=openai/gpt-oss-120b \
  --profile kind=sweep \
  --data kind=synthetic_text,prompt_tokens=1024,output_tokens=1024 \
  --constraint kind=max_duration,seconds=180 \
  --output kind=json,path=report.json
```

Then check `benchmarks[*].metrics.request_totals.errored` in the report.

The server-side half reduces to a single request:

```bash
curl -s http://localhost:8000/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model": "openai/gpt-oss-120b",
  "messages": [{"role": "user", "content": "Say hi."}],
  "max_completion_tokens": 512,
  "ignore_eos": true
}'
```

**Related**

- The server-side defect is filed at vllm-project/vllm#50690 — vLLM lets `ignore_eos` override its own Harmony stop tokens and then 500s on the result. It should be fixed there too, but GuideLLM sending `ignore_eos` to a Harmony model is the trigger, and prior art for the fix already exists in this repo (#533, and the structured-output exception).

Anai-Guo:
Fixes #979.

## Problem

`ignore_eos` is set unconditionally on all three OpenAI generation paths whenever an output-token count is requested:

| path | site |
|---|---|
| `/v1/completions` | `request_handlers.py` — `arguments.body["ignore_eos"] = True` |
| `/v1/chat/completions` | `request_handlers.py` — `body.update({..., "ignore_eos": True})` |
| `/v1/responses` | `request_handlers.py` — `arguments.body["ignore_eos"] = True` |

For a model whose response format carries its own end-of-turn token, suppressing EOS makes generation continue past the end of the turn and the server rejects what follows. On `openai/gpt-oss-120b` (Harmony, `<|return|>`) the reporter measured **971 of 1542 requests failing (65%)** with:

```
ValueError('Streaming response returned an error: Unexpected token 200002 while expecting start token 200006')
```

The error rate scales with output length — at OSL 128 the model rarely finishes a turn, so nothing errors; at OSL 1024 most requests do. Percentiles are computed over successful requests only, so the run still reports a plausible-looking number derived from the ~35% that survived.

## Change

The same file already carves this out for tool calling, where `ignore_eos` is popped because constrained decoding needs the model to stop on its own:

> Tool calling requires the model to stop naturally after producing valid JSON; `ignore_eos` would force generation past that point and break the server's constrained decoding grammar.

Harmony is the same class of problem, so this adds a backend-level knob rather than a model-name heuristic (issue option 3):

```bash
guidellm run \
  --backend kind=openai_http,target=http://127.0.0.1:8000,model=openai/gpt-oss-120b,ignore_eos=false \
  --profile kind=sweep \
  --data kind=synthetic_text,prompt_tokens=1024,output_tokens=1024
```

- `OpenAIHTTPBackendArgs.ignore_eos: bool = True` — default unchanged, so existing benchmarks behave identically.
- Threaded through the existing `request_handler.format(...)` kwargs alongside `stream` / `max_tokens`, and honored at all three sites.
- When disabled the key is **omitted** rather than sent as `false`, matching how the handlers already drop `None`-valued body keys via `deep_filter`.
- `max_tokens` / `max_completion_tokens` / `max_output_tokens` and `stop` are untouched, so output length is still pinned.

Options 1 (auto-detect structured response formats) and 2 (`min_tokens`, the direction of #533) are larger calls about default behavior and are left to maintainers; this only makes the current behavior opt-out-able.

## Tests

Added `test_format_ignore_eos_disabled` to each of the three handler test classes, asserting `ignore_eos` is absent while the token limit still applies. Existing tests that assert `ignore_eos is True` are unchanged and still pass, covering the default.

`ruff format --check` and `ruff check` clean on all three touched files.

---

# git log

commit dae0dc39928e11404fe9abd28ecc9772abfc900f
Date:   Tue Aug 4 05:00:00 2026 +0000

    docs(backends): document disabling ignore_eos for gpt-oss models (#979)
    
    Per review on #982, the openai_http backend already supports opting out
    of ignore_eos via extras.body.ignore_eos=false, so the code change is
    unnecessary. Instead, document the end-of-sequence behavior and how to
    disable it for Harmony / gpt-oss and other end-of-turn-sensitive models
    in the backends guide.
    
commit 55045d213e2d7a59196456464371ee497d1bf4da
Date:   Wed Aug 5 09:20:00 2026 +0000

    docs(backends): note the tradeoff of disabling ignore_eos
    
    Per review feedback on #982, call out that turning off ignore_eos lets
    the model stop at its own discretion, which can produce sequences much
    shorter than the requested output length.
    
commit f75e9ee8f683b3eeae5712115ce122824df4d220
Date:   Wed Aug 5 19:04:19 2026 +0000

    docs(backends): apply mdformat to backends guide
    
    Run pre-commit mdformat over docs/backends.md so precommit-checks passes;
    no content change.
    
---------

### DIFF
--- a/docs/guides/backends.md
+++ b/docs/guides/backends.md
@@ -151,6 +151,9 @@ guidellm run \
 
 Values in `extras.body` are merged into the request body and take precedence over the defaults, so `ignore_eos: false` overrides the built-in `true`. Leave `ignore_eos` unset (the default) for models where suppressing the end-of-sequence token is safe, which lets GuideLLM control the exact output length.
 
+> [!NOTE]
+> Removing `ignore_eos: true` allows the model to stop generating at its discretion which can result in significantly shorter sequence lengths than desired.
+
 ## Structured Chat Content Payloads
 
 Some chat templates require metadata alongside the text in each structured content object. Pass these fields through `extras.content` in the `openai_http` backend configuration. GuideLLM adds them to every generated text content object for Chat Completions and Responses API requests.
--- a/docs/guides/backends.md
+++ b/docs/guides/backends.md
@@ -151,8 +151,7 @@ guidellm run \
 
 Values in `extras.body` are merged into the request body and take precedence over the defaults, so `ignore_eos: false` overrides the built-in `true`. Leave `ignore_eos` unset (the default) for models where suppressing the end-of-sequence token is safe, which lets GuideLLM control the exact output length.
 
-> [!NOTE]
-> Removing `ignore_eos: true` allows the model to stop generating at its discretion which can result in significantly shorter sequence lengths than desired.
+> [!NOTE] Removing `ignore_eos: true` allows the model to stop generating at its discretion which can result in significantly shorter sequence lengths than desired.
 
 ## Structured Chat Content Payloads
 
--- a/docs/guides/backends.md
+++ b/docs/guides/backends.md
@@ -129,6 +129,28 @@ guidellm run \
 
 This will include `temperature`, `top_p`, and `top_k` in every request body sent to the server.
 
+## Controlling End-of-Sequence Behavior
+
+To measure throughput and latency for an exact output length, GuideLLM asks the server to generate precisely the requested number of tokens. For the `openai_http` backend, when you request a specific output token count each request sets `ignore_eos: true` (alongside `max_tokens`/`max_completion_tokens` and `stop: null`) so the server keeps generating instead of stopping when the model emits an end-of-sequence token.
+
+Some models should not have their end-of-sequence token suppressed. Formats such as Harmony / `gpt-oss` expect the model to stop on its own end-of-turn token; forcing generation past it makes the server reject the trailing tokens and fail the request. For these (or similar) models, disable `ignore_eos` by passing `false` through the backend `extras.body` field:
+
+```bash
+guidellm run \
+  --backend '{"kind":"openai_http","target":"http://localhost:8000/v1","model":"openai/gpt-oss-20b","extras":{"body":{"ignore_eos":false}}}' \
+  --data kind=synthetic_text,prompt_tokens=256,output_tokens=128
+```
+
+Or using GuideLLM's compact `key=value` parser:
+
+```bash
+guidellm run \
+  --backend kind=openai_http,target=http://localhost:8000/v1,model=openai/gpt-oss-20b,extras.body.ignore_eos=false \
+  --data kind=synthetic_text,prompt_tokens=256,output_tokens=128
+```
+
+Values in `extras.body` are merged into the request body and take precedence over the defaults, so `ignore_eos: false` overrides the built-in `true`. Leave `ignore_eos` unset (the default) for models where suppressing the end-of-sequence token is safe, which lets GuideLLM control the exact output length.
+
 ## Structured Chat Content Payloads
 
 Some chat templates require metadata alongside the text in each structured content object. Pass these fields through `extras.content` in the `openai_http` backend configuration. GuideLLM adds them to every generated text content object for Chat Completions and Responses API requests.


src/guidellm/backends/openai/http.py
```diff
@@ -135,6 +135,16 @@ class OpenAIHTTPBackendArgs(BackendArgs):
         default=True,
         description="Use streaming responses for generation requests when supported.",
     )
+    ignore_eos: bool = Field(
```
dbutenhof: This is redundant. There's already a mechanism to pass extra body parameters through to the HTTP POST:

```console
$ guidellm run \
  --backend '{
    "kind": "openai_http",
    "target": "<url>",
    "extras": {"body": {"ignore_eos": false}}
    }' \
  --profile kind=constant,rate=5 \
  --constraint kind=max_duration,seconds=60
```

or (more compact but a bit weirder looking, taking advantage of GuideLLM's k=v parser):

```console
$ guidellm run \
  --backend kind=openai_http,target=<url>,extras.body.ignore_eos=false \
  --profile kind=constant,rate=5 \
  --constraint kind=max_duration,seconds=60
```

This should show 0 errors.

The only effective difference between the current behavior with extras.body.ignore_eos=false and your `ignore_eos=false` is that you're dropping the field entirely rather than specifying false, which isn't worth the code change.

While I like your description text, this should instead be documented in `docs/guides/backends.md`, possibly as a new section after `## Configuring the API Key`.

dbutenhof (changes requested):
This code change is unnecessary, as the capability already exists.

However, given the importance for the GPT OSS models (behavior I've verified locally), I would like to see expanded text resembling your Pydantic field description added to the backends guide for anyone using these (or similar) models where `ignore_eos` might be a problem.

Anai-Guo:
Thanks @dbutenhof — you're right, `extras.body.ignore_eos=false` already covers this and the code change wasn't needed. I've dropped the code/test changes entirely and force-pushed the branch so this PR is now documentation-only.

The new **"Controlling End-of-Sequence Behavior"** section in `docs/guides/backends.md` (right after *Passing Sampling Parameters*, so it builds on the `extras.body` mechanism) explains:

- why GuideLLM sends `ignore_eos: true` by default when a specific output length is requested,
- why Harmony / `gpt-oss` and other end-of-turn-sensitive models fail when it's suppressed,
- and how to disable it via `extras.body.ignore_eos=false`, in both JSON and the compact `key=value` forms.

Let me know if you'd prefer it placed directly under *API Key Configuration* instead, or want any wording tweaks.

dbutenhof (dismissed):
Beautiful. Thanks!

jaredoconnell (dismissed):

docs/guides/backends.md
```diff
@@ -129,6 +129,28 @@ guidellm run \
...
+Or using GuideLLM's compact `key=value` parser:
+
+```bash
+guidellm run \
+  --backend kind=openai_http,target=http://localhost:8000/v1,model=openai/gpt-oss-20b,extras.body.ignore_eos=false \
+  --data kind=synthetic_text,prompt_tokens=256,output_tokens=128
+```
+
+Values in `extras.body` are merged into the request body and take precedence over the defaults, so `ignore_eos: false` overrides the built-in `true`. Leave `ignore_eos` unset (the default) for models where suppressing the end-of-sequence token is safe, which lets GuideLLM control the exact output length.
+
```
sjmonson, line 153: Could you add this note or something similar (might need to check this locally with linting):


```suggestion

> [!NOTE]
> Removing `ignore_eos: true` allows the model to stop generating at its discretion which can result in significantly shorter sequence lengths than desired. 
```

sjmonson (changes requested):

sjmonson:
@mergifyio dequeue

Anai-Guo:
Added in 55045d2 @sjmonson — the note is right after the paragraph on `extras.body` precedence, so it lands as the closing caveat of the section. Wording is yours verbatim (trailing space dropped for markdownlint).

dbutenhof (changes requested):
Looks good, but the `mdformat` checker is unhappy with your change.

We use the `pre-commit` tool for basic lint & formatter checks, which by default works on staged git files before commit: but you check all the files like this:

```console
$ pip install pre-commit
$ pre-commit install
$ pre-commit run --all-files
```

That'll modify your `backends.md` to follow the mdformat rules. If you run `pre-commit run --all-files` a second time, you should see no errors or changes.

Anai-Guo:
Thanks @dbutenhof — ran `pre-commit run --all-files` and applied the `mdformat` changes to `backends.md` (it collapsed the `[!NOTE]` callout per the repo's mdformat-gfm config). precommit-checks should be green now.

sjmonson (approved):

jaredoconnell (approved):

Anai-Guo:
DCO is green now — the `mdformat` commit I pushed a moment ago went up without a `Signed-off-by` trailer. I rebuilt it on the same tree with the trailer in place, so all three commits are signed off and the full check suite (including `precommit-checks`) is passing. Sorry for the extra round, @dbutenhof @sjmonson.

dbutenhof (approved):

PR #990, opened 2026-08-03
Fix multipart filenames for audio uploads

### LINKED ISSUE #983: [Bug]: In-memory audio is uploaded without a multipart filename
## Bug Description

Audio produced by the `encode_media` preprocessor from an in-memory source can have `file_name=None`. The OpenAI transcription/translation request handler passes that value directly to HTTPX as the multipart filename.

A multipart part without a filename is interpreted by Starlette/FastAPI-compatible servers as a normal form value rather than an `UploadFile`. The server consequently rejects an otherwise valid transcription request with an error similar to:

```text
Expected UploadFile, received: <class 'str'>
```

There are two contributing behaviors:

1. `encode_audio()` documents that a filename will be derived from the resolved format, but returns `None` for in-memory inputs unless `file_name` is explicitly supplied.
2. `AudioRequestHandler.format()` uses `audio_entry.get("file_name", "audio_input")`. The fallback does not apply when the key exists with value `None`.

An explicit `file_name` also does not currently override the source basename for path inputs, despite being documented as an override.

## Expected Behavior

Every transcription or translation multipart upload should have a nonempty filename.

For example:

- An explicit `file_name` takes precedence.
- A path input uses its basename when no override is provided.
- An in-memory input defaults to something compatible with the resolved format, such as `audio.wav`.

The request handler should also defensively replace missing or empty filenames.

## Steps to Reproduce

```python
import torch

from guidellm.backends.openai.request_handlers import AudioRequestHandler
from guidellm.schemas import GenerationRequest
from guidellm.utils.audio import encode_audio

encoded = encode_audio(
    torch.zeros((1, 16000)),
    sample_rate=16000,
    audio_format="wav",
)

assert encoded["file_name"] is None

request = GenerationRequest(columns={"audio_column": [encoded]})
arguments = AudioRequestHandler().format(request)

assert arguments.files["file"][0] is None
```

Sending the resulting `arguments.files` through HTTPX produces a multipart part without `filename=`, which compatible servers do not treat as an uploaded file.

## Operating System

Not OS-specific; observed in a Linux-hosted benchmark environment.

## Python Version

Not Python-version-specific.

## GuideLLM Version

`main` at `39383552962841086d05e25c37b58a83ef06c758`.

## Installation Method

Source checkout.

## Additional Context

A robust fix should apply defaults in both `encode_audio()` and the request handler.

fedirz:
## Summary

- honor explicit audio filenames before path-derived names
- derive in-memory filenames from the resolved audio format
- defensively provide a nonempty multipart filename in the audio request handler

## Testing

- uv run tox -e tests -- targeted filename regression tests (5 passed)
- uv run tox -e lint-check
- uv run tox -e type-check

The complete audio utility file was also attempted locally; 250 tests passed and 18 audio codec tests failed because TorchCodec could not load local FFmpeg dylibs. The new metadata-only regressions mock the codec boundary and pass.

Fixes #983

---

# git log

commit 7efdd8b65e53fdf5823cf4170696320665f1d507
Author: Fedir Zadniprovskyi <github.g1k56@simplelogin.com>
Date:   Mon Aug 3 10:42:54 2026 -0700

    fix: ensure audio uploads include filenames
    
    Signed-off-by: Fedir Zadniprovskyi <github.g1k56@simplelogin.com>

---------

Signed-off-by: Fedir Zadniprovskyi <github.g1k56@simplelogin.com>

### DIFF
--- a/src/guidellm/backends/openai/request_handlers.py
+++ b/src/guidellm/backends/openai/request_handlers.py
@@ -1324,7 +1324,7 @@ class AudioRequestHandler(ChatCompletionsRequestHandler):
 
         arguments.files = {
             "file": (
-                audio_columns[0].get("file_name", "audio_input"),
+                audio_columns[0].get("file_name") or "audio_input",
                 audio_columns[0].get("audio"),
                 audio_columns[0].get("mimetype"),
             )
--- a/src/guidellm/utils/audio.py
+++ b/src/guidellm/utils/audio.py
@@ -118,12 +118,16 @@ def encode_audio(
         mono=mono,
     )
 
+    resolved_file_name = file_name
+    if not resolved_file_name and isinstance(audio, str | Path):
+        resolved_file_name = get_file_name(audio)
+    if not resolved_file_name:
+        resolved_file_name = f"audio.{format_val}"
+
     return {
         "type": "audio_file",
         "audio": encoded_audio,
-        "file_name": get_file_name(audio)
-        if isinstance(audio, str | Path)
-        else file_name,
+        "file_name": resolved_file_name,
         "format": audio_format,
         "mimetype": f"audio/{format_val}",
         "audio_samples": samples.sample_rate,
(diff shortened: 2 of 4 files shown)


sjmonson (approved):
Sorry for the delay. I confirmed that this fixes datasets without filenames.

sjmonson:
@mergifyio queue

PR #1013, opened 2026-08-08
fix: return a Dataset from file dataset deserializers

### LINKED ISSUE #1002: Bug: AttributeError when using JSON/CSV/Parquet file datasets
## Issue Description

When running `guidellm run` with a `json_file`, `csv_file`, or `parquet_file` data source, the benchmark fails with:

```
Traceback (most recent call last):
  File ".../guidellm/data/preprocessors/mappers.py", line 192, in datasets_mappings
    if dataset.info and dataset.info.dataset_name
       ^^^^^^^^^^^^
AttributeError: 'DatasetDict' object has no attribute 'info'
```

## Root Cause

The file-based dataset deserializers (`JSONFileDatasetDeserializer`, `CSVFileDatasetDeserializer`, `ParquetFileDatasetDeserializer`) in `.venv/lib/python3.12/site-packages/guidellm/data/deserializers/file.py` are calling `load_dataset()` directly, which returns a `DatasetDict` object for these file types.

However:
1. The type annotations specify `-> Dataset` as the return type
2. Downstream code in `guidellm/data/preprocessors/mappers.py:192` expects a `Dataset` with an `.info` attribute
3. `DatasetDict` is dict-like and doesn't have an `.info` attribute - individual splits like `DatasetDict['train']` do

## Expected Behavior

The benchmark should complete successfully, processing the dataset file correctly.

## Actual Behavior

The benchmark fails during dataset preprocessing initialization with `AttributeError: 'DatasetDict' object has no attribute 'info'`.

## Fix

Modified the three deserializers to extract the 'train' split from the `DatasetDict` when it's returned:

1. Added `DatasetDict` to imports: `from datasets import Dataset, DatasetDict, load_dataset`
2. Changed each deserializer to check if the result is a `DatasetDict` and extract the train split:
   ```python
   dataset = load_dataset("json", data_files=str(path), **config.load_kwargs)
   if isinstance(dataset, DatasetDict):
       dataset = dataset["train"]
   return dataset
   ```

This ensures the deserializers return a `Dataset` object as typed, matching the behavior of other deserializers like `TextFileDatasetDeserializer`.

I've prepared a patch file which got it working on my machine: [guidellm_fix.patch](https://github.com/user-attachments/files/30828792/guidellm_fix.patch)

Maybe I find the time to open a PR, but currently I don't thinks so. Sorry 😬

Pragadeesh122:
## Summary

Running a benchmark with a `json_file`, `csv_file`, or `parquet_file` data source crashed during dataset preprocessing with:

```
AttributeError: 'DatasetDict' object has no attribute 'info'
  at guidellm/data/preprocessors/mappers.py
```

The file deserializers call `load_dataset(loader, data_files=...)`, which returns a `DatasetDict` (keyed by split), not a `Dataset` — contradicting their `-> Dataset` annotation. The column mapper then accesses `dataset.info`, which a `DatasetDict` does not have.

The same `load_dataset` pattern is used by five deserializers, so `arrow_file` and `tar_file` were affected by the same latent bug, not just the three in the report.

## Details

- Added a small `_load_file_dataset` helper in `data/deserializers/file.py` that loads the file and resolves the result to a single split via the existing `guidellm.data.utils.resolve_dataset_split`.
- Routed all five affected deserializers through it: `csv_file`, `json_file`, `parquet_file`, `arrow_file`, `tar_file`.
- Widened their return annotations from `Dataset` to `Dataset | IterableDataset` to match `resolve_dataset_split`'s return type.
- Reuses the project's existing split-resolution utility rather than adding new logic. `text_file`, `hdf5_file`, and `db_file` already returned a `Dataset` and are unchanged.

## Test Plan

- Updated `tests/unit/data/deserializers/test_file.py`: the csv/json/parquet/arrow/tar success tests previously asserted a `DatasetDict` return (which had encoded the buggy behavior) and now assert a `Dataset`.
- Added a regression test asserting the deserialized dataset exposes `.info` — the attribute whose absence caused the reported crash.
- `tox -e lint-check`, `tox -e type-check`, and `tox -e test-unit` all pass (2848 passed).
- Manually drove the reported crash path end to end (deserialize a `.jsonl` file → `GenerativeColumnMapper.setup_data` → `datasets_mappings`); it completes and maps `prompt` to `text_column` instead of raising.

## Related Issues

- Fixes #1002

Thanks to @Basster for the detailed report and root-cause analysis.

---

# git log

commit e3015d1ee6cdf345737919288f5c355028cdeb75
Date:   Fri Aug 7 21:24:17 2026 -0500

    fix: return a Dataset from file deserializers (#1002)
    
    load_dataset returns a DatasetDict for csv/json/parquet/arrow/tar file
    sources, which crashed column mapping with 'DatasetDict object has no
    attribute info'. Resolve the loaded result to a single dataset split via
    resolve_dataset_split so the deserializers return a Dataset as annotated.
    
---------

### DIFF
--- a/src/guidellm/data/deserializers/file.py
+++ b/src/guidellm/data/deserializers/file.py
@@ -2,10 +2,10 @@ from __future__ import annotations
 
 from collections.abc import Callable
 from pathlib import Path
-from typing import Literal
+from typing import Any, Literal
 
 import pandas as pd
-from datasets import Dataset, load_dataset
+from datasets import Dataset, IterableDataset, load_dataset
 from pydantic import Field
 from transformers import PreTrainedTokenizerBase
 
@@ -15,6 +15,7 @@ from guidellm.data.deserializers.deserializer import (
     DatasetDeserializerFactory,
 )
 from guidellm.data.schemas import DataArgs
+from guidellm.data.utils import resolve_dataset_split
 
 __all__ = [
     "ArrowFileDatasetDeserializer",
@@ -27,6 +28,25 @@ __all__ = [
     "TarFileDatasetDeserializer",
     "TextFileDatasetDeserializer",
 ]
+
+
+def _load_file_dataset(
+    loader: str, path: Path, **load_kwargs: Any
+) -> Dataset | IterableDataset:
+    """
+    Load a local data file and return a single dataset split.
+
+    ``load_dataset`` returns a ``DatasetDict`` keyed by split for file-based
+    loaders, which downstream preprocessing does not expect. This resolves the
+    result to a single ``Dataset`` (or ``IterableDataset``) split.
+
+    :param loader: The ``datasets`` loader name (e.g. ``"json"``, ``"csv"``).
+    :param path: Path to the local data file.
+    :param load_kwargs: Additional keyword arguments forwarded to ``load_dataset``.
+    :return: The resolved dataset split.
+    """
+    dataset = load_dataset(loader, data_files=str(path), **load_kwargs)
+    return resolve_dataset_split(dataset)
 
 
 @DataArgs.register(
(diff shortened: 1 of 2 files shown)


sjmonson (approved):
Seems like a reasonable fix, thx.

PR #1026, opened 2026-08-13
fix: use spawn multiprocessing context on macOS

### LINKED ISSUE #1022: Default `mp_context_type="fork"` segfaults all workers on macOS (supported platform); actionable hint was removed from the error message
## Summary

On macOS, `guidellm run` fails immediately with every worker dying on signal 11. The cause is that `mp_context_type` defaults to `"fork"` unconditionally, overriding CPython's macOS default of `spawn`. The documented fix (`GUIDELLM__MP_CONTEXT_TYPE=spawn`) works perfectly — but the error message no longer points at it, so the failure looks like a hard crash rather than a one-variable configuration issue.

macOS is listed as a supported platform (`README.md`, `docs/getting-started/install.md`: *"Operating System: Linux or MacOS"*), so filing this as a defect rather than a platform limitation.

I could not find an existing issue that reports this. #559 is related but proposes a *global* switch to `forkserver` and does not mention macOS; it would not obviously close this.

## Environment

| | |
|---|---|
| guidellm | 0.7.3 |
| OS | macOS 26.5.2 (build 25F84) |
| Chip | Apple M4 Pro (arm64) |
| Python | 3.13.5 |
| platform | `macOS-26.5.2-arm64-arm-64bit-Mach-O` |
| torch / transformers / sanic | 2.13.0 / 5.15.0 / 25.12.1 |

Installed into a clean isolated environment via `uv run --no-project --with "guidellm==0.7.3"`.

## Reproduction

Terminal 1 — mock server (this part works fine; it doesn't use the worker group):

```bash
guidellm mock-server --host 127.0.0.1 --port 8012 --model gpt2 \
  --workers 4 --ttft-ms 5 --itl-ms 1 --output-tokens 16
```

Terminal 2 — benchmark against it:

```bash
guidellm run \
  --backend "kind=openai_http,target=http://127.0.0.1:8012" \
  --data "kind=synthetic_text,prompt_tokens=32,output_tokens=16" \
  --profile "kind=constant,rate=8" \
  --constraint "kind=max_duration,seconds=5" \
  --output "kind=json"
```

## Actual behaviour

Exit code 1, no output file produced. All ten workers die simultaneously:

```
File ".../guidellm/scheduler/worker_group.py", line 273, in create_processes
    raise RuntimeError(f"Worker process group startup failed: {detail}")
RuntimeError: Worker process group startup failed: Worker process 525 died unexpectedly
(signal 11); Worker process 526 died unexpectedly (signal 11); Worker process 527 died
unexpectedly (signal 11); ... Check system logs for details
```

## Expected behaviour

Either the benchmark runs, or the error names the setting that fixes it.

## Root cause

`src/guidellm/settings.py:93` sets the default with no platform branch:

```python
mp_context_type: Literal["spawn", "fork", "forkserver"] | None = "fork"
```

and `src/guidellm/scheduler/worker_group.py:188` consumes it directly:

```python
self.mp_context = get_context(settings.mp_context_type)
```

CPython changed the macOS default to `spawn` in 3.8 because `fork()` is unsafe there (Objective-C runtime initialisation across `fork()` without `exec()`); this default reverts that.

## Confirmed fix

```bash
GUIDELLM__MP_CONTEXT_TYPE=spawn guidellm run ...   # exit 0, benchmark completes, JSON written
```

Verified on the exact repro above — same command, same machine, only the env var added.

## The part that makes this hard to diagnose

The fix *is* documented, in `docs/guides/troubleshooting.md` under "macOS worker crash (signal 11)" (added in #860, thank you). But commit `06707a1d` ("Removed suggested action in failed sub-process message") stripped the inline hint out of the runtime error, so on 0.7.x the message ends at `"Check system logs for details"`.

The result is that the one place a user is guaranteed to be looking — the traceback — no longer mentions the variable that fixes it. Ten simultaneous SIGSEGVs read as a serious native crash, not a config default, so the natural next step is to go debugging rather than to go read the troubleshooting guide.

## Suggested fixes, in increasing order of ambition

1. **Restore the hint in the error.** Smallest possible change; recovers most of the lost diagnosability. Something like: `... Check system logs for details. On macOS, try GUIDELLM__MP_CONTEXT_TYPE=spawn (see docs/guides/troubleshooting.md).`
2. **Make the default platform-conditional** — `"fork"` on Linux, `"spawn"` on Darwin. Keeps the measured `fork` startup advantage on the platform where it's safe, and makes the supported platform work out of the box. (Distinct from #559, which proposes `forkserver` everywhere; both could land.)
3. **Warn at startup** when `mp_context_type == "fork"` on a non-Linux platform.

Happy to open a PR for (1) or (2) if you'd like — just say which you'd prefer.

nightcityblade:
## Summary

Use the `spawn` multiprocessing context on macOS so worker startup does not
override CPython's platform-safe default with `fork`. Signal-related startup
errors now also point users to the existing troubleshooting guide.

## Details

- [x] Force `spawn` when `sys.platform == "darwin"` while preserving the configured context elsewhere.
- [x] Restore an actionable troubleshooting reference for signal-killed workers.
- [x] Add regression coverage for macOS context selection.

## Test Plan

- `tox -e test-unit` — full unit invocation completed successfully (3,053 collected).
- `tox -e test-unit -- -k macos_uses_spawn_context` — 1 passed, 3,052 deselected.
- `tox -e lint-check` — Ruff format/check and mdformat passed.
- `tox -e type-check` — Mypy passed for 162 source files.
- `git diff --check` — passed.

## Related Issues

- Resolves #1022

---

## Use of AI

The commit includes the required `Generated-by: OpenAI Codex GPT-5` trailer,
and the generated regression test includes the marker required by `AGENTS.md`.

---

# git log

commit 7b8941efa0966a862ac4ecc15d9198d97cdad503
Author: nightcityblade <nightcityblade@gmail.com>
Date:   Thu Aug 13 23:29:10 2026 +0800

    fix: use spawn multiprocessing context on macOS
    
    Generated-by: OpenAI Codex GPT-5
    Signed-off-by: nightcityblade <nightcityblade@gmail.com>

---------

Generated-by: OpenAI Codex GPT-5
Signed-off-by: nightcityblade <nightcityblade@gmail.com>

### DIFF
--- a/src/guidellm/scheduler/worker_group.py
+++ b/src/guidellm/scheduler/worker_group.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 import asyncio
 import contextlib
 import math
+import sys
 import threading
 import time
 import uuid
@@ -185,7 +186,9 @@ class WorkerProcessGroup(Generic[RequestT, ResponseT]):
         per_proc_max_buffer_size = 1
 
         # Initialize multiprocessing components
-        self.mp_context = get_context(settings.mp_context_type)
+        self.mp_context = get_context(
+            "spawn" if sys.platform == "darwin" else settings.mp_context_type
+        )
         self.mp_manager = self.mp_context.Manager()
         self.startup_barrier = self.mp_context.Barrier(num_processes + 1)
         self.requests_generated_event = self.mp_context.Event()
@@ -299,7 +302,10 @@ class WorkerProcessGroup(Generic[RequestT, ResponseT]):
             if dead:
                 message = "; ".join(dead)
                 if killed_by_signal:
-                    message += ". Check system logs for details"
+                    message += (
+                        ". Check system logs and "
+                        "docs/guides/troubleshooting.md for details"
+                    )
                 self._worker_error_details = message
                 if self.error_event is not None:
                     self.error_event.set()
--- a/tests/unit/scheduler/test_worker_group.py
+++ b/tests/unit/scheduler/test_worker_group.py
@@ -105,6 +105,22 @@ class TestWorkerProcessGroup:
     def teardown_method(self):
         pass
 
+    @pytest.mark.regression
+    @pytest.mark.asyncio
+    @patch("guidellm.scheduler.worker_group.sys.platform", "darwin")
+    @patch("guidellm.scheduler.worker_group.get_context")
+    async def test_macos_uses_spawn_context(self, mock_get_context):
+        """Use the safe multiprocessing context on macOS. ## WRITTEN BY AI ##"""
+        mock_get_context.side_effect = RuntimeError("context selected")
+        instance = WorkerProcessGroup(
+            requests=["request"], backend=MockBackend(), strategy=SynchronousStrategy()
+        )
+
+        with pytest.raises(RuntimeError, match="context selected"):
+            await instance.create_processes()
+
+        mock_get_context.assert_called_once_with("spawn")
+
     @pytest.fixture(
         params=[
             {


sjmonson (approved):
LGTM, would like a test from @jaredoconnell who is a Mac user.

jaredoconnell (approved):
I can confirm that this works. I agree that this is a better solution than just exclusively documenting it in the troubleshooting guide.

PR #1031, opened 2026-08-16
docs: add Kubernetes/OpenShift client Job example

amit-chaubey:
## Summary

Adds a docs example for running GuideLLM as a **client-only** batch Job against an
existing OpenAI-compatible Service (for example an in-cluster vLLM deployment). The
Job requires no GPU, is configured entirely through the `GUIDELLM__SPEC__*`
environment variables already documented for the container image, and persists
`benchmarks.{json,csv,html}` to a PVC.

`docs/examples/index.md` currently states that few specific examples are available
and invites contributions; there is no Kubernetes or OpenShift guidance anywhere in
the repo today. Running the benchmark client in-cluster is a common deployment shape,
and the non-obvious parts — why a `Job` rather than a `Deployment`, and how the
image's `1001:0` / group-writable-`HOME` layout interacts with OpenShift's SCC-assigned
UID and fsGroup ranges — are exactly what this example documents.

## Details

- [ ] Add `docs/examples/kubernetes-openshift-job.md`: a complete PVC + Job manifest
      with inline `CHANGE ME` annotations on the knobs operators retune
- [ ] Explain the choice of `Job` over `Deployment` (`backoffLimit: 0`,
      `restartPolicy: Never`, `activeDeadlineSeconds`, `ttlSecondsAfterFinished`)
- [ ] Document the security posture: `runAsNonRoot`, `drop: [ALL]`,
      `readOnlyRootFilesystem` with explicit writable mounts, `RuntimeDefault`
      seccomp, and `automountServiceAccountToken: false`
- [ ] Add OpenShift notes covering SCC-assigned `runAsUser`/`fsGroup`, why an explicit
      `fsGroup: 0` is rejected by `restricted-v2`, and disabling Istio sidecar
      injection so the mesh proxy does not distort latency measurements
- [ ] Add an image pinning / upgrade section (release tag vs. digest, avoiding
      `:latest` and `:stable` so results stay comparable across runs)
- [ ] Add a customization cheat sheet mapping each `GUIDELLM__SPEC__*` variable to
      its typical use
- [ ] Promote the example list in `docs/examples/index.md` into its own
      `## Available Examples` section above the call for contributions, and link all
      three examples with short descriptions

## Test Plan

- `tox -e lint-check` passes (`mdformat --check` over `docs/`)
- Both YAML documents in the example parse cleanly with `yaml.safe_load_all`
- Every `GUIDELLM__SPEC__*` value in the manifest was validated against the real
  schemas via `BenchmarkScenario.model_validate`, confirming `backend`, `profile`
  (`streams: 4` coerces to `[4]`), both constraints, `data`, and all three outputs
  resolve as intended to `/results/*`
- Image tag, `ENTRYPOINT`/`CMD`, `HOME`, `GUIDELLM__DEFAULT_RESULTS_DIR`, and the
  `1001:0` user layout were cross-checked against `Containerfile`; the multi-arch
  claim was verified against the release tags where
  `build-multiarch-container.yml` was introduced

## Related Issues

- Resolves #

---

## Use of AI

---

# git log

commit 5cd2cece0feb6ed8de00f48d2461a423b5e40e42
Author: AmitChaubey <amit.katyayana@gmail.com>
Date:   Sun Aug 16 16:31:58 2026 +0100

    docs: add Kubernetes/OpenShift client Job example
    
    Add a docs example for running GuideLLM as a client-only, non-GPU batch
    Job against an in-cluster OpenAI-compatible Service, with benchmark
    artifacts persisted to a PVC.
    
    The manifest is configured entirely through the GUIDELLM__SPEC__*
    environment variables already documented for the container image, and is
    written to pass restricted Pod Security Standards and the OpenShift
    restricted-v2 SCC. Also link the example from the examples index and
    promote the example list above the call-for-contributions section.
    
    Signed-off-by: AmitChaubey <amit.katyayana@gmail.com>

---------

Signed-off-by: AmitChaubey <amit.katyayana@gmail.com>

### DIFF
--- a/docs/examples/kubernetes-openshift-job.md
+++ b/docs/examples/kubernetes-openshift-job.md
@@ -1,0 +1,259 @@
+# Kubernetes / OpenShift client Job
+
+Run GuideLLM as a **client-only** batch Job against an existing OpenAI-compatible Service (for example in-cluster vLLM). The Job does **not** need a GPU.
+
+This example uses the published GHCR image. Static benchmark settings are passed as CLI arguments in `args`; only the values you are likely to source from a ConfigMap or Secret — the endpoint URL and API key — are environment variables. Results persist on a PVC.
+
+## Why a Job (not a Deployment)
+
+A benchmark is run-to-completion batch work. A Deployment would restart the pod after the run finishes and re-execute the benchmark indefinitely, producing duplicate result files and load you did not ask for. `Job` with `restartPolicy: Never` and `backoffLimit: 0` guarantees exactly one attempt; `activeDeadlineSeconds` caps a hung run (for example when the serve restarts mid-benchmark).
+
+## Prerequisites
+
+- A reachable OpenAI-compatible endpoint (Service DNS or URL)
+- Cluster permission to create a `Job` and a `PersistentVolumeClaim` in your namespace
+- Image pull access to `ghcr.io/vllm-project/guidellm`
+
+Pin a release tag (`vX.Y.Z`). Prefer that over `:latest` in production.
+
+## Quick start
+
+1. Copy the YAML below into a file (for example `guidellm-job.yaml`).
+2. Replace `REPLACE_WITH_OPENAI_BASE_URL` (example: `http://vllm.my-ns.svc.cluster.local:8000`).
+3. Adjust profile / data / constraints / sample count as needed. Knobs marked `CHANGE ME` in comments are the ones operators typically retune.
+4. Apply:
+
+```bash
+kubectl apply -f guidellm-job.yaml
+# OpenShift:
+oc apply -f guidellm-job.yaml
+```
+
+5. Follow logs. Artifacts land on the PVC at `/results` (`benchmarks.json`, `benchmarks.csv`, `benchmarks.html`):
+
+```bash
+kubectl logs -f job/guidellm-openai-bench
+kubectl get pvc guidellm-results
+```
+
+```yaml
+# Client-only GuideLLM benchmark Job against an OpenAI-compatible Service.
+#
+# Why a Job (not a Deployment): a benchmark is run-to-completion batch work.
+# A Deployment would restart the pod and re-run the benchmark indefinitely.
+#
+# Edit REPLACE_* placeholders before apply.
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: guidellm-results
+  labels:
+    app.kubernetes.io/name: guidellm
+    app.kubernetes.io/component: results
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      # CHANGE ME: size for your result retention needs (json/csv/html per run).
+      storage: 5Gi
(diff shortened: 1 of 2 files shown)


docs/examples/kubernetes-openshift-job.md
```diff
@@ -0,0 +1,232 @@
...
+            # synthetic_text, which generates exactly what the run consumes.
+            # - name: GUIDELLM__SPEC__DATA_LOADER
+            #   value: >-
+            #     {"kind":"pytorch","samples":100}
+            # The image already defaults results to /results, so these paths only
+            # need to be explicit because html is not one of the default outputs
+            # (defaults are json + csv).
+            - name: GUIDELLM__SPEC__OUTPUTS
+              value: >-
+                [{"kind":"json","path":"/results/benchmarks.json"},{"kind":"csv","path":"/results/benchmarks.csv"},{"kind":"html","path":"/results/benchmarks.html"}]
```
sjmonson: Just an fyi, CLI environment variables are nest-able; instead of having `GUIDELLM__SPEC__PROFILE='{"kind":"concurrent","streams":4}'` you could have `GUIDELLM__SPEC__PROFILE__KIND=concurrent` and `GUIDELLM__SPEC__PROFILE__STREAMS=4` which might make more sense for some of the variables you set.

Its up to personal preference, but generally my recommendation is to only use variables when you might need to read the value from somewhere else (like a Secret or CM). So for example:

```yaml
args:
    - run
    - --backend
    - kind=openai_http
    - --profile 
    - kind=concurrent,streams=4
    - --constraint
    - kind=max_requests,count=100
    - --constraint
    - kind=max_duration,seconds=600
    - --data
    - kind=synthetic_text,prompt_tokens=256,output_tokens=128
    - --output
    - kind=json,path=/results/benchmarks.json
    - --output
    - kind=csv,path=/results/benchmarks.csv
env:
    # Could be from CM
    - name: GUIDELLM__SPEC__BACKEND__TARGET
      value: "http://example.com"
    # Could be from secret
    - name: GUIDELLM__SPEC__BACKEND__API_KEY
      value: "xxxxxxxxxxxxx"
```

docs/examples/kubernetes-openshift-job.md
```diff
@@ -0,0 +1,232 @@
...
+          resources:
+            # CHANGE ME: the client is CPU/memory bound only; never request GPUs here.
+            requests:
+              cpu: "500m"
+              memory: 1Gi
+            limits:
+              cpu: "2"
+              memory: 4Gi
```
sjmonson, line 173: Can you make requests equal to the limits so by default this example gets CPU isolation.

docs/examples/kubernetes-openshift-job.md
amit-chaubey: Good Call!

docs/examples/kubernetes-openshift-job.md
amit-chaubey: Done!

sjmonson (approved):

amit-chaubey:
thanks @sjmonson

PR #1032, opened 2026-08-17
Minimal Self-Contained HTML Report

### LINKED ISSUE #584: Replace of current HTML based UI

The current frontend is not widely used and maintenance is required.

Find a much simpler frontend solution for those who really need it

Related to #579

jaredoconnell:
## Summary

This PR replaced the library-heavy existing HTML report with a much simpler one, and includes a lot more useful displays. 

Attached are three scenarios exported as reports. Note that they were just ran on my laptop with GuideLLM's mock server. There is a single concurrency run, a multi-concurrency one, a multi-turn one, and a default sweep run.

image

## Examples

Updated:

[single.html](https://github.com/user-attachments/files/31229605/single.html)
[multi4.html](https://github.com/user-attachments/files/31229608/multi4.html)
[turns1_8.html](https://github.com/user-attachments/files/31229610/turns1_8.html)
[sweep.html](https://github.com/user-attachments/files/31229633/sweep.html)

Old:

[multi4.html](https://github.com/user-attachments/files/31158761/multi4.html)
[turns1_8.html](https://github.com/user-attachments/files/31158762/turns1_8.html)
[single.html](https://github.com/user-attachments/files/31158760/single.html)

## Details

- Has no dependencies. Everything is self-contained in the HTML. This fixes problems where the old report depended on improperly versioned hosted dependencies.
- Does not use any JS libraries, and is vanilla JS+CSS+HTML.
- Is designed as a template with data inserted to form a single file.
- Includes a tab for multiturn data broken down by turn, if present.
- Displays request stats, latency stats, throughput stats, and more.
- Designed for less technical users. Includes help bubbles alongside most metrics and sections.
- Has slightly different displays for single runs vs multi runs depending on what conveys the info better. Bar graphs were unhelpful with single bars.
- The report is responsive.

## Test Plan

- You can run benchmarks or use the `guidellm export` with `--output kind=html,path=filename.html`

The focus of the review should probably be on how the HTML report interfaces with the python code. The HTML, CSS, and JS are all AI generated, and are held to a lower quality standard than GuideLLM's code.

## Related Issues

- Resolves #584  

---

## Use of AI

---

# git log

commit 7f39128f8d7690dde95c6851061b51618b6d0cec
Date:   Fri Aug 14 22:54:20 2026 -0400

    Replaced HTML report with new self-contained HTML report
    
    Generated-by: Cursor AI Grok 4.5

commit c33120ab5d84d80b98c267b5a2ba85e653d9a40a
Date:   Mon Aug 17 18:54:49 2026 -0400

    Improvements to HTML report
    
    Generated-by: Cursor AI Grok 4.5

commit 4f1967e25aca68d6d36e103cf8ac7352ec7fb63e
Date:   Mon Aug 17 19:03:11 2026 -0400

    Remove unnecessary code, and fix linter errors
    
    Assisted-by: Cursor AI Grok 4.5

commit b362cafecd54a76c27e90709c3aeba1d00969692
Date:   Wed Aug 19 02:17:21 2026 -0400

    Address review feedback
    
    Switch to integer graph labels
    Use target rather than measured concurrency
    Added measured concurrency details to efficiency graph
    Fix log-scale
    
    Generated-by: Cursor AI Grok 4.5

commit 02df9ac609f118480606676b2049e2f6a3532513
Date:   Wed Aug 19 11:23:11 2026 -0400

    Fix missing model, and use strongly typed refs
    
    Generated-by: Cursor AI Grok 4.5

commit 58970fa3a64c9d0ddf547805fc98e9c6674ca721
Date:   Thu Aug 20 11:00:59 2026 -0400

    Updated help messages and reduced redundancy
    
    Assisted-by: Cursor AI Grok 4.5

commit 3c1663552d3194b4488096196670d7749b6912ef
Date:   Thu Aug 20 20:36:28 2026 -0400

    Address review feedback
    
    Add graph bar padding
    Removed Log option
    Updated metric selection
    Renamed graphs to use "by benchmark" wording
    Added alternative metric for non-concurrency-based benchmarks for efficiency graph
    
    Generated-by: Cursor AI Grok 4.5

commit b446322d99880f7ca2dcd1fd2a7883b98a1f89e5
Date:   Wed Aug 26 00:09:45 2026 -0400

    Address review comments
    
    Fixes usability concerns in the HTML report
    Adds basic JS test
    
    Generated-by: Cursor AI Grok 4.5

commit 7fc7e29273e8dcf5782ae1b146746a03f9de7f16
Date:   Wed Aug 26 11:30:16 2026 -0400

    Improve multiturn view with subagents
    
    Generated-by: Cursor AI Grok 4.5

---------

Assisted-by: Cursor AI Grok 4.5
Generated-by: Cursor AI Grok 4.5


sjmonson (changes requested):
Still working on a full code review but here are a few visual things I noticed:

- Multi-run results have a Linear/Log scale toggle on the performance tab which does not work.
- The `concurrent@<streams>` is using what I assume is mean `request_concurrency` for streams rather then the user defined streams. I also assume that RPS benchmarks would have the same problem using response rate rather then the user defined rate. Ideally this string is taken from the scheduling strategy rather then handled in the output.
- The "Latency & prompt size by turn" graph should use an integer axis label for the turns axis.

src/guidellm/benchmark/outputs/html_report/report.js
```diff
@@ -0,0 +1,2875 @@
...
+      lines: [
+        "End-to-end (E2E) latency: full request time from start to finished response.",
+        "Same idea as “request latency.”",
+      ],
+      also: ["p95", "p99"],
+    },
+    ttft: {
+      title: "TTFT",
+      lines: [
+        "Time to First Token (TTFT): wait until the first piece of the reply appears.",
```
dbutenhof: I don't like "wait" here; the term is ambiguous, easy to read as instructions (even though that makes no sense here) rather than description. How about "time to produce the first response token"? Or "delay before the first token appears"???

src/guidellm/benchmark/outputs/html_report/report.js
```diff
@@ -0,0 +1,2875 @@
...
+    errored: {
+      title: "Errored",
+      lines: [
+        "Requests that failed with an error (bad request, server fault, or invalid response).",
+      ],
+    },
+    lat_components: {
+      title: "Latency components",
+      lines: [
+        "TTFT: wait for the first token.",
```
dbutenhof: Again, I think something like "time to produce the first token" would be more clear than "wait".

dbutenhof (commented):
I haven't tried to review the code in detail -- it's not as important in this case, though I'll try to glance through later.

I did pull the branch and experiment a bit, and I like the HTML format. Much smaller than I'd feared, too.

I did note a couple of phrasings that felt awkward in the tool tips; but today I was just trying to get my toes damp again, and I might have more later...

src/guidellm/benchmark/outputs/html_report/report.js
jaredoconnell: I agree that it was odd. I glanced over that. Fixed.

src/guidellm/benchmark/outputs/html_report/report.js
jaredoconnell: Fixed.

dbutenhof (dismissed):

sjmonson (changes requested):
I ran some of my own benchmarks and have a few more nits:

### 1. The longest bar in each graph is always extended as far to the right as possible.

Maybe add some padding so that the bar fits nicer into the graph space

Image

Image

### 2. For a single benchmark run "Per-request efficiency" contains TPOT P95 and ITL P95 but not TTFT

I would replace TPOT with TTFT, as TPOT is not really used anymore since TTFT+ITL quantify workloads better.

Image

### 3. Many graphs are titled "...by concurrency" even when the data isn't concurrency based.

I think it should be "...by benchmark" as the terminology we normally use is that a single rate/stream is a "benchmark" and the whole run though all rates/streams is a "run".

Image

### 4. The throughput efficiency graph doesn't make sense for non-concurrent

See the above example. Not sure the solution here yet, just pointing it out in-case you have ideas.

jaredoconnell:
> I ran some of my own benchmarks and have a few more nits:
> ### 1. The longest bar in each graph is always extended as far to the right as possible.
> 
> Maybe add some padding so that the bar fits nicer into the graph space
> ### 2. For a single benchmark run "Per-request efficiency" contains TPOT P95 and ITL P95 but not TTFT
> 
> I would replace TPOT with TTFT, as TPOT is not really used anymore since TTFT+ITL quantify workloads better.
> ### 3. Many graphs are titled "...by concurrency" even when the data isn't concurrency based.
> 
> I think it should be "...by benchmark" as the terminology we normally use is that a single rate/stream is a "benchmark" and the whole run though all rates/streams is a "run".
> ### 4. The throughput efficiency graph doesn't make sense for non-concurrent
> 
> See the above example. Not sure the solution here yet, just pointing it out in-case you have ideas.

All of these problems should now be addressed. Attached are updated HTML outputs.

[turns1_8.html](https://github.com/user-attachments/files/31285677/turns1_8.html)
[constant.html](https://github.com/user-attachments/files/31285678/constant.html)
[sweep.html](https://github.com/user-attachments/files/31285679/sweep.html)
[multi4.html](https://github.com/user-attachments/files/31285680/multi4.html)
[single.html](https://github.com/user-attachments/files/31285682/single.html)

sjmonson:
The linear/log toggle is gone. Was that intentional?

dbutenhof (commented):
On a whim, I ran a Cursor code review. A few of its comments that may make sense:

1. (In JavaScript) Dead code. niceNum, vertical drawStackedBars / drawGroupedBars, and HELP.prompt_vs_turn are unused. Python still emits top-level kpis and per-run modalities; JS reads KPIs from runs[selected] and ignores data.kpis. use_history_axis is set in Python and never consumed.
2. runChartLabel strips the strategy name at “@” so sweep charts show “@2.00”, “@4”. Full labels remain in tables and the KPI select.
Image
 
3. No runtime tests for the UI. Seven Python tests check the compact dict and that the HTML string contains glossary fragments and no CDN URLs. Nothing executes report.js. 
4. Multi-agent turns are silently filtered. Dominant agent_id is not in the payload or turn_note.
5. Turn P95 ≠ run P95. Nearest-rank vs interpolated DistributionSummary; n=1 makes P95 and P99 the same sample.
6. Weak accessibility. No tab semantics; help aria-expanded is stuck; most SVGs have no accessible name.
7. JS-only document. Locked-down or no-JS viewers see an empty briefing even though the JSON is in the file.
8. Outcome bars are counts, not mix. A longer duration-bounded run looks “greener” at the same error rate. The single-run pie already shows share.
9. Low: unused drawers and payload fields; SVG labels capped at 88px but not truncated; pie vs donut inconsistency; Red Hat fonts named but not embedded; str.replace templating is order-sensitive; ITL on the turn chart uses the TPOT orange.

Suggested merge bar
1. Fix the KPI-picker sentence (or stop redrawing the breakdown chart).
2. Put the selected agent_id (and request share) on turn_note.
3. Add one DOM/jsdom test: multi-run HTML → KPI values, comparison table rows, SVG present after switching to Performance.
4. Align turn percentiles with DistributionSummary, or document the difference.

A few other comments:
1. Tabs are plain buttons, not a tablist. (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tablist_role)
2. Layout still uses a few inline style= attributes that belong in CSS.

Once I looked for it, I found that dropping the sweep "constant" strategy name was a bit grating. Not sure it's actually a show-stopper, but some of the comments are worth consideration.

jaredoconnell:
> On a whim, I ran a Cursor code review. A few of its comments that may make sense:
> 
>   1. (In JavaScript) Dead code. niceNum, vertical drawStackedBars / drawGroupedBars, and HELP.prompt_vs_turn are unused. Python still emits top-level kpis and per-run modalities; JS reads KPIs from runs[selected] and ignores data.kpis. use_history_axis is set in Python and never consumed.
> 
>   2. runChartLabel strips the strategy name at “@” so sweep charts show “@2.00”, “@4”. Full labels remain in tables and the KPI select.
> 
> 
>   3. No runtime tests for the UI. Seven Python tests check the compact dict and that the HTML string contains glossary fragments and no CDN URLs. Nothing executes report.js.
> 
>   4. Multi-agent turns are silently filtered. Dominant agent_id is not in the payload or turn_note.
> 
>   5. Turn P95 ≠ run P95. Nearest-rank vs interpolated DistributionSummary; n=1 makes P95 and P99 the same sample.
> 
>   6. Weak accessibility. No tab semantics; help aria-expanded is stuck; most SVGs have no accessible name.
> 
>   7. JS-only document. Locked-down or no-JS viewers see an empty briefing even though the JSON is in the file.
> 
>   8. Outcome bars are counts, not mix. A longer duration-bounded run looks “greener” at the same error rate. The single-run pie already shows share.
> 
>   9. Low: unused drawers and payload fields; SVG labels capped at 88px but not truncated; pie vs donut inconsistency; Red Hat fonts named but not embedded; str.replace templating is order-sensitive; ITL on the turn chart uses the TPOT orange.
> 
> 
> Suggested merge bar
> 
>   1. Fix the KPI-picker sentence (or stop redrawing the breakdown chart).
> 
>   2. Put the selected agent_id (and request share) on turn_note.
> 
>   3. Add one DOM/jsdom test: multi-run HTML → KPI values, comparison table rows, SVG present after switching to Performance.
> 
>   4. Align turn percentiles with DistributionSummary, or document the difference.
> 
> 
> A few other comments:
> 
>   1. Tabs are plain buttons, not a tablist. (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tablist_role)
> 
>   2. Layout still uses a few inline style= attributes that belong in CSS.
> 
> 
> Once I looked for it, I found that dropping the sweep "constant" strategy name was a bit grating. Not sure it's actually a show-stopper, but some of the comments are worth consideration.

This should be addressed now.

> The linear/log toggle is gone. Was that intentional?

Yes, this was intentional. I didn't find any value in it for the dynamic range of the data we're working with here, and it added complexity.

Attached are updated reports.

[constant.html](https://github.com/user-attachments/files/31448373/constant.html)
[turns1_8.html](https://github.com/user-attachments/files/31448374/turns1_8.html)
[sweep.html](https://github.com/user-attachments/files/31448375/sweep.html)
[multi4.html](https://github.com/user-attachments/files/31448376/multi4.html)
[single.html](https://github.com/user-attachments/files/31448377/single.html)

src/guidellm/benchmark/outputs/html.py
```diff
@@ -155,92 +182,220 @@ def build_report_view(report: Any) -> dict[str, Any]:
...
+            note_parts: list[str] = []
+            # Only mention the agent when more than one was present (filtering applied).
+            if turn_bundle["agent_count"] > 1:
+                agent_label = turn_bundle["agent_id"]
+                agent_display = "default" if agent_label is None else str(agent_label)
+                note_parts.append(
+                    f'Turn curves use agent "{agent_display}" '
+                    f"({turn_bundle['agent_request_count']}/"
+                    f"{turn_bundle['total_successful']} requests)."
+                )
```
dbutenhof: Interesting choice. I'm not sure how I feel about it. The HTML view doesn't need to be "everything for everyone", but I can also easily imagine a pulldown to switch between agent conversations analogous to your strategy pulldown.

dbutenhof (dismissed):
The output looks good. The code seems a bit "scattered", but generally seems reasonably clear. I think a lot of it could be improved, but nothing says this needs to be "all for everyone" or "everything all at once", so I'll say "good enough". (I'll resolve my one comment regardless of any response, but I'm curious...)

sjmonson (dismissed):
Lets go!

src/guidellm/benchmark/outputs/html.py
jaredoconnell: I improved it more 🙂

jaredoconnell:
[branches_moderate.html](https://github.com/user-attachments/files/31478593/branches_moderate.html)
Made some more improvements to the subagent multiturn by-turn view.

dbutenhof (approved):

sjmonson (approved):

sjmonson:
@mergifyio queue

PR #1044, opened 2026-08-24
Report worker 0 correctly in RequestInfo.scheduler_node_id

### LINKED ISSUE #1043: [Bug]: worker 0 is always reported as scheduler_node_id -1
### Bug Description

`WorkerProcess` reports which worker handled a request through `RequestInfo.scheduler_node_id`, but worker **0** is always reported as **-1**.

The three assignments in `src/guidellm/scheduler/worker.py` use

```python
request_info.scheduler_node_id = self.messaging.worker_index or -1
```

(`_cancel_requests_loop` twice, `_prepare_node` once). `or -1` is meant as a fallback for `worker_index is None`, but `0` is falsy too, so the first worker's index is replaced by the "unknown" sentinel.

Effect: every request handled by worker 0 shows `"scheduler_node_id": -1` in the benchmark JSON. This is the field you need when debugging per-worker scheduling behaviour (I hit it while tracing #1041: worker 0's requests looked "unassigned", and per-worker limits computed from the output were wrong for it).

### Expected Behavior

Requests handled by worker 0 have `scheduler_node_id == 0`; `-1` only when no worker index is known.

### Steps to Reproduce

Any multi-process run, e.g. against `guidellm mock-server`:

```bash
guidellm mock-server --port 8010 --model mock-model
guidellm run \
  --backend "kind=openai_http,target=http://127.0.0.1:8010,model=mock-model" \
  --data "kind=synthetic_text,prompt_tokens=64,output_tokens=32" \
  --tokenizer "kind=hf_auto,model=gpt2" \
  --profile "kind=throughput,max_concurrency=20" \
  --constraint "kind=max_requests,count=20" \
  --output "kind=json,path=out.json"
python -c "import json,collections; d=json.load(open('out.json')); print(collections.Counter(r['info']['scheduler_node_id'] for r in d['benchmarks'][0]['requests']['successful']))"
```

With 10 worker processes the counter shows ids `-1, 1, 2, …, 9` and never `0`.

### Proposed fix

`WorkerProcess` already stores its own `self.worker_index: int` (used for `strategy.next_request_time(worker_index=self.worker_index)`), so the three sites can simply use that. I have a one-line-per-site PR ready with a regression test.

himanshu1573:
## Summary

Every request handled by worker **0** is reported with `scheduler_node_id == -1` instead of `0` (#1043).

The three assignments in `src/guidellm/scheduler/worker.py` use `self.messaging.worker_index or -1`. The `or -1` is meant as a fallback for `worker_index is None`, but `0` is falsy too, so the first worker's index is replaced by the "unknown" sentinel. This corrupts the per-worker view of the benchmark output that you need when debugging scheduling behaviour (I hit it while working on #1041 / #1042).

## Details

- `_prepare_node`, and both loops in `_cancel_requests_loop`, now use the worker's own `self.worker_index` (an `int` that is always set in `__init__` and already used for `strategy.next_request_time(worker_index=self.worker_index)`), so no `None` fallback is needed at these sites.
- Regression test `TestWorkerProcessMultiturn::test_worker_index_zero_reported_as_scheduler_node_id` builds a worker with `worker_index=0` and checks `scheduler_node_id == 0` on both the dequeue path and the cancel path.

## Test Plan

- Regression test fails on `main` (`AssertionError: assert -1 == 0`) and passes with this change.
- `python -m pytest tests/unit/scheduler/test_worker.py -q` → 11 passed, 12 xfailed (main: 10 passed, 12 xfailed; the xfails are pre-existing).
- `ruff check` / `ruff format --check` on both files: clean.
- `mypy --check-untyped-defs src/guidellm/scheduler/worker.py`: no issues.

## Related Issues

- Resolves #1043

---

## Use of AI

Code and test were produced with Claude Code under my direction and review; the commit carries the `Generated-by:` trailer.

---

# git log

commit 2132b8943e6e31134c91d3ca787949668ff35e05
Author: Himanshu Prajapati <himanshuprajapati15072003@gmail.com>
Date:   Mon Aug 24 23:10:57 2026 +0530

    Report worker 0 correctly in RequestInfo.scheduler_node_id
    
    `self.messaging.worker_index or -1` treats worker index 0 as missing
    because 0 is falsy, so every request handled by the first worker was
    reported with the -1 "unknown" sentinel. Use the worker's own
    `self.worker_index`, which is always set, at the three sites.
    
    Adds a regression test covering both the dequeue path (_prepare_node)
    and the cancel path (_cancel_requests_loop) for worker 0.
    
    Fixes #1043
    
    Generated-by: Claude Code
    Co-Authored-By: Claude <noreply@anthropic.com>
    Signed-off-by: Himanshu Prajapati <himanshuprajapati15072003@gmail.com>

---------

Co-Authored-By: Claude <noreply@anthropic.com>
Generated-by: Claude Code
Signed-off-by: Himanshu Prajapati <himanshuprajapati15072003@gmail.com>

### DIFF
--- a/src/guidellm/scheduler/worker.py
+++ b/src/guidellm/scheduler/worker.py
@@ -329,7 +329,7 @@ class WorkerProcess(Generic[RequestT, ResponseT]):
                 node = state.graph.nodes[node_id]
                 request_info = state.request_infos.get(node_id)
                 if request_info is not None:
-                    request_info.scheduler_node_id = self.messaging.worker_index or -1
+                    request_info.scheduler_node_id = self.worker_index
                     request_info.error = "Request was cancelled"
                     request_info.timings.resolve_end = time.time()
                     self._send_update("cancelled", None, node.request, request_info)
@@ -345,7 +345,7 @@ class WorkerProcess(Generic[RequestT, ResponseT]):
             for node_id, node in graph.nodes.items():
                 request_info = graph.request_infos.get(node_id)
                 if request_info is not None:
-                    request_info.scheduler_node_id = self.messaging.worker_index or -1
+                    request_info.scheduler_node_id = self.worker_index
                     request_info.error = "Request was cancelled"
                     request_info.timings.resolve_end = time.time()
                     self._send_update("cancelled", None, node.request, request_info)
@@ -438,7 +438,7 @@ class WorkerProcess(Generic[RequestT, ResponseT]):
                 f"No RequestInfo for node '{node_id}' in graph '{state.graph.graph_id}'"
             )
         request_info.timings.dequeued = time.time()
-        request_info.scheduler_node_id = self.messaging.worker_index or -1
+        request_info.scheduler_node_id = self.worker_index
         request_info.timings.targeted_start = target_start
         self._send_update("pending", None, request, request_info)
         return request, request_info
--- a/tests/unit/scheduler/test_worker.py
+++ b/tests/unit/scheduler/test_worker.py
@@ -1012,3 +1012,57 @@ class TestWorkerProcessMultiturn:
             pass
         assert info_m2.history_len == 3
         assert info_m2.turn_index == 2
+
+    @pytest.mark.regression
+    @pytest.mark.asyncio
+    async def test_worker_index_zero_reported_as_scheduler_node_id(self):
+        """Worker 0 must report scheduler_node_id 0, not the unknown sentinel -1.
+
+        ## WRITTEN BY AI ##
+        """
+        worker = WorkerProcess(
+            worker_index=0,
+            messaging=MockMessaging(worker_index=0),
+            backend=MockBackend(),
+            strategy=SynchronousStrategy(),
+            async_limit=5,
+            fut_scheduling_time_limit=10.0,
+            startup_barrier=Barrier(2),
+            requests_generated_event=Event(),
+            constraint_reached_event=Event(),
+            shutdown_event=Event(),
+            error_event=Event(),
+        )
+        graph = ConversationGraph(
+            graph_id="worker_zero",
+            nodes={
+                "n0": ConversationNode(node_id="n0", agent_id="a", request="r0"),
+                "n1": ConversationNode(node_id="n1", agent_id="a", request="r1"),
+            },
+            edges=[],
+            request_infos={
+                "n0": RequestInfo(request_id="id0", node_id="n0"),
+                "n1": RequestInfo(request_id="id1", node_id="n1"),
+            },
+        )
+        state = DAGExecutionState(graph)
+
+        # Dequeue path
+        _, request_info = worker._prepare_node(state, "n0", target_start=time.time())
+        assert request_info.scheduler_node_id == 0
+
+        # Cancel path for graphs still queued on the worker
+        worker.turns_queue.append(state)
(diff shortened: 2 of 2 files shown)


sjmonson (approved):
LGTM

dbutenhof (commented):
That looks good.

There are a couple other places where `(Worker).messaging.worker_index` is used in error messages, and it's not entirely clear from context that they're necessarily created by `create_worker_copy` (which is the only place we seem to set the message `worker_index`), so it's possible there are cases where we report (in `WorkerProcess.run` + others) "None" in `"Worker process {self.messaging.worker_index} encountered an error: {err}"`. Perhaps those should also use `self.worker_index`?.

And ... it got merged while I was reviewing, so this doesn't matter anyway.

sjmonson:
> And ... it got merged while I was reviewing, so this doesn't matter anyway.

Ah sorry my bad. Since this PR solved the original problem (wrong worker index in json output) I didn't consider looking for other uses.

dbutenhof:
> > And ... it got merged while I was reviewing, so this doesn't matter anyway.
> 
> Ah sorry my bad. Since this PR solved the original problem (wrong worker index in json output) I didn't consider looking for other uses.

I was going to approve anyway, with the comment ... this fixes "happy path" JSON output... the others are exceptions reporting worker failures -- less likely and the internal "worker index" is actually of questionable value at that point.

(That does, however, make me wonder why we only set the message worker index on the "copy" path, but I didn't analyze all the ways into the constructors...)

himanshu1573:
@dbutenhof I was curious about your question, so I went and checked all the ways a messaging object gets created. Short answer: the copy path is the only one that needs the worker index, and nothing is broken today.

Here is what I found on main (963d1a15):

1. The main process builds the object directly in `worker_group.py` (lines 200, 209, 219) and does not pass `worker_index`, so it stays `None`. That is correct, because the main process is the publisher and not a worker.
2. `worker_group.py:244` is the only place in `src/` that creates a worker side object, and it always passes the index: `create_worker_copy(worker_index=rank)`.
3. All three `create_worker_copy` versions (Queue, ManagerQueue, Pipe) put `"worker_index": worker_index` into the copy args.
4. The other constructor calls in `messaging.py` are only examples inside docstrings.

I also checked three ways it could go wrong, and all three are fine:

- `final_args = {**copy_args, **kwargs}` lets kwargs win, so I tried to override the index through kwargs. Python blocks it, because `worker_index` is a named argument: `TypeError: got multiple values for argument`.
- There is no custom `__getstate__`, `__setstate__` or `__reduce__` on the messaging classes, so the index is a normal attribute and survives pickling into a spawned process. If it did not survive, every worker would act like the main process and read the wrong queue, so we would notice immediately.
- Worker 0 is safe here. Every role check in `messaging.py` uses `is None` or `is not None`, never a truthy test, so index 0 is still treated as a worker. The truthy bug only existed in `worker.py`, which this PR fixed.

So the rule "worker_index is None means this is the main process endpoint" is true everywhere. The only thing I noticed is that this rule is not written down anywhere. The constructor shows `worker_index: int | None = None` like a normal optional value, but it actually decides which queue you read from and which end of the pipe you use. Someone who builds a worker side object directly one day would get the wrong behaviour with no error message.

If you think it is worth it, I am happy to send a small doc only PR that states this rule in the class docstring. I did not want to send it without asking first.

PR #1053, opened 2026-08-26
Add scheduler dispatch delay and schedule-relative latency metrics

### LINKED ISSUE #1048: Reported latency percentiles exclude scheduler dispatch delay (coordinated omission)
### Summary

When the scheduler cannot dispatch requests at the configured rate, the resulting backlog is excluded from every reported latency percentile.

Reported `request_latency` measures `request_end - request_start`, where `request_start` is when the client actually sent the request — not `targeted_start`, the time the load schedule intended it to be sent. The gap between the two is absorbed silently.

This is coordinated omission, the measurement error Gil Tene described for open-loop load generators: when the generator's own dispatch stalls, the resulting delay is never attributed to any request, so reported tail latency improves precisely when the system under test is falling behind. HdrHistogram corrects for it with `recordValueWithExpectedInterval`, and `wrk2` was written specifically to avoid it. The correction below is the same one, applied at the reporting layer.

### Reproduction

Uses only the shipped mock server. The server stays healthy for the whole run: ~190 ms service time, no server-side queueing. The client is the bottleneck.

Terminal 1:

```bash
guidellm mock-server --host 127.0.0.1 --port 8123 --model mock \
  --ttft-ms 100 --itl-ms 5 --output-tokens 16
```

Terminal 2:

```bash
guidellm run \
  --backend kind=openai_http,target=http://127.0.0.1:8123 \
  --profile kind=constant,rate=40,max_concurrency=4 \
  --constraint kind=max_duration,seconds=15 \
  --data kind=synthetic_text,prompt_tokens=32,output_tokens=16 \
  --tokenizer kind=huggingface_auto,model=tests/fixtures/tokenizers/minimal \
  --output kind=json,path=co.json
```

Comparing the reported percentiles against the same requests measured from `targeted_start` (n = 316, seconds):

| source | p50 | p95 | p99 |
| --- | --- | --- | --- |
| reported `request_latency` | 0.189 | 0.195 | 0.198 |
| `request_end - targeted_start` | 3.718 | 6.865 | 7.148 |

Reported p99 is 36x lower than the latency implied by the requested arrival schedule.

What the console shows for that run:

```
| Benchmark | Request Latency || TTFT         ||
| Strategy  | Sec             || ms           ||
|           | Mdn     | p95    | Mdn   | p95   |
|-----------|---------|--------|-------|-------|
| constant  | 0.2     | 0.2    | 110.5 | 115.8 |
```

The throughput table reports 20.0 req/s against a requested 40 req/s. The shortfall is not shown next to the requested rate, and no output mentions dispatch delay.

A control run at a sustainable rate (10 req/s, no concurrency cap) gives a mean dispatch delay of 0.003 s, so the effect above is not measurement noise.

### Where it comes from

`WorkerProcess._process_requests_loop` acquires the concurrency semaphore before asking the strategy for the next scheduled time:

https://github.com/vllm-project/guidellm/blob/963d1a15ad20337f41a28ab3b5e3ade6f73fb6d4/src/guidellm/scheduler/worker.py#L302-L303

The schedules themselves are correct. `AsyncConstantStrategy` and `AsyncPoissonStrategy` derive arrival times from a monotonic index/offset, so a stalled dispatch loop does not shift the schedule, and `targeted_start` remains the true intended arrival time. The data needed to detect this is already captured.

The omission is in the reporting layer:

- `GenerativeRequestStats.request_latency` = `request_end - request_start` ([request_stats.py#L103](https://github.com/vllm-project/guidellm/blob/963d1a15ad20337f41a28ab3b5e3ade6f73fb6d4/src/guidellm/schemas/request_stats.py#L103))
- `time_to_first_token_ms` = `first_token - request_start` ([request_stats.py#L162](https://github.com/vllm-project/guidellm/blob/963d1a15ad20337f41a28ab3b5e3ade6f73fb6d4/src/guidellm/schemas/request_stats.py#L162))
- `SchedulerMetrics.request_targeted_start_delay` does record `request_start - targeted_start`, but only as a running mean, with no distribution, and it is not printed by any console output.

For the run above that mean was 3.764 s, against a reported latency mean of 0.187 s.

### Why this is reachable on default settings

`sweep` is the default profile, and `SweepProfileArgs.max_concurrency` defaults to 512 ([sweep.py#L44](https://github.com/vllm-project/guidellm/blob/963d1a15ad20337f41a28ab3b5e3ade6f73fb6d4/src/guidellm/benchmark/profiles/sweep.py#L44)).

The cap binds hardest at the top of the sweep range, where the interpolated constant rates approach a throughput figure that was measured in closed-loop mode. An open-loop constant arrival process generally cannot sustain that rate, so the dispatch loop falls behind exactly in the region where tail latency matters most.

### Proposed fix

Surface the delay rather than redefine existing numbers:

1. Add a per-request dispatch delay and a targeted-start latency (`request_end - targeted_start`) to `GenerativeRequestStats`.
2. Expose both as `StatusDistributionSummary` on `GenerativeMetrics`, so they carry percentiles like every other metric.
3. Add them to the console latency table and the CSV output.

This change is purely additive. `request_latency`, `time_to_first_token_ms` and every other existing metric keep their current definitions and values, so historical comparisons and saved reports stay valid. The new fields are added alongside them, and new CSV columns are appended rather than inserted.

I am putting together a PR along these lines, including a regression test that asserts the delay stays near zero when the harness keeps up and grows when the cap binds. Happy to adjust the direction before I open it.

QHarshil:
## Summary

`request_latency` is measured from the moment a request is dispatched, so any time a request spent waiting on a saturated concurrency limit is excluded from every reported latency percentile. This is the coordinated omission failure mode: reported tail latency improves precisely when the load generator falls behind.

Most of what is needed to detect this is already captured. The scheduler records the intended arrival time in `RequestTimings.targeted_start`, and `SchedulerMetrics.request_targeted_start_delay_avg` already reports a scalar mean of the delay in the CSV. What is missing is a distribution behind that mean, and a metric that folds the delay back into an end-to-end latency.

This change is purely additive. `request_latency`, `time_to_first_token_ms`, and every other existing metric keep their current definitions and values, so historical comparisons and saved reports remain valid. Two new metrics are added alongside them:

```
request_dispatch_delay    = request_start - targeted_start
request_scheduled_latency = request_end   - targeted_start
```

When all three timestamps are present, `request_scheduled_latency == request_dispatch_delay + request_latency`.

Following the feedback on #1048, these are not added to the console latency table, and they are reported as `null` for the strategies where `targeted_start` is an ASAP-style value rather than an arrival schedule. Both are explained under Notes for review.

## Details

- [x] Add `request_dispatch_delay` and `request_scheduled_latency` computed fields to `GenerativeRequestStats`, returning `None` when the timestamps they require are absent
- [x] Add matching `StatusDistributionSummary | None` fields to `GenerativeMetrics`, aggregated in `GenerativeMetrics.compile()`, defaulting to `None` so reports written before these fields existed still deserialize
- [x] Add `SchedulingStrategy.defines_arrival_schedule`, `True` for `constant`, `poisson`, and `trace`, and `False` for the ASAP-style strategies. When `False`, both metrics compile to `None`
- [x] Export both metrics with full distributions in the CSV output, emitting no columns when they are `None`
- [x] Document both metrics in `docs/guides/metrics.md`, including which profiles they apply to and why they are absent from console output
- [x] Add an optional `max_concurrency` parameter to the E2E `GuidellmClient` helper so a test can bind the concurrency cap

## Test Plan

- Unit semantics on fixed timestamps, including the `scheduled == dispatch_delay + latency` invariant, and `None` when `targeted_start` is absent
- Aggregation through real `GenerativeMetrics.compile()`, including a percentile case where `request_latency` p99 stays at roughly 0.1s while `request_scheduled_latency` p99 exceeds 90s, making the coordinated-omission effect visible
- Requests that never reached dispatch are excluded from the distributions rather than recorded as zero delay
- Strategy gating: `defines_arrival_schedule` asserted per strategy, and `compile()` returning `None` for both metrics under `throughput`
- Backward compatibility: a compiled `GenerativeMetrics` is dumped, both new keys are stripped to mimic a report written before they existed, and the payload is re-validated
- CSV: both metrics are exported when applicable, no columns are emitted when they are `None`, and pre-existing latency columns keep their relative order
- Console: regression test asserting neither metric appears in the Request Latency table and that the existing columns are unchanged
- E2E, two regimes against the shipped mock server: dispatch delay becomes large when the concurrency cap binds, and stays near zero when the scheduler keeps up

Manual reproduction against the mock server on a `constant` profile, at a rate the concurrency cap cannot sustain. The server stays healthy throughout at roughly 190ms service time:

| source | p50 | p95 | p99 |
| --- | --- | --- | --- |
| reported `request_latency` | 0.189 | 0.195 | 0.198 |
| `request_end - targeted_start` | 3.718 | 6.865 | 7.148 |

## Notes for review

**This goes one step beyond the console-only change requested in #1048, and is easy to scale back.** The ask was to omit these from console output. I also gated them out of the compiled report, because under `throughput` every request targets the benchmark start time, so the values track elapsed run time rather than any delay. Since `sweep` is the default profile and runs `throughput`, a default CSV would otherwise carry a `Scheduled Latency` column roughly equal to half the run duration, and a CSV consumer never sees the documentation caveat. If you would rather keep the raw values in the report and only suppress the console, that is a one-property change.

**Why all three strategies, and why `null` rather than zero.** `defines_arrival_schedule` is `False` for exactly the three profiles named in the issue thread. `synchronous` and `concurrent` target the previous request's completion, so a delay measured against it is circular: it describes harness turnaround rather than lag against an arrival schedule. `throughput` targets the benchmark start, so the value grows with elapsed run time. They are null rather than a zero-filled distribution because a zero serializes as `mean = 0.0` and reads as "no delay measured", which is a stronger and more misleading claim than "not applicable".

**Deliberate asymmetry with `request_latency`.** `request_latency` maps a missing timing to `0.0` via `or 0.0` in `compile()`. The two new metrics instead pass the raw value, so a request with no `request_start` is dropped by `from_values_function` rather than recorded as zero. That is deliberate. A request cancelled while the scheduler was backed up is exactly the population these metrics describe, and recording it as zero delay would bias the tail toward the answer this PR exists to correct. I left `request_latency` alone rather than change an existing metric's values.

**Relationship to the existing scheduler metric.** `Dispatch Delay` is the per-status distribution of the same quantity as `Request Targeted Start Delay Avg`. The scheduler value is a running mean over the whole run, while the new one is scoped to the measurement window, so the two will not match exactly.

**Known limitation.** Applicability is modelled per strategy, but it is really a per-request property of where `targeted_start` came from. Two consequences, both documented in the metrics guide: a `trace` request carrying no relative timestamp falls back to the benchmark start and is still included, and for multi-turn datasets the think time between turns lands inside Dispatch Delay. Moving the signal onto `RequestTimings` per request would fix both and remove the branch in `compile()`. I did not do that here because it touches the worker dispatch path and the per-request schema. I can implement it in this PR if you prefer that shape.

**CSV column placement in mixed-strategy runs.** `_align_columns` unions headers in first-seen order, so in a `sweep`, where the first benchmark is `synchronous` and contributes no columns for these metrics, the two new groups land at the right-hand end of the CSV rather than beside `Request Latency`. That is existing behaviour of the aligner, not something introduced here.

## Related Issues

- Resolves #1048

---

## Use of AI

---

# git log

commit b88ce8c7536547046ae9b2828e04bf3a8e6fe832
Date:   Mon Aug 24 16:14:20 2026 -0700

    feat: add scheduler dispatch delay and schedule-relative latency metrics
    
    request_latency is measured from the moment a request is dispatched, so
    any time it spent waiting on a saturated concurrency limit is excluded
    from every reported percentile. The scheduler already records the
    intended arrival time in RequestTimings.targeted_start, and
    SchedulerMetrics reports a scalar mean of that delay, but there is no
    distribution behind it and nothing folds it back into an end-to-end
    latency.
    
    Add two metrics alongside the existing ones:
    
      request_dispatch_delay    = request_start - targeted_start
      request_scheduled_latency = request_end - targeted_start
    
    Dispatch delay is the per-status distribution of the same quantity as
    the existing Request Targeted Start Delay Avg, scoped to the measurement
    window rather than the whole run, so the two will not match exactly.
    Scheduled latency is new.
    
    request_latency and every other existing metric keep their current
    definitions and values. The new GenerativeMetrics fields carry defaults
    so reports written before they existed still load.
    
    Both only apply to strategies that define an arrival schedule. A new
    SchedulingStrategy.defines_arrival_schedule property is True for
    constant, poisson, and trace, and False for synchronous, concurrent, and
    throughput. Synchronous and concurrent target the previous request's
    completion, so a delay measured against it is circular rather than lag
    against an arrival schedule; throughput targets the benchmark start, so
    the value would grow with elapsed run time. When the property is False
    both metrics are reported as None rather than zero, since zero would
    read as "no delay measured" instead of "not applicable", and the CSV
    emits no columns for them. This matters for the default sweep profile,
    which runs synchronous and throughput alongside its rate-based
    strategies.
    
    Per-request values are left None when a request never reached dispatch,
    so requests cancelled during a backlog are excluded from the
    distributions rather than being counted as zero delay.
    
    Where they do apply they are exposed through the serialized report and
    the CSV output. They are left out of the console latency table in all
    cases, since that table is shared across every profile in a run. The
    metrics guide documents the applicability rules.
    
    Against the shipped mock server on a constant-rate profile at a rate the
    concurrency cap cannot sustain, reported p99 request latency was 0.198s
    while p99 measured from targeted_start was 7.148s.
    
    Closes #1048
    
commit 4f0dd5ae2516e8d37ef49276436dcab367ae9fdf
Date:   Thu Aug 27 12:02:02 2026 -0700

    docs: move scheduling metric applicability notes
    
    Move the Dispatch Delay and Scheduled Latency applicability section out
    of the flat list of metric definitions and to the end of the metrics
    guide, leaving a pointer to it from the definitions.
    
commit c0a0273893a6fc8292fc30bbfc97f0cb27c988ae
Date:   Thu Aug 27 12:12:21 2026 -0700

    docs(metrics): simplify schedule-relative metric descriptions
    
    Drop the note about reports predating these fields from the
    request_dispatch_delay and request_scheduled_latency descriptions. The
    default of None still allows older reports to deserialize; only the
    description text changes.
    
commit 38a93bbf85b7bcedb5bc9630383c9772a8f5aedd
Date:   Thu Aug 27 13:59:49 2026 -0700

    docs(metrics): note units in schedule-relative metric docstrings
    
    State seconds in the summary line of request_dispatch_delay and
    request_scheduled_latency so the unit is visible without reading the
    return description.
    
---------

### DIFF
--- a/docs/guides/metrics.md
+++ b/docs/guides/metrics.md
@@ -66,23 +66,7 @@ These metrics provide a breakdown of the overall request statuses, helping users
 - **Definition**: `request_scheduled_latency` is `request_end - targeted_start`: request latency measured from the scheduled arrival time rather than from dispatch. When all three timestamps are present it equals Dispatch Delay plus Request Latency.
 - **Use Case**: Describes what a client holding to the configured arrival schedule would have experienced, including time spent waiting to be dispatched. When the benchmark keeps up this matches Request Latency; when it falls behind, the gap between the two is latency that Request Latency alone does not show.
 
-### Applicability of Dispatch Delay and Scheduled Latency
-
-Both metrics are derived from `targeted_start`, so they only describe arrival-schedule delay for strategies that define an arrival schedule:
-
-- `constant` and `poisson` derive each target from the configured rate.
-- `trace` derives each target from the replayed dataset timestamps. Trace requests that carry no relative timestamp fall back to the benchmark start time and take on the `throughput` caveat below.
-
-The `synchronous`, `concurrent`, and `throughput` strategies set an ASAP-style target instead, and each is unusable for a different reason:
-
-- `synchronous` and `concurrent` target the previous request's completion, apart from the requests staggered across a configured rampup. The target is therefore derived from the system's own responses, so a delay measured against it is circular: it describes harness turnaround rather than lag against an arrival schedule.
-- `throughput` targets the benchmark start time for every request, so the value grows with elapsed run time and is not a delay at all.
-
-For those three strategies **both metrics are reported as `null`** rather than as zero, and their columns do not appear in the CSV. Reporting zero would read as "no delay measured", which is a stronger and more misleading claim than "not applicable". This matters for the default `sweep` profile, which runs `synchronous` and `throughput` alongside its rate-based strategies.
-
-For multi-turn conversation datasets, a turn's target is fixed when a scheduler slot opens rather than when the preceding turn completes, so configured think time between turns is counted as Dispatch Delay.
-
-Where the metrics do apply they are recorded in the serialized report and the CSV output for offline analysis. They are omitted from the final console latency table in all cases, since that table is shared across every profile in a run.
+Dispatch Delay and Scheduled Latency only apply to some of the scheduling strategies. See [Applicability of Dispatch Delay and Scheduled Latency](#applicability-of-dispatch-delay-and-scheduled-latency) below.
 
 ### Time to First Token (TTFT)
 
@@ -148,3 +132,21 @@ GuideLLM calculates a comprehensive set of percentiles for each metric, includin
 - **Percentiles**: Offer a detailed view of the distribution, helping identify outliers and performance at different levels of service.
 
 By combining these metrics and statistical summaries, GuideLLM enables users to gain a deep understanding of their LLM deployments, optimize performance, and ensure scalability and cost-effectiveness.
+
+## Applicability of Dispatch Delay and Scheduled Latency
+
+Both metrics are derived from `targeted_start`, so they only describe arrival-schedule delay for strategies that define an arrival schedule:
+
+- `constant` and `poisson` derive each target from the configured rate.
+- `trace` derives each target from the replayed dataset timestamps. Trace requests that carry no relative timestamp fall back to the benchmark start time and take on the `throughput` caveat below.
+
+The `synchronous`, `concurrent`, and `throughput` strategies set an ASAP-style target instead, and each is unusable for a different reason:
+
+- `synchronous` and `concurrent` target the previous request's completion, apart from the requests staggered across a configured rampup. The target is therefore derived from the system's own responses, so a delay measured against it is circular: it describes harness turnaround rather than lag against an arrival schedule.
+- `throughput` targets the benchmark start time for every request, so the value grows with elapsed run time and is not a delay at all.
+
+For those three strategies **both metrics are reported as `null`** rather than as zero, and their columns do not appear in the CSV. Reporting zero would read as "no delay measured", which is a stronger and more misleading claim than "not applicable". This matters for the default `sweep` profile, which runs `synchronous` and `throughput` alongside its rate-based strategies.
+
+For multi-turn conversation datasets, a turn's target is fixed when a scheduler slot opens rather than when the preceding turn completes, so configured think time between turns is counted as Dispatch Delay.
+
+Where the metrics do apply they are recorded in the serialized report and the CSV output for offline analysis. They are omitted from the final console latency table in all cases, since that table is shared across every profile in a run.
--- a/docs/guides/metrics.md
+++ b/docs/guides/metrics.md
@@ -56,6 +56,34 @@ These metrics provide a breakdown of the overall request statuses, helping users
 - **Definition**: The time taken to process a single request, from start to finish.
 - **Use Case**: A critical metric for evaluating the responsiveness of the system.
 
+### Dispatch Delay
+
+- **Definition**: `request_dispatch_delay` is `request_start - targeted_start`: the time between when a request was scheduled to arrive and when it was actually sent.
+- **Use Case**: Reveals whether the benchmark itself kept up with the configured arrival schedule. A request cannot be dispatched while the concurrency limit is saturated, and that wait is not part of Request Latency. A delay near zero means the reported latencies reflect the full picture; a large delay means the load generator fell behind and the server saw a lower arrival rate than was requested.
+
+### Scheduled Latency
+
+- **Definition**: `request_scheduled_latency` is `request_end - targeted_start`: request latency measured from the scheduled arrival time rather than from dispatch. When all three timestamps are present it equals Dispatch Delay plus Request Latency.
+- **Use Case**: Describes what a client holding to the configured arrival schedule would have experienced, including time spent waiting to be dispatched. When the benchmark keeps up this matches Request Latency; when it falls behind, the gap between the two is latency that Request Latency alone does not show.
+
+### Applicability of Dispatch Delay and Scheduled Latency
+
+Both metrics are derived from `targeted_start`, so they only describe arrival-schedule delay for strategies that define an arrival schedule:
+
+- `constant` and `poisson` derive each target from the configured rate.
+- `trace` derives each target from the replayed dataset timestamps. Trace requests that carry no relative timestamp fall back to the benchmark start time and take on the `throughput` caveat below.
+
+The `synchronous`, `concurrent`, and `throughput` strategies set an ASAP-style target instead, and each is unusable for a different reason:
+
+- `synchronous` and `concurrent` target the previous request's completion, apart from the requests staggered across a configured rampup. The target is therefore derived from the system's own responses, so a delay measured against it is circular: it describes harness turnaround rather than lag against an arrival schedule.
+- `throughput` targets the benchmark start time for every request, so the value grows with elapsed run time and is not a delay at all.
+
+For those three strategies **both metrics are reported as `null`** rather than as zero, and their columns do not appear in the CSV. Reporting zero would read as "no delay measured", which is a stronger and more misleading claim than "not applicable". This matters for the default `sweep` profile, which runs `synchronous` and `throughput` alongside its rate-based strategies.
+
+For multi-turn conversation datasets, a turn's target is fixed when a scheduler slot opens rather than when the preceding turn completes, so configured think time between turns is counted as Dispatch Delay.
+
+Where the metrics do apply they are recorded in the serialized report and the CSV output for offline analysis. They are omitted from the final console latency table in all cases, since that table is shared across every profile in a run.
+
 ### Time to First Token (TTFT)
 
 - **Definition**: The time taken to generate the first token of the output.
--- a/src/guidellm/benchmark/schemas/metrics.py
+++ b/src/guidellm/benchmark/schemas/metrics.py
@@ -790,7 +790,7 @@ class GenerativeMetrics(StandardBaseDict):
             "Distribution of delays between a request's targeted start and the "
             "time it was dispatched. Grows when the scheduler cannot keep up "
             "with the configured rate. None when the strategy does not define "
-            "an arrival schedule, or for reports predating this field"
+            "an arrival schedule"
         ),
     )
     request_scheduled_latency: StatusDistributionSummary | None = Field(
@@ -799,7 +799,7 @@ class GenerativeMetrics(StandardBaseDict):
             "Distribution of request latencies measured from each request's "
             "targeted start rather than its dispatch, including scheduler "
             "delay. None when the strategy does not define an arrival "
-            "schedule, or for reports predating this field"
+            "schedule"
         ),
     )
     request_streaming_iterations_count: StatusDistributionSummary = Field(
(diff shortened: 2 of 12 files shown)


src/guidellm/benchmark/schemas/metrics.py
```diff
@@ -784,6 +784,24 @@ class GenerativeMetrics(StandardBaseDict):
     request_latency: StatusDistributionSummary = Field(
         description="Distribution of request latencies for completed requests"
     )
+    request_dispatch_delay: StatusDistributionSummary | None = Field(
+        default=None,
+        description=(
+            "Distribution of delays between a request's targeted start and the "
+            "time it was dispatched. Grows when the scheduler cannot keep up "
+            "with the configured rate. None when the strategy does not define "
+            "an arrival schedule, or for reports predating this field"
```
dbutenhof: "or for reports predating this field" seems a bit obvious, and I don't see any value in including that phrase (either here or on latency, below).

docs/guides/metrics.md
```diff
@@ -56,6 +56,34 @@ These metrics provide a breakdown of the overall request statuses, helping users
...
+- `throughput` targets the benchmark start time for every request, so the value grows with elapsed run time and is not a delay at all.
+
+For those three strategies **both metrics are reported as `null`** rather than as zero, and their columns do not appear in the CSV. Reporting zero would read as "no delay measured", which is a stronger and more misleading claim than "not applicable". This matters for the default `sweep` profile, which runs `synchronous` and `throughput` alongside its rate-based strategies.
+
+For multi-turn conversation datasets, a turn's target is fixed when a scheduler slot opens rather than when the preceding turn completes, so configured think time between turns is counted as Dispatch Delay.
+
+Where the metrics do apply they are recorded in the serialized report and the CSV output for offline analysis. They are omitted from the final console latency table in all cases, since that table is shared across every profile in a run.
+
```
sjmonson, line 70: This is useful to have but maybe more it to the bottom of this file.

sjmonson (commented):
One minor nit on the docs. Nothing stands out to me as a problem with the code though I'll try to run a few test benchmarks just to see if it matches my expectations.

dbutenhof (commented):
Mostly looks OK ... one minor documentation comment (which I thought I'd submitted several days ago...)

src/guidellm/benchmark/schemas/metrics.py
QHarshil: Removed that phrase from both metric descriptions, thanks.

docs/guides/metrics.md
QHarshil: Moved the applicability section to the bottom of the metrics guide and left a short pointer by the metric definitions, thanks.

src/guidellm/schemas/base/request_stats.py
```diff
@@ -113,6 +113,51 @@ def request_latency(self) -> float | None:
 
         return end - start
 
+    @computed_field  # type: ignore[misc]
+    @property
+    def request_dispatch_delay(self) -> float | None:
+        """
+        Delay between the scheduled arrival time and the actual dispatch.
```
sjmonson:
```suggestion
        Delay between the scheduled arrival time and the actual dispatch in seconds.
```

src/guidellm/schemas/base/request_stats.py
```diff
@@ -113,6 +113,51 @@ def request_latency(self) -> float | None:
...
+        if targeted is None or start is None:
+            return None
+
+        return start - targeted
+
+    @computed_field  # type: ignore[misc]
+    @property
+    def request_scheduled_latency(self) -> float | None:
+        """
+        Request latency measured from the scheduled arrival time.
```
sjmonson:
```suggestion
        Request latency measured from the scheduled arrival time in seconds.
```

sjmonson (changes requested):
One more documentation thing. Otherwise I tested a run and the results seem good.

src/guidellm/schemas/base/request_stats.py
QHarshil: Applied, thanks.

src/guidellm/schemas/base/request_stats.py
QHarshil: Applied, thanks.

dbutenhof (approved):

sjmonson (approved):

jaredoconnell (approved):

PR #1054, opened 2026-08-26
fix: reject empty trace conversations

tangming1996:
## Summary

- reject empty trace conversations during trace dataset validation
- fail early with an actionable `DataNotSupportedError` instead of an `IndexError` during iteration
- add a regression test for a later empty conversation in a WEKA trace

## Testing

- `tox -e tests -- tests/unit/data/deserializers/test_trace_weka.py -q` (16 passed)

---

- [x] I certify that my contribution is made under the terms of the Developer Certificate of Origin.

---

# git log

commit 0eec7735018996c0369e540ac243f3fb52043754
Date:   Wed Aug 26 13:57:13 2026 +0800

    fix: reject empty trace conversations
    
---------

### DIFF
--- a/src/guidellm/data/deserializers/trace_common.py
+++ b/src/guidellm/data/deserializers/trace_common.py
@@ -339,6 +339,8 @@ def _validate_dataset(config: TraceDataArgs, trace_format: TraceFormatBase) -> N
         }
     )
     for conv in trace_format:  # type: ignore[attr-defined]
+        if len(conv) == 0:
+            raise DataNotSupportedError("Trace conversation is empty")
         if config.conversation_id_column in features:
             features.pop(config.conversation_id_column)
         _raise_if_nonetype_found(conv)
--- a/tests/unit/data/deserializers/test_trace_weka.py
+++ b/tests/unit/data/deserializers/test_trace_weka.py
@@ -339,6 +339,23 @@ class TestWEKATraceFormat:
         with pytest.raises(DataNotSupportedError, match=match):
             self.deserialize(deserializer, trace)
 
+    @pytest.mark.regression
+    def test_rejects_empty_conversation_after_first_row(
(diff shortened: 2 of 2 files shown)


sjmonson (approved):
Possibly it would be better to skip empty rows rather then error. I'll approve for now but another maintainer may disagree.

dbutenhof (commented):
> Possibly it would be better to skip empty rows rather then error. I'll approve for now but another maintainer may disagree.

I got hung up over that question, too. Is it an error, or noise? Another option would be a warning (perhaps only on the first occurrence). Anyway, I did some searches and didn't find a clear canonical answer; so I guess I'm OK with this. However, I'm going to submit this as a Comment rather than approval so I don't auto-merge before someone else has a chance to chime in. If there's no objection before tomorrow, I'll merge it.

jaredoconnell:
Warning logs might be better. Erroring would make it so you just can't use a dataset if it has one bad entry.

tangming1996:
To clarify, this fix does not introduce a new restriction. Previously, an empty conversation after a valid one already caused deserialization to fail: dataset.cast(features) rejected the empty dataset because its columns did not match the expected schema, resulting in a DataNotSupportedError.
This fix simply catches that condition earlier and reports a clearer error: Trace conversation is empty.
Skipping empty conversations with a warning is a reasonable improvement, but it would change the existing behavior rather than correct a regression introduced by this fix.

sjmonson:
> To clarify, this fix does not introduce a new restriction.

Yeah, I reread the code after posting my original comment and realized this happens during initial validation and not during dataset iteration. I think ideally in the future we just get rid of this code, we shouldn't be iterating though the entire dataset before the benchmark. So for now this change looks good @jaredoconnell?

jaredoconnell (approved):

PR #1055, opened 2026-08-26
fix: make terminal request bookkeeping idempotent under max_duration cancellation

### LINKED ISSUE #1049: [Bug]: max_duration at high concurrency double-reports in-flight requests, causing a KeyError abort or a shutdown hang
### Bug Description

When a run ends on a `max_duration` constraint with many requests in flight, the same request can be reported **twice** — once as `cancelled` by the worker's cancel loop, and once as `completed`/`errored` by its own in-flight task. The scheduler's bookkeeping is not tolerant of that, so the run ends in one of two ways:

1. **Abort** — `Error processing received update: '<request_id>'` (a `KeyError`), which sets `error_event`, and every worker then dies with `received error signal`.
2. **Hang** — all requests reach a terminal state, `end_processing_time` is set, `processing_requests` reaches 0, but `Scheduler.run()` never returns.

On `main` (963d1a15), 10 runs at `max_concurrency=500` with `max_duration=8`: **3 hung, 3 aborted, 4 completed cleanly.** It is timing dependent, so it does not reproduce every run, but it is frequent at high concurrency.

### Root cause

Captured directly by printing the status that raises in `WorkerGroupState.received_callback`:

```
raising_status='completed' request_id=6a00e914-f015-40fe-8af2-342f2da41a24 err=KeyError: '6a00e914-f015-40fe-8af2-342f2da41a24'
```

So a **`completed`** update arrives for a request that is no longer in `_processing_request_ids`, because a `cancelled` update for the same request already removed it.

The sequence:

1. Request is in flight; the main process has its id in `_processing_request_ids`.
2. The duration constraint fires. `WorkerProcess._process_requests` calls `processing_task.cancel()` (`worker.py:240`) and then immediately `await self._cancel_requests_loop()` (`worker.py:243`). `cancel()` only *schedules* cancellation — the request tasks have not processed their `CancelledError` yet.
3. `_cancel_requests_loop` walks `self.turns_queue` and calls `state.abort()` (`worker.py:326-327`), which returns **every node that is not completed, including nodes an in-flight task still owns**, and sends a `cancelled` update for each. The main process removes the id (the `errored`/`cancelled` branch at `worker_group.py:750` uses guarded removes) and does `processed_requests += 1`.
4. The in-flight task, which was about to finish anyway, completes and sends its own `completed` update. `worker_group.py:744-746` then does an unguarded `self._processing_request_ids.remove(info.request_id)` → **`KeyError`** (manifestation 1).

When the timing works out so the second update does not hit the unguarded `remove`, the counters still drift instead: both branches do `processed_requests += 1` for the same request, so `processed_requests` overshoots `created_requests`, the equality check at `worker_group.py:684` never holds, `shutdown_event` is never set, and `request_updates()` loops forever (manifestation 2).

Both manifestations therefore come from the same cause: **a request that is cancelled by the cancel loop while its own task is still running gets two terminal updates.**

### Expected Behavior

A `max_duration` run should end cleanly at any concurrency: exactly one terminal update per request, and `Scheduler.run()` returns.

### Steps to Reproduce

No model or server needed — this is scheduler-only, with an in-process backend that just sleeps:

```python
# race.py
import asyncio, sys, time, uuid
from typing import Any
from pydantic import BaseModel, Field
from guidellm.scheduler import (
    BackendInterface, NonDistributedEnvironment, Scheduler, ThroughputStrategy,
)
from guidellm.scheduler.constraints import MaxDurationConstraint, MaxDurationConstraintArgs
from guidellm.scheduler.schemas import ConversationGraph, ConversationNode
from guidellm.schemas import RequestSettings


class MockRequest(BaseModel):
    payload: str
    id_: str = Field(default_factory=lambda: str(uuid.uuid4()))


class MockConversationGraph(ConversationGraph[MockRequest]):
    """Bound graph type so IPC deserialization restores MockRequest nodes."""


class MockBackend(BackendInterface):
    @property
    def processes_limit(self): return None
    @property
    def requests_limit(self): return None
    def info(self) -> dict[str, Any]: return {"type": "race"}
    async def process_startup(self): pass
    async def validate(self): pass
    async def process_shutdown(self): pass
    async def resolve(self, request, request_info, request_history):
        request_info.timings.request_start = time.time()
        await asyncio.sleep(1.0)
        yield f"r_{request.payload}", request_info


def graphs():
    i = 0
    while True:
        yield MockConversationGraph(
            graph_id=str(uuid.uuid4()),
            nodes={"t0": ConversationNode(
                node_id="t0", agent_id="d",
                request=MockRequest(payload=f"q{i}"), settings=RequestSettings())},
            edges=[])
        i += 1


async def main(conc: int, duration: float):
    if hasattr(Scheduler, "singleton_instance"):
        Scheduler.singleton_instance = None
    n = 0
    async for _resp, _req, _info, _state in Scheduler().run(
        requests=graphs(), backend=MockBackend(),
        strategy=ThroughputStrategy(max_concurrency=conc),
        env=NonDistributedEnvironment(),
        max_duration=MaxDurationConstraint(
            args=MaxDurationConstraintArgs(seconds=duration)).create_constraint(),
    ):
        n += 1
    print(f"clean exit after {n} updates")


if __name__ == "__main__":
    asyncio.run(main(int(sys.argv[1]), float(sys.argv[2])))
```

```bash
for i in 1 2 3 4 5 6 7 8 9 10; do timeout 40 python race.py 500 8; echo "run$i exit=$?"; done
```

`exit=124` is the hang; a `RuntimeError: error_event is set in WorkerProcessGroup` preceded by `Error processing received update: '<uuid>'` is the abort; `clean exit after N updates` is a good run. On my machine (Apple M1, 10 worker processes, `spawn`) 6 of 10 runs failed.

To see the raising status directly, temporarily add a print to the `except` in `WorkerGroupState.received_callback`:

```python
print(f"DIAG raising_status={request_info.status!r} request_id={request_info.request_id} err={type(err).__name__}: {err}", file=sys.stderr, flush=True)
```

### Possible fixes

I did not want to guess which direction you prefer, so I have not opened a PR:

1. **Await cancellation before reporting.** In `_process_requests`, await the cancelled `processing_task` (and its child request tasks) before `_cancel_requests_loop` walks `turns_queue`, so each task reports its own terminal status and the cancel loop only reports what is genuinely still queued.
2. **Do not cancel nodes that a task owns.** Have `DAGExecutionState.abort()` skip claimed nodes (or have `_cancel_requests_loop` skip them) and let the owning task emit the terminal update.
3. **Make the bookkeeping idempotent** in `_update_state_request_counts` — guard the `completed`/`pending`/`in_progress` removes the way the `cancelled` branch already is, and ignore a second terminal update for a request that is already finalized. This alone would stop the crash and the count drift even if a duplicate slips through, and looks like a sensible safety net regardless of which of the above is chosen.

### Environment

- guidellm `main` @ 963d1a15
- Python 3.12.10, macOS (Apple M1), `spawn` start method, 10 worker processes
- Found while benchmarking #1042 at 500–2000 concurrency; that PR neither causes nor fixes this (it reproduces on `main` without it).

AmirF194:
## Summary

A `max_duration` run at high concurrency can abort with a `KeyError` or hang forever, because a request can get two terminal status updates for the same lifecycle.

## Details

- [x] `WorkerProcess._process_requests` (`worker.py`) calls `processing_task.cancel()` then immediately runs `_cancel_requests_loop`, without waiting for the cancellation to actually land. `cancel()` only schedules the exception; the cancel loop can walk `turns_queue` and report a node as `cancelled` while that node's own task is still finishing and about to report itself `completed`.
- [x] `WorkerGroupState._update_state_request_counts` (`worker_group.py`) treated `completed`, `pending`, and `in_progress` as always-valid transitions with unguarded set removals. A second terminal report for an already-finalized request either raised `KeyError` on the unguarded `_processing_request_ids.remove(...)`, or, when the removal happened to hit a different (already-empty) branch, silently incremented `processed_requests` a second time, so the scheduler's shutdown equality check never held and `Scheduler.run()` never returned.

Fix, matching the maintainer's own comment on the issue (options 1 + 3):

- `_process_requests` now awaits the cancelled `processing_task` (suppressing `CancelledError`) before running `_cancel_requests_loop`, so in-flight nodes report their own terminal status first.
- `_update_state_request_counts` is now idempotent: a `pending`/`in_progress` update for a request no longer in the set it expects, or a `completed`/`errored`/`cancelled` update for a request already finalized, is treated as a stray duplicate and ignored instead of raising or double-counting. The completed/errored/cancelled path is extracted into `_finalize_terminal_request`, which returns whether it actually applied the update.

## Test Plan

- New regression tests in `tests/unit/scheduler/test_worker.py` and `tests/unit/scheduler/test_worker_group.py`: fail on `main` (`KeyError`, wrong event ordering, or a double-counted `processed_requests`), pass on this branch. Full `tests/unit/scheduler` suite green on Python 3.10 and 3.12 (549 passed, 16 pre-existing unrelated xfails).
- Reproduced the issue's own `race.py` script directly: 6/6 runs hung on unmodified `main` at `max_concurrency=500, max_duration=8`; 10/10 clean exits on this branch.
- `ruff format --check`, `ruff check`, and `mypy --check-untyped-defs` pass on the changed files. Coverage confirms every changed line is exercised by the new tests.
- Not checked: `tests/integration` and `tests/e2e`, which need a live backend/model server and don't import the changed code paths.

## Related Issues

- Resolves #1049

---

## Use of AI

---

# git log

commit bf59fcd1942c8d3db573cae90bea3ebfc6c629fc
Author: Amir Fathi <amirfathi.me@gmail.com>
Date:   Wed Aug 26 09:02:51 2026 +0000

    fix: make terminal request bookkeeping idempotent under max_duration cancellation
    
    WorkerProcess._process_requests cancelled processing_task and immediately
    ran _cancel_requests_loop without awaiting the cancellation, so a node the
    cancel loop reported as "cancelled" could still be finishing its own
    "completed" report. WorkerGroupState._update_state_request_counts treated
    completed/errored/cancelled as first-report-wins, so the second report hit
    an unguarded set removal (KeyError abort) or, when it missed the removal,
    double-incremented processed_requests so the shutdown equality check never
    held (hang).
    
    Await the cancelled task before the cancel loop runs, so it is the one
    reporting nodes it still owns, and make _update_state_request_counts
    idempotent: a pending/in_progress/terminal update for a request no longer
    in the set it expects is a stray duplicate and is ignored rather than
    raising or double-counting.
    
    Fixes #1049
    
    Generated-by: Claude Code claude-sonnet-5
    Signed-off-by: Amir Fathi <amirfathi.me@gmail.com>

---------

Generated-by: Claude Code claude-sonnet-5
Signed-off-by: Amir Fathi <amirfathi.me@gmail.com>

### DIFF
--- a/src/guidellm/scheduler/worker.py
+++ b/src/guidellm/scheduler/worker.py
@@ -11,6 +11,7 @@ status updates throughout request processing.
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import time
 import traceback
 from collections.abc import AsyncIterator
@@ -238,6 +239,10 @@ class WorkerProcess(Generic[RequestT, ResponseT]):
                 poll_interval=self.messaging.poll_interval,
             )
             processing_task.cancel()
+            # Let in-flight nodes finish reporting their own terminal update
+            # before the sweep below, so a node is never reported twice.
+            with contextlib.suppress(asyncio.CancelledError):
+                await processing_task
 
             # 4. Cancel pending requests until proc canceled (manual, shutdown, error)
             await self._cancel_requests_loop()
--- a/src/guidellm/scheduler/worker_group.py
+++ b/src/guidellm/scheduler/worker_group.py
@@ -730,38 +730,27 @@ class WorkerGroupState(Generic[RequestT, ResponseT]):
             self._state.queued_requests = len(self._queued_request_ids)
             self._state.created_requests += 1
         elif info.status == "pending":
+            if info.request_id not in self._queued_request_ids:
+                # Already finalized by a racing terminal update; ignore.
+                return
             self._queued_request_ids.remove(info.request_id)
             self._state.queued_requests = len(self._queued_request_ids)
             self._pending_request_ids.add(info.request_id)
             self._state.pending_requests = len(self._pending_request_ids)
         elif info.status == "in_progress":
+            if info.request_id not in self._pending_request_ids:
+                # Already finalized by a racing terminal update; ignore.
+                return
             self._pending_request_ids.remove(info.request_id)
             self._state.pending_requests = len(self._pending_request_ids)
             self._processing_request_ids.add(info.request_id)
             self._state.processing_requests = len(self._processing_request_ids)
         elif info.status == "first_token":
             pass
-        elif info.status == "completed":
-            info.timings.finalized = finalized
-            self._processing_request_ids.remove(info.request_id)
-            self._state.processing_requests = len(self._processing_request_ids)
-            self._state.processed_requests += 1
-            self._state.successful_requests += 1
-        elif info.status in ("errored", "cancelled"):
-            info.timings.finalized = finalized
-            if info.request_id in self._queued_request_ids:
-                self._queued_request_ids.remove(info.request_id)
-                self._state.queued_requests = len(self._queued_request_ids)
-            elif info.request_id in self._pending_request_ids:
-                self._pending_request_ids.remove(info.request_id)
-                self._state.pending_requests = len(self._pending_request_ids)
-            elif info.request_id in self._processing_request_ids:
-                self._processing_request_ids.remove(info.request_id)
-                self._state.processing_requests = len(self._processing_request_ids)
-
-            self._state.processed_requests += 1
-            self._state.errored_requests += 1 if info.status == "errored" else 0
-            self._state.cancelled_requests += 1 if info.status == "cancelled" else 0
+        elif info.status in ("completed", "errored", "cancelled"):
+            if not self._finalize_terminal_request(info, finalized):
+                # Already finalized by a racing terminal update; ignore.
+                return
         else:
             raise ValueError(f"Unknown request_info status {info.status} for {info}")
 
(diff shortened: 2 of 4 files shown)


sjmonson (approved):
Yeah overall seems good, maybe some cleanup is in order for `_update_state_request_counts` but we can take that as a TODO for later assuming no one else wants to raise it now.

dbutenhof (approved):

jaredoconnell (commented):
Looks good to me. The AI review indicated that this does make counts idempotent, but it says that constraints are not fully idempotent yet. But this can wait, since it's an improvement.

> _locked_update still runs _update_with_constraints after _update_state_request_counts ignores a duplicate. MaxErrorRateConstraint will append a second error_window sample for that request. Returning a bool from the count update and skipping constraints on ignore would make the safety net complete.

AmirF194:
Thanks for the reviews. Good point on the constraints ordering, agreed a bool return from the count update to gate the constraints call would close that gap cleanly. Happy to take that as a follow-up if nobody beats me to it.

PR #1067, opened 2026-08-31
feat: add OrcaRouter HTTP backend

nissrin2020ali-ux:
## Summary

GuideLLM lets teams benchmark real-world LLM workloads against OpenAI-compatible endpoints — but until now, benchmarking a hosted gateway meant treating it as an anonymous custom `openai_http` base URL. This PR adds **orcarouter_http** as a first-class backend, mirroring how GuideLLM already treats OpenAI-compatible HTTP servers, so users of the platform can point a benchmark at [OrcaRouter](https://www.orcarouter.ai) directly and get the gateway's full provider/model namespace out of the box.

## Details

OrcaRouter is an OpenAI-compatible AI gateway built for both models and agents. Like OpenRouter, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding orcarouter as a first-class provider means this project's users can use that stack directly, without treating OrcaRouter as an anonymous custom base URL. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

Changes:

- New `orcarouter_http` backend kind registered in the `Backend` registry, mirroring the existing `openai_http` implementation (`src/guidellm/backends/orcarouter/http.py`).
- New `OrcaRouterHTTPBackendArgs` schema that extends `OpenAIHTTPBackendArgs`, defaulting `target` to `https://api.orcarouter.ai` and documenting `orcarouter/*` and provider-prefixed model identifiers (`src/guidellm/schemas/backends/orcarouter_http.py`).
- Specialized connectivity validation: OrcaRouter exposes no `/health` route, so `validate()` probes the `/v1/models` endpoint instead.
- Default model of `orcarouter/auto` when none is configured — OrcaRouter routes each request to the best provider for the workload.
- Re-exports from `guidellm.backends` and `guidellm.schemas.backends`, plus unit tests and a docs entry in `docs/guides/backends.md`.

## Test Plan

- `ruff check`, `ruff format --check`, and `mypy` pass on all changed files.
- All backend unit tests pass (`tests/unit/backends`, 516 tests), including 11 new tests for `orcarouter_http`.
- Import-linter contracts pass (`lint-imports`).
- Live-tested against the OrcaRouter API using a real key: `Backend.create` → `process_startup` → `validate` (200 on `/v1/models`) → `resolve` returned a chat completion from `orcarouter/auto`.

```bash
guidellm run \
  --backend kind=orcarouter_http,api_key=sk-orca-...,model=orcarouter/auto \
  --data kind=synthetic_text,prompt_tokens=256,output_tokens=128
```

I'm an engineer on the OrcaRouter team.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

---

## Use of AI

---

# git log

commit e95d4c26d0e74238fdbd16bcc5eec7251d4f6fe3
Author: nissrin2020ali-ux <nissrin2020ali-ux@users.noreply.github.com>
Date:   Mon Aug 31 13:02:51 2026 +0000

    feat: add OrcaRouter HTTP backend
    
    Add a first-class `orcarouter_http` backend that mirrors the existing
    `openai_http` backend and targets the OrcaRouter API, an OpenAI-compatible
    AI gateway. OrcaRouter exposes a provider/model namespace across many models
    with adaptive routing and automatic failover, so users can benchmark through
    it directly instead of configuring an anonymous custom base URL.
    
    The backend reuses the OpenAI-compatible request machinery, specializes the
    validation probe to the `/v1/models` endpoint (OrcaRouter has no `/health`
    route), and defaults the model to `orcarouter/auto` when none is configured.
    Documentation and unit tests are included.
    
    Generated-by: Claude
    Signed-off-by: nissrin2020ali-ux <nissrin2020ali-ux@users.noreply.github.com>

---------

Generated-by: Claude
Signed-off-by: nissrin2020ali-ux <nissrin2020ali-ux@users.noreply.github.com>

### DIFF
--- a/docs/guides/backends.md
+++ b/docs/guides/backends.md
@@ -36,6 +36,24 @@ GuideLLM supports running inference in the same process using the **vLLM Python 
 
 The **vLLM Python batch backend** (`vllm_python_batch`) uses vLLM's synchronous `LLM` engine for batch-oriented inference. Requests are queued and dispatched in configurable batches, removing per-request scheduling overhead. This is ideal for throughput benchmarking. For setup and examples, see [vLLM Python batch backend](vllm-python-batch-backend.md).
 
+### OrcaRouter
+
+[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible AI gateway that exposes a provider/model namespace across many models. GuideLLM supports it as a first-class backend (`orcarouter_http`), so you can benchmark through OrcaRouter without treating it as an anonymous custom base URL. It defaults to the `https://api.orcarouter.ai` endpoint and the `orcarouter/auto` model, which routes each request to the best provider for the workload:
+
+```bash
+guidellm run \
+  --backend kind=orcarouter_http,api_key=sk-orca-...,model=orcarouter/auto \
+  --data kind=synthetic_text,prompt_tokens=256,output_tokens=128
+```
+
+Or with JSON:
+
+```bash
+--backend '{"kind":"orcarouter_http","api_key":"sk-orca-...","model":"orcarouter/auto"}'
+```
+
+The backend accepts the same parameters as `openai_http` (`request_format`, `stream`, `timeout`, `extras`, and more).
+
 ## Examples for Spinning Up Compatible Servers
 
 ### 1. vLLM
--- a/src/guidellm/backends/__init__.py
+++ b/src/guidellm/backends/__init__.py
@@ -19,6 +19,7 @@ from .openai import (
     OpenAIWebSocketBackend,
     TextCompletionsRequestHandler,
 )
+from .orcarouter import OrcaRouterHTTPBackend
 from .vllm_python import (
     VLLMPythonAsyncBackend,
     VLLMPythonBatchBackend,
@@ -33,6 +34,7 @@ __all__ = [
     "OpenAIRequestHandler",
     "OpenAIRequestHandlerFactory",
     "OpenAIWebSocketBackend",
+    "OrcaRouterHTTPBackend",
     "TextCompletionsRequestHandler",
     "VLLMPythonAsyncBackend",
     "VLLMPythonBatchBackend",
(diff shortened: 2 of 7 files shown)


jaredoconnell (commented):
I like to see that it's reusing a lot of the existing HTTP backend, but I'm curious what made it so that separating it out was worth it. The main things I see is you overrode the validation method and added a new default model.

sjmonson (commented):
Yeah aside from offering some offering some better defaults for this use-case I don't see the reason to have this as a separate backend.

nissrin2020ali-ux:
Thanks @jaredoconnell and @sjmonson for taking a look.

To answer the "why a separate backend" question directly: the request path is 100% reused. `OrcaRouterHTTPBackend` subclasses `OpenAIHTTPBackend` and adds no new request/response/streaming code. The only real divergences are the two places the base class hard-codes, and both are API differences rather than cosmetic defaults:

1. **Validation endpoint.** `OpenAIHTTPBackend.validate()` health-checks `GET {target}/health` (via `api_routes['/health']`). OrcaRouter exposes no `/health` route, so pointing a plain `openai_http` backend at the OrcaRouter target fails `validate()` out of the box — the workaround would be `validate_backend=False`, which silently disables the connectivity check. The subclass probes `GET {target}/v1/models` instead, the same route `available_models()` already uses.

2. **Default model when none is configured.** `OpenAIHTTPBackend.default_model()` falls back to `available_models()[0]` when no model is set. On a multi-provider routing gateway, that picks an arbitrary first entry from the model namespace rather than the intended routing entrypoint. `orcarouter/auto` is a deterministic default that exercises the gateway's routing.

The new args schema exists mainly to give those two defaults a home (default target + default model); everything else (streaming, request formats, extras, timeouts, auth) is inherited unchanged. That's why the delta is small — it is, by design.

If you'd still rather not add a backend kind for this, I'm happy to instead make those two hard-coded bits configurable on `openai_http` (a configurable validation route and an optional default-model override), and document OrcaRouter as a config example in the backends guide. Let me know which approach you'd prefer.

PR #1071, opened 2026-08-31
WEKA Subagent Support + Tool Calls

### LINKED ISSUE #992: Implement full subagent WEKA trace replay
Complete WEKA format replay with full tool calling and branching conversation support.

Some references:

* https://github.com/kubernetes-sigs/inference-perf/blob/main/docs/weka_trace_replay.md
* https://github.com/ai-dynamo/aiperf/pull/1070

jaredoconnell:
## Summary

Adds support for weka subagents and tool calls

## Details

- Adds WEKA subagent support
- Adds tool call support to WEKA implementation.
  - Tool calls do not include all of the info needed to reconstruct the tool call output, so the synthetic design has been used here.

## Test Plan

For subagents, you can test it with this huggingface dataset: https://huggingface.co/datasets/semianalysisai/cc-traces-weka-062126-256k
For tool calling, you can test it with this huggingface dataset: https://huggingface.co/datasets/semianalysisai/cc-traces-weka-042026 

For now you will need to download the file, and reference it.

Here is the command used to test it:
`
guidellm run     --backend "kind=openai_http,target=http://localhost:8000,request_format=/v1/chat/completions,validate_backend=false,model=Qwen/Qwen3-0.6B"     --tokenizer "kind=huggingface_auto,model=Qwen/Qwen3-0.6B"     --profile kind=replay  --constraint "kind=max_requests,count=150" --data "kind=weka,path=weka_traces.jsonl" --constraint kind=max_duration,seconds=240
`

## Related Issues

- Resolves #992 

---

## Use of AI

---

# git log

commit 45bd6f231c1ddc4d6313fa69e35af4aaeca218b6
Date:   Wed Sep 2 14:42:35 2026 -0400

    Added subagent support for WEKA traces
    
commit a389a2bddd9b40f53277a96fc047f9f396a64948
Date:   Wed Sep 2 14:42:35 2026 -0400

    Add tool call support to WEKA traces
    
    A lot of the data for the tool call response is not present in the trace, so the synthetic design is used in those places.
    
---------

### DIFF
--- a/src/guidellm/data/deserializers/trace_common.py
+++ b/src/guidellm/data/deserializers/trace_common.py
@@ -126,6 +126,8 @@ def create_distinct_token_block(
 
 
 class TraceFormatBase(Protocol):
+    config: TraceDataArgs
+
     def __init__(self, config, dataset: Dataset) -> None: ...
 
     def __iter__(self) -> Iterable[Dataset]:
@@ -144,11 +146,73 @@ class TraceFormatBase(Protocol):
         """Called during initialization, immediately after doing
         format-agnostic validation."""
 
+    def validate_conversation(self, conversation: Dataset) -> None:
+        """Validate one conversation yielded by ``__iter__``.
+
+        The default treats every row as an API request with timestamp and
+        token-count columns. Formats with mixed row types (for example WEKA
+        subagent groups) should override this.
+        """
+        features = Features(
+            {
+                self.config.timestamp_column: Value("float"),
+                self.config.prompt_tokens_column: Value("int32"),
+                self.config.output_tokens_column: Value("int32"),
+                **dict(self.required_columns()),
+            }
+        )
+        if self.config.conversation_id_column in features:
+            features.pop(self.config.conversation_id_column)
+        _raise_if_nonetype_found(conversation, features)
+        _raise_if_incorrect_types(conversation, features)
+        for row in conversation:
+            _validate_row(row, self.config)
+            self.validate_row(row)
+
     def create_prompt(
         self, row: dict, processor: PreTrainedTokenizerBase, faker: Faker
     ) -> str:
         """Called within `trace_common.TraceExamplesIterable` on each iteration.
         Returns a generated synthetic prompt."""
+
+    def build_conversation_graph(
+        self,
+        conversation: Dataset,
+        processor: PreTrainedTokenizerBase,
+        faker: Faker,
+    ) -> ConversationGraphData:
+        """Build a conversation graph from one ``__iter__`` conversation.
+
+        The default emits a linear ``main_*`` chain. Formats with branches
+        or subagents should override this rather than branching in the
+        shared iterable.
+        """
+        start_ts = conversation[0][self.config.timestamp_column]
+        turns = []
+        for turn_idx, turn in enumerate(conversation):
+            parents = []
+            if turn_idx > 0:
+                parents.append(
+                    ConversationParentRef(parent_node_id=f"main_{turn_idx - 1}")
+                )
+
+            prompt = self.create_prompt(turn, processor, faker)
+            relative_timestamp = turn[self.config.timestamp_column] - start_ts
+            columns = {
+                "text_column": [prompt],
+                "prompt_tokens_count_column": [turn[self.config.prompt_tokens_column]],
+                "output_tokens_count_column": [turn[self.config.output_tokens_column]],
+                "relative_timestamp_column": [relative_timestamp],
+            }
+            turns.append(
+                ConversationTurnData(
+                    node_id=f"main_{turn_idx}",
+                    agent_id="default",
+                    parents=parents,
+                    columns=columns,
+                )
+            )
+        return ConversationGraphData(turns=turns)
 
 
 class TraceFormatRegistry(RegistryMixin[type[TraceFormatBase]]):
@@ -192,53 +256,21 @@ class TraceExamplesIterable(_BaseExamplesIterable):
             time_scale=self.config.time_scale,
         )
         for conv in self.format:  # type: ignore[attr-defined]
-            start_ts = conv[0][self.config.timestamp_column]
-            row = self._create_conversation_row(conv, start_ts, timing)
-            samples_count += len(conv)
-            yield samples_count, row
+            graph_data = self.format.build_conversation_graph(
+                conv, self.processor, self.faker
+            )
+            timing.apply(graph_data)
+            samples_count += len(graph_data.turns)
+            payload = json.dumps(graph_data.model_dump(mode="json"))
+            yield (
+                samples_count,
+                {
+                    "conversation_turns": (
+                        payload.decode() if isinstance(payload, bytes) else payload
+                    )
+                },
+            )
             self.format.reset()
-
-    def _create_conversation_row(
-        self,
-        conversation: Dataset,
-        start_ts: float,
-        timing: TraceSessionTiming,
-    ) -> dict[str, Any]:
-        """Build a ``conversation_turns`` payload for linear or branched graphs."""
-        turns = []
-        for turn_idx, turn in enumerate(conversation):
-            parents = []
-            if turn_idx > 0:
-                parents.append(
-                    ConversationParentRef(parent_node_id=f"main_{turn_idx - 1}")
-                )
-
-            # TODO: Branches & subagents
-
-            prompt = self.format.create_prompt(turn, self.processor, self.faker)
-            relative_timestamp = turn[self.config.timestamp_column] - start_ts
-            columns = {
-                "text_column": [prompt],
-                "prompt_tokens_count_column": [turn[self.config.prompt_tokens_column]],
-                "output_tokens_count_column": [turn[self.config.output_tokens_column]],
-                "relative_timestamp_column": [relative_timestamp],
-            }
-            turns.append(
-                ConversationTurnData(
-                    node_id=f"main_{turn_idx}",
-                    agent_id="default",
-                    parents=parents,
-                    columns=columns,
-                )
-            )
-        graph_data = ConversationGraphData(turns=turns)
-        timing.apply(graph_data)
-        payload = json.dumps(graph_data.model_dump(mode="json"))
-        return {
-            "conversation_turns": (
-                payload.decode() if isinstance(payload, bytes) else payload
-            )
-        }
 
     @property
     def is_typed(self) -> bool:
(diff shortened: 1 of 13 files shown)


src/guidellm/scheduler/worker_group.py
```diff
@@ -697,6 +710,16 @@ def received_callback(
...
 
+    def recheck_constraints(self) -> _StateUpdate:
+        """Re-evaluate constraints from scheduler state without a request update.
+
+        Used by the coordinator poll loop so time-based constraints such as
+        ``max_duration`` can fire while workers are sleeping.
+
+        :return: Copied scheduler state and stop flags
+        """
+        return self._locked_update(info=None)
```
dbutenhof: The `info=None` seems redundant since that's the default. While it serves a minor documentation purpose, you don't similarly make `add_constraints=None` explicit, and the "half-way" state just makes the choice a bit distracting.

For that matter, although I get the "drop the '_' because it's referenced outside the class", you could have just as easily renamed `_locked_update` without losing anything. (Although I like the docstring...)

src/guidellm/scheduler/worker_group.py
```diff
@@ -751,11 +775,17 @@ def _update_state_request_counts(self, info: RequestInfo):
...
 
+        self._update_request_time_bounds(info, finalized)
+
+    def _update_request_time_bounds(self, info: RequestInfo, finalized: float) -> None:
         # Keep global count of the earliest start and latest end
-        self._state.start_requests_time = min(
-            info.timings.request_start or float("inf"),
-            self._state.start_requests_time or float("inf"),
-        )
+        if info.timings.request_start is not None:
```
dbutenhof, line 807: Before, if `info.timings.request_start` was `None`, `start_requests_time` was set to `info.timings.request_start or float("inf")` ... now it's unchanged. Is that an oversight, or intentional?

dbutenhof (commented):
First commit: question & minor comment...

src/guidellm/benchmark/profiles/replay.py
```diff
@@ -28,7 +28,10 @@ class ReplayProfile(Profile):
     Each request is scheduled at
     ``start_time + time_scale * relative_timestamp`` via ``RequestSettings`` on
     the dataset finalizer output. For this profile, ``rate`` is interpreted as
-    ``time_scale`` (not requests per second).
+    ``time_scale`` (not requests per second). ``max_wait`` clamps gaps inside
```
dbutenhof: This extends back outside the change bar ... I guess I missed this when I did the late-0.7 CLI cleanup (Likely because this is the only case where that aliasing somehow leaked over into the `Profile` vs just the `ProfileArgs`...)

The "``rate`` is interpreted as ``time_scale`` (not requests per second)." is meaningless now and should be removed. If you don't want to do it here, I can put up a separate PR "one of these days".

src/guidellm/scheduler/worker_group.py
jaredoconnell: It's intentional. Apparently setting it to infinite would make it use infinite as the origin, that it made the UI report it as warmup forever.
Now with it being None it will fall back to start.start_time.

src/guidellm/scheduler/worker_group.py
jaredoconnell: Sure. I renamed the function to an appropriate non-internal name. I also removed the explicit None inputs.

src/guidellm/data/deserializers/trace_common.py
```diff
@@ -143,12 +145,65 @@ def validate_row(self, row: dict) -> None:
...
+            self.validate_row,
+        )
+
     def create_prompt(
         self, row: dict, processor: PreTrainedTokenizerBase, faker: Faker
     ) -> str:
         """Called within `trace_common.TraceExamplesIterable` on each iteration.
         Returns a generated synthetic prompt."""
 
+    def build_conversation_graph(
```
dbutenhof, line 169: Sadly, here's where "review by commit" breaks down. I got to the 3rd and found myself trying to follow call trees in my pulled branch ... for a bunch of support functions used here that you ended up removing. (Which is great, since we seemed to be doing multiple traversals through the tree whereas it seems you've combined them here.)

dbutenhof (commented):
2nd drop ... one of these is from the start of a review on the 3rd commit (subagent support): but I got lost trying to track some helper calls you later removed, and gave up. I'll start a new pass with all commits in scope, since I can't review the 3rd in isolation (without wasting a lot of time on transient artifacts), and I don't want to just skip it either.

Not really a complaint ... just the nature of the beast sometimes.

src/guidellm/benchmark/profiles/replay.py
jaredoconnell: Just to lower churn here I'll leave it as a follow up.

src/guidellm/scheduler/worker_group.py
```diff
@@ -404,7 +404,20 @@ async def request_updates(
...
+                        state_update.stop_processing
+                        and self.constraint_reached_event is not None
+                        and not self.constraint_reached_event.is_set()
+                    ):
+                        self.constraint_reached_event.set()
+                    if state_update.stop_queueing:
+                        self.state.stop_send_requests_event.set()
                 if self.shutdown_event.is_set():  # type: ignore[union-attr]
                     # Everything yielded, exit
                     break
```
sjmonson: Shouldn't it check shutdown event first?

src/guidellm/scheduler/strategies.py
```diff
@@ -764,5 +815,136 @@ async def resolve_dequeued_target_start(
...
+            return
+        unscaled_trim = scaled_trim / self.time_scale
+        for node in graph.nodes.values():
+            relative_timestamp = node.settings.relative_timestamp
+            if relative_timestamp is None:
+                continue
+            node.settings = node.settings.model_copy(
+                update={"relative_timestamp": relative_timestamp - unscaled_trim}
+            )
+
```
sjmonson: I **really really really** don't like this being done in the scheduler. Most of this could be handled on the dataset creation itself with the exception of `max_session_wait`.

sjmonson (changes requested):
I think my comments on the former half of this PR are going to be very different to the latter half. Could you move the first two commits of this PR to their own PR to make it a little easier to discuss.

src/guidellm/scheduler/strategies.py
jaredoconnell: Moved to the data side on another PR. I documented some trade-offs in the PR. https://github.com/vllm-project/guidellm/pull/1081

src/guidellm/scheduler/worker_group.py
jaredoconnell: Fixed in the other PR: https://github.com/vllm-project/guidellm/pull/1080/

src/guidellm/utils/hf_datasets.py
```diff
@@ -3,21 +3,87 @@
...
+    :return: Dataset with one row per JSON record.
+    :raises ValueError: If the file cannot be parsed or converted.
+    """
+    path = Path(path)
+    records = _read_json_records(path)
+    try:
+        return Dataset.from_list(records)
+    except (TypeError, ValueError) as exc:
+        raise ValueError(f"Failed to load JSON dataset from '{path}': {exc}") from exc
+
```
sjmonson, line 75: This kinda defeats the point of using `datasets` since it is reading the entire jsonl file manually. Would normally be a blocker except that we designed to handle it in a follow-up.

src/guidellm/utils/hf_datasets.py
```diff
@@ -3,21 +3,87 @@
...
+        if prefix == "[":
+            try:
+                data = json.loads(handle.read())
+            except json.JSONDecodeError as exc:
+                raise ValueError(f"Invalid JSON in '{path}': {exc}") from exc
+            if not isinstance(data, list):
+                raise ValueError(
+                    f"JSON file '{path}' must contain a list of records or JSON Lines."
+                )
+            return data
```
dbutenhof, line 41: So effectively it seems we have either a single line list of objects (though we're not checking the element type) or a JSONL of objects?

And `json.loads` will already ignore whitespace ... it seems wasteful to read by character to find the hypothetical `[` instead of just loading the line.

Although it'd be unlikely (and arguably incorrect) to read a JSONL with both lists and objects, you could easily absorb that by doing a `.extend` for a list type line, and avoid a lot of this extra code. (If desirable, you could even fail out if you find a list type after reading at least one object, or vice versa.)

src/guidellm/data/deserializers/trace_common.py
```diff
@@ -318,6 +341,35 @@ def _validate_path(path: Path) -> None:
         raise DataNotSupportedError(f"Trace file is empty: {path}")
 
 
+def _validate_api_conversation(
```
dbutenhof, line 344: Since you're calling this from outside `trace_common.py`, I don't think this should have a dunder name.

Reading through this, I do find myself getting annoyed with the weird mix of class/instance helpers and module level helpers. I'm not even sure this is a rational annoyance, since OO helpers can still be scattered across modules... but dunder names definitely shouldn't, either way.

src/guidellm/data/deserializers/trace_weka.py
```diff
@@ -174,3 +351,268 @@ def create_prompt(
...
+                    random_seed=faker.random.getrandbits(32),
+                )
+            )
+        body = _generate_remaining_prompt(
+            next(self._tool_response_sampler), processor, faker
+        )
+        raw = json.dumps({"result": body})
+        return raw.decode() if isinstance(raw, bytes) else raw
+
+    def _unpack_conversation(
```
dbutenhof, line 434: I'd like to see a docstring describing exactly what this "unpack" means, especially with reference to the "magic indices" 0 and "_weka_index".

src/guidellm/data/deserializers/trace_weka.py
```diff
@@ -174,3 +351,268 @@ def create_prompt(
...
+        api_rows: list[dict[str, Any]] = []
+        columns = (
+            self.config.timestamp_column,
+            self.config.prompt_tokens_column,
+            self.config.output_tokens_column,
+            self.config.hash_ids_column,
+        )
+        for row in requests:
+            if _is_subagent_entry(row):
+                api_rows.extend(self._collect_api_rows(list(row.get("requests") or [])))
```
dbutenhof, line 450: Do we really need to make a recursive call if `row.get("requests")` is empty?

src/guidellm/data/deserializers/trace_weka.py
```diff
@@ -174,3 +351,268 @@ def create_prompt(
         if not remainder:
             return prompt
         return f"{prompt} {remainder}"
+
+    def build_conversation_graph(
```
dbutenhof, line 355: I find myself wondering how much of this we'll be refactoring to generalize for OTEL, and whether we could make that easier now. I tell myself it's more efficient to not worry about it and refactor when we know more, but ...

sjmonson (approved):
The dataset loading and validation is a bit of an issue from a start performance perspective but this isn't a new problem to the trace datasets. To avoid blocking the WEKA support I'll approve this now and followup with some loading optimization.

src/guidellm/data/deserializers/trace_weka.py
```diff
@@ -174,3 +351,268 @@ def create_prompt(
...
+
+        if pending_join_ids:
+            logger.debug(
+                "WEKA subagent(s) in conversation '{}' on agent '{}' have no "
+                "following parent turn; replaying as background without a join",
+                conversation_id,
+                agent_id,
+            )
+
+        self._warn_chain_overlap(conversation_id, agent_id, chain_events)
```
dbutenhof, line 571: Can we do this as we build the chain instead of traversing it again?

dbutenhof (approved):
Complicated to review. I don't see anything wrong, but I've accumulated a bunch of nits and some potential efficiency comments. On the other hand, I see that Sam already approved with a note on the efficiency, so I'm going to break here for the day and just submit what I've got for consideration.

I _think_ we don't auto-merge with open conversations so I'm going to approve in the interests of being able to move forward, should you feel like just resolving the conversations... none are show-stoppers.

sjmonson:
@mergifyio queue

PR #1012, opened 2026-08-07
Switch current trace formats to use a DAG

SkiHatDuckie:
## Summary
Restructures all preexisting trace formats to work with the DAG, switching to a non-flattened approach to dataset validation and grabbing conversations. Done as a step towards full WEKA trace format support ([#992](https://github.com/vllm-project/guidellm/issues/992)). I'll make a PR in the future if I find the time to finish the WEKA trace replay, containing the features still missing for full support.

The following still needs to be implemented for WEKA:
- Subagent conversations for WEKA
- Handling input/output types and stop reason columns for added context when making nodes
- Nonlinear histories (context/token collapse -> hash-ids becoming incorrect mid-conversation)
- Tool call events

## Details
Current (changed) approach for trace formats:
- JSON load the whole dataset. Yes it's probably inefficient, but it makes the rest of the validation chain significantly simpler. The processes are also self-contained, which should make optimizations easier.
- Added a `conversation_locations` field to TraceFormatBase. This is a list of a list of ints which point to the start of every conversation. Each integer acts as an index to somewhere in the dataset. The number of ints indicates depth. It's the responsibility of the format to generate these locations and interpret them.
- Finding required columns has been moved to formats. The benefit is that instead of searching for required columns with no context, formats can instead just search in the place where the columns are supposed to be.
- Conversations are served as an `Iterable[Dataset]` to abstract away getting the next conversation via conversation locations, which trace_common will not know how to interpret.

Other
- Repaired tests
- Switched Minimal, Mooncake, and partial WEKA to using the DAG format instead of the flattened datasets
- Removed VirtualColumnLocation & helpers, as they've been replaced by format-specific handling
- Removed format-agnostic JSON unwrapping. I overall regretted adding this feature, as it added the responsibility of needing to support ad-hoc trace formats and format variants.

## Test Plan
NOTE: This PR has not been fully tested for performance in comparison to the current setup.

python -m pytest tests/unit/data/deserializers/
tox -e test-unit
tox -e test-integration
tox -e lint-fix && tox -e type-check

## Related Issues

- Related to [#992](https://github.com/vllm-project/guidellm/issues/992)

---

- [x] "I certify that all code in this PR is my own, except as noted below."

## Use of AI

- [ ] Includes code generated or substantially modified by an AI agent
- [ ] Includes tests generated or substantially modified by an AI agent

[...]

PR #1072, opened 2026-08-31
fix: validate only the required columns in trace files

QHarshil:
## Summary

Trace validation casts each conversation to the required feature set and scans every column for nulls. `Dataset.cast` requires the column sets to match, so a trace carrying any column beyond the required ones is rejected before the benchmark starts. A null in a column the deserializer never reads fails the same way.

Real traces carry that metadata. The WEKA specification defines `model`, `type`, `input_types`, `output_types` and `stop` on every request record, plus optional `api_time`, `ttft` and `think_time`. A trace that follows it cannot be loaded:

```
DataNotSupportedError: The columns in features (['t', 'in', 'out', 'hash_ids'])
must be identical as the columns in the dataset:
['t', 'type', 'model', 'in', 'out', 'hash_ids', 'input_types', 'output_types', 'stop']
```

The reference trace from the specification fails on this, and so does `semianalysisai/cc-traces-weka-no-subagents-051226`, which `docs/guides/trace_replay.md` links to as an example. That dataset carries five of those fields, and 471 of its 949 conversations omit `ttft` or `think_time` on at least one record, which trips the null scan.

The guide already describes the intended behavior: `trace_synthetic` looks only at the timestamp, prompt token count and output token count columns, "ignoring all other features contained in a dataset". This change restores that for all three formats. Traces containing only the required columns are unaffected.

Before #1012 the validation loop ran per required column, and its docstring described the contract as "A required column contains a NoneType" and "A required column failed during cast to feature type". #1012 widened both checks to the whole dataset.

This does not implement nested WEKA subagent replay. With this fixed, the canonical subagent trace gets past top-level schema validation and then fails on the nested records, which is tracked by #992.

## Details

- [x] `_raise_if_incorrect_types` selects the required columns before casting, so unrelated columns are neither constrained nor copied
- [x] a shape mismatch in a required column raises `TypeError` rather than `ValueError` and previously escaped the deserializer uncaught, so both are now converted to `DataNotSupportedError`
- [x] `_raise_if_nonetype_found` takes the required features and checks only those columns
- [x] tests that extra columns and nulls outside the required columns are accepted, and that a null, a bad scalar type or a bad shape in a required column is still rejected

## Test Plan

- `tox -e lint-check` and `tox -e type-check`: clean
- `tox -e tests -- tests/unit/data/deserializers/`: 200 passed, 1 skipped
- `tox -e test-unit`: 2936 passed, excluding 14 `tests/unit/utils/test_audio.py` failures from a missing local FFmpeg library that fail identically on `main`
- `tox -e test-integration`: 27 passed
- `tox -e test-e2e`: 13 passed, with the same local FFmpeg error on `main`
- checked all three registered formats: `trace_synthetic`, `mooncake` and `weka` traces carrying extra columns are rejected on `main` and load on this branch
- loaded the reference `sample_trace.json` from the specification and the first 30 conversations of `cc-traces-weka-no-subagents-051226`, both of which fail on `main`

## Related Issues

- Prerequisite for #992

---

## Use of AI

---

# git log

commit 08e5a77801fc1dc58b4a8ceb61c4ff5c01e9ffb9
Date:   Mon Aug 31 13:36:36 2026 -0700

    fix: validate only the required columns in trace files
    
    Trace validation cast each conversation to the required feature set and
    scanned every column for nulls. Dataset.cast requires the column sets to
    match, so a trace carrying any extra column was rejected, and a null in a
    column the deserializer never reads failed validation the same way.
    
    The WEKA specification defines model, type, input_types, output_types and
    stop on every request record, alongside optional api_time, ttft and
    think_time, so no trace that follows it can be loaded. The reference trace
    from the specification fails on this, as does the
    semianalysisai/cc-traces-weka-no-subagents-051226 dataset linked from
    docs/guides/trace_replay.md, which carries five of those fields. That guide
    already describes trace_synthetic as ignoring all other features contained
    in a dataset.
    
    Select the required columns before casting, and check only those for nulls,
    so unrelated columns are ignored. A shape mismatch in a required column
    surfaces as TypeError rather than ValueError, and previously escaped the
    deserializer uncaught, so convert both to DataNotSupportedError.
    
    Generated-by: Claude Code claude-opus-5

---------

Generated-by: Claude Code claude-opus-5

### DIFF
--- a/src/guidellm/data/deserializers/trace_common.py
+++ b/src/guidellm/data/deserializers/trace_common.py
@@ -316,16 +316,16 @@ def _validate_row(row: dict, config: TraceDataArgs) -> None:
         )
 
 
-def _raise_if_nonetype_found(dataset: Dataset) -> None:
-    for col in dataset.column_names:
+def _raise_if_nonetype_found(dataset: Dataset, features: Features) -> None:
+    for col in features:
         if dataset.data[col].null_count != 0:
             raise DataNotSupportedError(f"Missing column values in {col}")
 
 
 def _raise_if_incorrect_types(dataset: Dataset, features: Features) -> None:
     try:
-        dataset.cast(features)
-    except ValueError as e:
+        dataset.select_columns(list(features)).cast(features)
+    except (TypeError, ValueError) as e:
         raise DataNotSupportedError(str(e)) from e
 
 
@@ -343,7 +343,7 @@ def _validate_dataset(config: TraceDataArgs, trace_format: TraceFormatBase) -> N
             raise DataNotSupportedError("Trace conversation is empty")
         if config.conversation_id_column in features:
             features.pop(config.conversation_id_column)
-        _raise_if_nonetype_found(conv)
+        _raise_if_nonetype_found(conv, features)
         _raise_if_incorrect_types(conv, features)
         for row in conv:
             _validate_row(row, config)
--- a/tests/unit/data/deserializers/test_trace_common.py
+++ b/tests/unit/data/deserializers/test_trace_common.py
@@ -220,6 +220,52 @@ class TestTraceDatasetDeserializer:
         for i, turn in enumerate(conv):
             assert turn.columns["relative_timestamp_column"][0] == i
 
+    @pytest.mark.regression
+    def test_accepts_columns_beyond_the_required_ones(
+        self, tmp_path: Path, deserializer
+    ):
+        """Extra trace columns are ignored rather than rejected.
+
+        Real trace files carry per-request metadata (model, type, timings) that
+        the deserializer does not consume.
+
+        ## WRITTEN BY AI ##
+        """
+        trace = write_trace(
+            tmp_path,
+            '{"timestamp": 1, "input_length": 10, "output_length": 1, '
+            '"model": "a", "type": "s", "api_time": 4.87}\n'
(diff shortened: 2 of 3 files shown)


QHarshil:
Some overlap with #1071, which opened just before this one.

I ran the same validation cases against its branch. It fixes the WEKA spec-metadata case as part of the subagent work, by projecting nested API rows down to the required columns before validation. The generic problem is still there: `trace_synthetic` and `mooncake` with extra columns are rejected, and a shape mismatch in a required column still escapes as `TypeError`.

The two conflict in `trace_common.py`, since #1071 moves validation behind a new `TraceFormatBase.validate_conversation()` hook while this changes the shared helpers. If #1071 lands first I am happy to rebase and apply the required-column-only behavior to the default `validate_conversation()`, leaving the WEKA override as it is.

jaredoconnell:
This looks good. I can approve it once CI is passing. It's currently failing.

QHarshil:
@jaredoconnell 
> This looks good. I can approve it once CI is passing. It's currently failing.

Thank you. The two failures don’t seem related to this: the quality job failed fetching the uv manifest before tox ran, and integration failed on the synthetic video timeout.

 I don’t appear to have permission to rerun the upstream Actions run. Could you rerun the failed jobs when you get a chance?

jaredoconnell (approved):

dbutenhof (approved):

PR #1077, opened 2026-09-02
docs: clarify default benchmark outputs

tangming1996:
Clarify that guidellm run writes JSON and CSV files by default, while HTML reports require an explicit output option. Update the README and analysis guide, and correct the YAML output example.

## Details

- [ ]

## Test Plan

-

## Related Issues

- Resolves #

---

## Use of AI

---

# git log

commit 6d0caabc2e6e2e88de804207242330d0eb0c1c9d
Date:   Wed Sep 2 13:48:19 2026 +0800

    docs: clarify default benchmark outputs
    
    Clarify that guidellm run writes JSON and CSV files by default, while HTML reports require an explicit output option. Update the README and analysis guide, and correct the YAML output example.
    
commit b881fe24ce63eec8daffe88884956007973c5857
Date:   Thu Sep 3 10:00:27 2026 +0800

    docs: document benchmark output selection
    
    Explain output override behavior, default result directories, all supported file formats, and console handling across the README and output guides.
    
    Generated-by: Codex GPT-5

commit bbb5bbc10fd11f6df06dfa409a58327ce298ed2c
Date:   Mon Sep 7 09:43:42 2026 +0800

    docs: centralize benchmark output guidance
    
    Replace repeated output configuration examples with links to the output guide. Consolidate supported formats and default filenames, and clarify that path may be omitted or specified explicitly.
    
    Generated-by: Codex

---------

Generated-by: Codex
Generated-by: Codex GPT-5

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,33 +142,54 @@ You will see progress updates and per-benchmark summaries during the run, as giv
 
 ### Inspect Outputs
 
-After the benchmark completes, GuideLLM saves JSON and CSV results into the output directory you specified (default: the current directory). You'll see a summary printed in the console along with the generated file locations. An HTML report is optional and must be requested explicitly with `--output kind=html,path=benchmarks.html`.
+After the benchmark completes, GuideLLM writes `benchmarks.json` and `benchmarks.csv` by default. The files are saved in the directory configured by `GUIDELLM__DEFAULT_RESULTS_DIR`, or in the current directory when the variable is not set. GuideLLM also prints a summary and the generated file locations to the console.
+
+Additional file formats, such as HTML, YAML, and PLOT, must be requested with `--output`. Specifying any `--output` replaces the default JSON and CSV outputs, so repeat the option for every format you want. For example, to keep both defaults and add an HTML report:
+
+```bash
+guidellm run \
+  --backend kind=openai_http,target=http://localhost:8000 \
+  --profile kind=sweep \
+  --constraint kind=max_duration,seconds=30 \
+  --data kind=synthetic_text,prompt_tokens=256,output_tokens=128 \
+  --output kind=json \
+  --output kind=csv \
+  --output kind=html
+```
+
+Each output type supplies a default filename, so `path` is only needed to change the name or destination.
 
 The following section, **Output Files and Reports**, explains what each file contains and how to use them for analysis, visualization, or automation.
 
 ## Output Files and Reports
 
-After running the Quick Start benchmark, GuideLLM writes several output files to the directory you specified. Each one focuses on a different layer of analysis, ranging from a quick on-screen summary to fully structured data for dashboards and regression pipelines.
+The Quick Start benchmark produces console, JSON, and CSV output. Other formats can be selected explicitly. Each format focuses on a different layer of analysis, ranging from a quick on-screen summary to fully structured data for dashboards and regression pipelines.
 
-**Console Output**
+**Console output**
 
-The console provides a lightweight summary with high-level statistics for each benchmark in the run. It's useful for quick checks to confirm that the server responded correctly, the load sweep completed, and the system behaved as expected. Additionally, the output tables can be copied and pasted into spreadsheet software using `|` as the delimiter. The sections will look similar to the following:
+The console provides a lightweight summary with high-level statistics for each benchmark in the run. It's useful for quick checks to confirm that the server responded correctly, the load sweep completed, and the system behaved as expected. Additionally, the output tables can be copied and pasted into spreadsheet software using `|` as the delimiter. Console output is independent of the file formats selected with `--output`; disable it with `--disable-console`. The sections will look similar to the following:
 
 <img alt="Sample GuideLLM benchmark output" src="https://raw.githubusercontent.com/vllm-project/guidellm/main/docs/assets/sample-output.png" />
 
-**benchmarks.json**
+**JSON output (`benchmarks.json` by default)**
 
 This file is the authoritative record of the entire benchmark session. It includes configuration, metadata, per-benchmark statistics, and sample request entries with individual request timings. Use it for debugging, deeper analysis, or loading into Python with `GenerativeBenchmarksReport`.
 
-Alternatively, a YAML version of this file can be generated for easier human readability with the same content as `benchmarks.json` using `--output kind=yaml,path=benchmarks.yaml`.
-
-**benchmarks.csv**
+**CSV output (`benchmarks.csv` by default)**
 
 This file provides a compact tabular view of each benchmark with the fields most commonly used for reporting—throughput, latency percentiles, token counts, and rate information. It opens cleanly in spreadsheets and BI tools and is well-suited for comparisons across runs.
 
-**benchmarks.html** (optional)
+**YAML output (`benchmarks.yaml` by default)**
 
-A self-contained HTML report with charts and tables for throughput and latency (emphasizing P95/P99). The file embeds its own CSS and JavaScript, so it can be shared without network access or a versioned UI dependency.
+This file contains the same detailed benchmark data as the JSON output in a more human-readable format. Generate it with `--output kind=yaml`; it can be produced alongside any other format.
+
+**HTML output (`benchmarks.html` by default)**
+
+This self-contained report includes charts and tables for throughput and latency (emphasizing P95/P99). The file embeds its own CSS and JavaScript, so it can be shared without network access or a versioned UI dependency. Generate it with `--output kind=html`.
+
+**PLOT output (`benchmarks.png` by default)**
+
+This static image contains benchmark performance graphs. Generate it with `--output kind=plot`. The output format is selected by the `path` extension and can be PNG, JPG/JPEG, SVG, or PDF. For example, `--output kind=plot,path=benchmarks.pdf` creates a PDF, and the optional `dpi` parameter controls image resolution.
 
 ## Common Use Cases and Configurations
 
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ You will see progress updates and per-benchmark summaries during the run, as giv
 
 ### Inspect Outputs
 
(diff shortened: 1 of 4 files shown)


docs/getting-started/analyze.md
```diff
@@ -64,11 +64,12 @@ The p99 (99th percentile) values are particularly important for SLO analysis, as
 
 ## Analyzing the Results File
 
-For deeper analysis, GuideLLM saves detailed results to multiple files by default in your current directory:
+For deeper analysis, GuideLLM saves these detailed results by default in your current directory:
```
dbutenhof: I don't think "these" detailed results implies what you think to a reader at first glance. It suggests more specific data rather than the canonical files. Maybe "by default saves detailed results to these files in your current directory".

Also, "current directory" isn't entirely accurate. GuideLLM supports a global settings environment variable (`GUIDELLM__DEFAULT_RESULTS_DIR`) which supplies the default location for all of the output files. That is, `GUIDELLM__DEFAULT_RESULTS_DIR=/var/tmp/results guidellm [...]` will write `/var/tmp/results/benchmarks.json` and `/var/tmp/results/benchmarks.csv` by default. That should probably be explained here.

docs/getting-started/analyze.md
```diff
@@ -64,11 +64,12 @@ The p99 (99th percentile) values are particularly important for SLO analysis, as
...
 
 - `benchmarks.json`: Complete benchmark data in JSON format
 - `benchmarks.csv`: Summary of key metrics in CSV format
-- `benchmarks.html`: Self-contained HTML report with visualizations
+
+To also generate a self-contained HTML report, add `--output kind=html,path=benchmarks.html` to the benchmark command.
```
dbutenhof: There's an unfortunate nuance here: the *default* value for `--output` amounts to `--output kind=json --output kind=csv`: however, specifying any explicit `--output` *overrides* that default, and you'll get only what you request. So `--output kind=html` gives you *only* the HTML output.

To *add* HTML to the "default", you need to also explicitly replace the default. So to "add" HTML, `--output kind=json --output kind=csv --output kind=html`.

(I'm not entirely certain this is really desirable, but the alternative would be to implement something odd like `--no-output kind=json` if you really didn't want it, so practically I think we're stuck with this.)

README.md
```diff
@@ -142,7 +142,7 @@ You will see progress updates and per-benchmark summaries during the run, as giv
 
 ### Inspect Outputs
 
-After the benchmark completes, GuideLLM saves all results into the output directory you specified (default: the current directory). You'll see a summary printed in the console along with a set of file locations (`.json,` `.csv`, `.html`) that contain the full results of the run.
+After the benchmark completes, GuideLLM saves JSON and CSV results into the output directory you specified (default: the current directory). You'll see a summary printed in the console along with the generated file locations. An HTML report is optional and must be requested explicitly with `--output kind=html,path=benchmarks.html`.
```
dbutenhof: Not just HTML, but also the PLOT output. You can list all the current output formats if you want, or maybe just "other output formats are optional". But additionally, if you specify *any* output format, you need to explicitly specify *all* output formats you want: you can no longer rely on defaults.

Additionally, there's no need to specify a `path` unless you want it to be named something other than `benchmarks.<type>` or if you want it to be written somewhere other than the default working directory or the path specified by the `GUIDELLM__DEFAULT_RESULTS_DIR` environment variable. The default file path amounts to `GUIDELLM__DEFAULT_RESULTS_DIR / benchmarks.json` (or csv, html, yaml, etc.)

(The interesting exception is for the `plot` output, where the file type actually selects the format, e.g., `pdf` or `png`. The *default* path is `GUIDELLM__DEFAULT_RESULTS_DIR / benchmarks.png`, which fixes the output format as well as the path.)

README.md
```diff
@@ -160,13 +160,13 @@ The console provides a lightweight summary with high-level statistics for each b
...
 
-Alternatively, a YAML version of this file can be generated for easier human readability with the same content as `benchmarks.json` using `--output yaml "path=benchmarks.yaml"`.
+Alternatively, a YAML version of this file can be generated for easier human readability with the same content as `benchmarks.json` using `--output kind=yaml,path=benchmarks.yaml`.
 
 **benchmarks.csv**
 
 This file provides a compact tabular view of each benchmark with the fields most commonly used for reporting—throughput, latency percentiles, token counts, and rate information. It opens cleanly in spreadsheets and BI tools and is well-suited for comparisons across runs.
 
-**benchmarks.html**
+**benchmarks.html** (optional)
```
dbutenhof: Technically, they're all "optional" -- json and csv are the defaults; but, again, only if you don't specify `--output` at all, which is an important distinction. The " (optional)" here is redundant.

README.md
```diff
@@ -160,13 +160,13 @@ The console provides a lightweight summary with high-level statistics for each b
 
 This file is the authoritative record of the entire benchmark session. It includes configuration, metadata, per-benchmark statistics, and sample request entries with individual request timings. Use it for debugging, deeper analysis, or loading into Python with `GenerativeBenchmarksReport`.
 
-Alternatively, a YAML version of this file can be generated for easier human readability with the same content as `benchmarks.json` using `--output yaml "path=benchmarks.yaml"`.
+Alternatively, a YAML version of this file can be generated for easier human readability with the same content as `benchmarks.json` using `--output kind=yaml,path=benchmarks.yaml`.
```
dbutenhof: And you can generate both if you want, since it's a separate registry item.

I think this should be a separate section (though if you'd prefer to consider that beyond the scope of your PR, that's fine), because it's not "one or the other".

As a more pedantic comment, I think the names of these subsections are poor, since you're describing the "YAML output file" which only *defaults* to `benchmarks.yaml` and can be redirected with `path=whateveriwant.foo`.

The list also wasn't updated when the `plot` output format was added, which is unfortunate. Don't feel obligated to add it, but if you want to be thorough, something like

**benchmarks.png** (or, better, something like "PLOT output file")

An image file containing benchmark performance graphs. The default output path is "benchmarks.png" (PNG image format), but other formats can be specified using the `path` file type: ".png", ".jpg", ".jpeg", ".svg", ".pdf"), for example, `--output kind=plot,path=benchmarks.pdf`. You can also specify the image resolution using the `dpi` parameter, for example, `--output kind=plot,dpi=72,path=benchmarks.jpg`.

docs/getting-started/analyze.md
dbutenhof: "Console" is an interesting special case: you *can* specify `--output kind=console`, but it actually has no effect on the `run` command because it's an "implicit default" -- that is, unlike json and csv, it won't be overridden by specifying an explicit `--output`, and you can only drop console output using the `--disable-console` option. (And you can't really "customize" it, since it has no parameters. 😁)

More importantly, this list also omits the PLOT output type: the list was overlooked when that format was added, unfortunately. If you want, you can add it "while you're in the neighborhood". (I describe that in more detail in comments on your `README.md` changes.)

dbutenhof (changes requested):
Thanks for your interest in helping!

I've written a lot of commentary. I apologize for that, since I really had only a few minor comments on what you actually did. Rather, the area of documentation you've changed has some, uh, weaknesses, suggesting this area deserved more attention in the past than it's gotten. If you're willing, you can help improve it.

I'm marking this "request changes"; but if you don't want to get into additional cleanup, I'd say only the comment on your "these detailed results" and the unnecessary (and possibly misleading) addition of " (optional)" are really complaints about what you wrote. The rest are mostly unrelated issues that happen to become obvious when reading these areas of documentation. (Consider this an "opportunity", or "just your bad luck" 😆)

tangming1996:
@dbutenhof  Thanks a lot for your thorough review. You’re right that the current changes could mislead users. I’ve pushed a new commit to address your feedback:
- Clarified that specifying any --output replaces the default JSON and CSV outputs.
- Updated the HTML example to explicitly preserve JSON and CSV.
- Documented GUIDELLM__DEFAULT_RESULTS_DIR and clarified when path is optional.
- Split JSON, CSV, YAML, HTML, and PLOT into separate output sections.
- Added PLOT formats/DPI details and clarified the special behavior of console output.
- Synchronized the README, analysis guide, benchmark guide, and output reference to avoid contradictory guidance.

Please take another look when convenient. Thank you very much!

docs/getting-started/analyze.md
```diff
@@ -62,24 +62,37 @@ This is the most critical section for performance analysis. It displays detailed
...
+```bash
+guidellm run \
+  --backend kind=openai_http,target=http://localhost:8000 \
+  --data kind=synthetic_text,prompt_tokens=256,output_tokens=128 \
+  --output kind=json \
+  --output kind=csv \
+  --output kind=html
+```
+
+Each file output has a default filename. Add `path=` only when you want to change its name or destination.
```
dbutenhof, line 85: Minor nit, but I'd drop the "only" ... you only *need* to specify `path` to change the default, but there's no reason you can't anyway. I think just "add `path=` when you want to" is just fine.

dbutenhof (dismissed):
That feels oddly repetitive, and maybe it's too bad we "introduce" output controls in separate contexts. But that's not specifically a problem with this PR, and (despite one minor suggestion) I'm going to approve because this looks like a useful improvement.

jaredoconnell (commented):
I'm not opposed to merging once Dave's review comments are addressed, but this is fairly redundant. My recommendation would be to link to other sections more so that you don't need to repeat similar sections multiple times.

tangming1996:
Thanks @dbutenhof and @jaredoconnell for the suggestions. I’ve pushed the updated version in bbb5bbc.
The detailed output configuration now lives in the output guide. The README and getting-started pages retain brief explanations and link to the relevant sections, replacing repeated examples and format descriptions. I also removed the restrictive “only when” wording around path.
tox -e lint-check passes, and all newly added local links and heading anchors have been verified. Please take another look when convenient. Thanks!

jaredoconnell (approved):
It seems reasonable and less redundant.

dbutenhof (approved):
I think I could nit-pick on documentation forever; I decided this was "good enough" before Jared asked for the consolidation, which looks good, so I'll just say "thanks".

dbutenhof:
@mergifyio queue

PR #1079, opened 2026-09-02
Bound token events by when they occur

### LINKED ISSUE #1078: Token thoughput and latency metrics do not proper occluded events in warmup
### Bug Description

When a `warmup` is configured in the profile all requests that overlap the main phase are used for metric calculations even if that request only slightly overlaps. For example:

Image

In this figure requests 1 and 2 are excluded but requests 3-7 are included because they are still active after the warmup ends. Most other tools will exclude 1-7 because they fall partially outside the main phase. This works for latencies at the cost of some valid samples but does not work for throughput metrics since the entire population is required to calculate throughput.

### Expected Behavior

The solution is to look at individual token events rather then full requests when determining if the event falls in-bounds. We should credit events based on their end time, so in the above example 1-6 would be omitted from first token / prefill metrics, but 7 would be counted because its first token event falls in-bounds.

### Steps to Reproduce

Any run with warmup has this issue, but the above example is specifically for a concurrent run.

### Operating System

N/A

### Python Version

N/A

### GuideLLM Version

main

### Installation Method

pip install guidellm

### Installation Details

_No response_

```shell

```

### Additional Context

_No response_

sjmonson:
## Summary

Modify the request bounding code to use different event bounds for each latency event. See #1078 for more details.

## Test Plan

Run a benchmark with warmup and verify differences in collected ttft, token throughput, etc. Note that when running a shorter concurrent run with rampup+warmup vs a longer run with no rampup+warmup the results of the shorter run should be closer than without this patch.

```sh
guidellm run \
    --backend kind=openai_http,target=http://127.0.0.1:8000 
    --profile kind=concurrent,rampup_duration=45,warmup=75 \
    --override profile.streams 200 --data "kind=synthetic_text,prompt_tokens=500,output_tokens=512" \
    --constraint kind=max_duration,seconds=275
```

## Related Issues

- Resolves #1078 

---

## Use of AI

---

# git log

commit 202b75d8b2837de269801aee34c8c4577257e6a3
Date:   Tue Sep 1 21:06:45 2026 +0000

    Clip metrics by start/stop
    
    Only count thoughput that falls within start/stop and only count TTFTs
    that fall fully inside start/stop.
    
    Assisted-by: Codex

commit efb83142c3dbc2b1fcc0b28c7788393e9f358aa9
Date:   Mon Sep 14 18:01:01 2026 -0400

    Bound token latency events by when they occur
    
    TTFT, TTFOT, and ITL are all events the occur during the part of the
    request. When calculating which events are inside the main phase (happen
    after warmup and before cooldown) we want to bound by the event itself
    rather than the request. For first token also be a bit more strict and
    ensure that the whole event is within latency bounds.
    
commit b70695cebb03dafcb725545ea3a6a4e7e5830666
Date:   Tue Sep 15 12:00:18 2026 -0400

    Switch TTFT and TTFOT to open range
    
    When getting prefill events within the main range switch to including
    requests that partially overlap the main phase. This matches how other
    metrics are handled. It is still debatable which approch is better,
    since we are on a bit of a time crunch to get this out stick with this
    approch for now and do more testing in post.
    
commit 308d051cdc81ab2b545c53a04b834cd8f1c29a54
Date:   Tue Sep 15 17:48:30 2026 +0000

    Fix and add tests
    
    Generated-by: Cursor

commit 91637e5655c83f6d5d10e593988c925e34eef141
Date:   Tue Sep 15 18:25:59 2026 +0000

    Add a little documentation for warmup/cooldown
    
    Assisted-by: Cursor

commit 72fc6d5728d2c8bef60d492bf607b92beec69a16
Date:   Tue Sep 15 16:21:12 2026 -0400

    Address review
    
---------

Assisted-by: Codex
Assisted-by: Cursor
Generated-by: Cursor


src/guidellm/benchmark/schemas/accumulator.py
```diff
@@ -629,30 +630,54 @@ def get_sampled(self) -> list[GenerativeRequestStats]:
...
+
+            # Treat instantaneous events as starting exactly when they end
+            actual_start = start_event if start_event is not None else end_event
+
+            if closed:
+                # Fully enclosed: Event starts after range starts AND ends before ends
+                is_in_range = actual_start >= start_time and end_event <= end_time
+            else:
+                # Partial overlap: Event ends after range starts AND starts before ends
+                is_in_range = end_event >= start_time and actual_start <= end_time
```
dbutenhof, line 675: Definition technicality: You say "must start after" and "must end before", but use `>=` & `<=`. Either the description or implementation are wrong. Intuitively, I think the code is right and the description is wrong; although I could imagine an argument for a strict `< end`.

src/guidellm/benchmark/schemas/metrics.py
```diff
@@ -1104,11 +1104,29 @@ def compile(cls, accumulator: GenerativeBenchmarkAccumulator) -> GenerativeMetri
                 incomplete=incomplete,
                 errored=errored,
             ),
+            # TODO: Need to evaluate closed=False vs closed=True for first-token
+            # latencies. Whether excluding more latency data is better for ensuring
```
dbutenhof: The second sentence seems to be a conditional clause without a point. "We need to decide whether"?

src/guidellm/benchmark/schemas/metrics.py
```diff
@@ -1142,27 +1175,45 @@ def compile(cls, accumulator: GenerativeBenchmarkAccumulator) -> GenerativeMetri
...
+                incomplete=accumulator.incomplete.get_within_range(
+                    start_time,
+                    end_time,
+                    start_func=lambda req: req.first_token_iteration,
+                ),
+                errored=accumulator.errored.get_within_range(
+                    start_time,
+                    end_time,
+                    start_func=lambda req: req.first_token_iteration,
+                ),
```
dbutenhof, line 1185: This made me jump back to the method definition. Above, you used `closed=False`; here  you omit it. OK, because it's the default: but if we're going to omit the default in some calls, we should be consistent. That makes it easier to read.

src/guidellm/benchmark/schemas/accumulator.py
```diff
@@ -629,30 +630,54 @@ def get_sampled(self) -> list[GenerativeRequestStats]:
...
+            # Skip if we don't know when the event ended
+            if end_event is None:
+                continue
+
+            # Treat instantaneous events as starting exactly when they end
+            actual_start = start_event if start_event is not None else end_event
+
+            if closed:
+                # Fully enclosed: Event starts after range starts AND ends before ends
+                is_in_range = actual_start >= start_time and end_event <= end_time
```
dbutenhof, line 672: It's funny that you never actually call this with `closed=True`, although the description of token latency metrics is ... somewhat mixed, ending with the explicit comment asking which we want while we go ahead and use an open range calculation.

dbutenhof (commented):
Hmm. Code is straightforward; comments and documentation seem a bit inconsistent about token event metrics, but while some suggest these should be open, all actual calls are closed. Since I'm not sure which is right either (logic kinda breaks down here, and I understand why someone might run a warmup phase to completion and then start metrics clean to avoid the whole issue) ... I'm not objecting. I also want to think a bit more before approving.

src/guidellm/benchmark/schemas/metrics.py
sjmonson: This was originally longer but I truncated it to avoid a giant comment block. I am just going to remove the second sentence since I can't think of a succinct way of describing the problem.

src/guidellm/benchmark/schemas/metrics.py
sjmonson: Yeah... I did that for the ones that were up for debate. Removal is probably better so will-do.

src/guidellm/benchmark/schemas/accumulator.py
sjmonson: I am going to remove the description from the comment because the python conditionals are basically self-documenting.

src/guidellm/benchmark/schemas/accumulator.py
sjmonson: Yeah I debated removing it entirely after changing TTFT/TTFOT to `closed=False` but thought we should at least keep it just in case we change them back.

dbutenhof (approved):

jaredoconnell (approved):

PR #1080, opened 2026-09-02
Fix/stalled scheduled requests

jaredoconnell:
## Summary

Makes it so the scheduler can signal to the scheduler to stop all waiting requests to fix the case where all requests being scheduled in the far-future result in the benchmark hanging until the request happens. 

## Test Plan

Attached is a basic trace file that has a few immediate requests, followed by some over a minute later.
Run a benchmark with that file with a 10 second timeout to observe main waiting, and this branch stopping after the 10 seconds.

```
uv run guidellm run \
  --backend kind=openai_http,target=http://localhost:8000 \
  --profile kind=replay,time_scale=1.0 \
  --data kind=trace_synthetic,path=replay_stall_delayed_requests.jsonl \
  --constraint kind=max_duration,seconds=10
```

## Related Issues

- Improves usability of real-world datasets for testing #1071 

---

## Use of AI

---

# git log

commit 3fa2c6d1f4279be65182d51a9b49b3828d40d568
Date:   Wed Sep 2 14:16:11 2026 -0400

    Poll for cancellation after end condition is reached
    
    This fixes the scheduler sleeping until the next scheduled request, which can be multiple minutes in the future.
    
commit 1311d35b5268cb6a3cba05a7540aa6f513c75017
Date:   Wed Sep 2 16:45:23 2026 -0400

    Address review comment
    
commit f6d4c5d8a53a8783a9e3d82219fc60a247771a25
Author: Jared O'Connell <46976761+jaredoconnell@users.noreply.github.com>
Date:   Thu Sep 3 15:24:19 2026 -0400

    Apply suggestions from code review
    
    Signed-off-by: Jared O'Connell <46976761+jaredoconnell@users.noreply.github.com>

commit f2c558cd9d20f86e2f915f42cf71f14091464a6c
Date:   Thu Sep 3 15:57:36 2026 -0400

    Address review feedback
    
---------

Signed-off-by: Jared O'Connell <46976761+jaredoconnell@users.noreply.github.com>

### DIFF
--- a/docs/guides/datasets.md
+++ b/docs/guides/datasets.md
@@ -218,7 +218,7 @@ GuideLLM supports various file formats for datasets, including text, CSV, JSON, 
     --data kind=trace_synthetic,path=replay.jsonl,timestamp_column=timestamp,prompt_tokens_column=input_length,output_tokens_column=output_length
   ```
 
-  For replay, `time_scale` on the profile is a time scale for the intervals between trace events rather than requests per second. Use `--data-loader kind=pytorch,samples=1000` to limit how many trace rows are loaded and replayed. Use `--constraint kind=max_requests,count=<n>` only as a runtime completion constraint; it does not limit the trace rows loaded from the file. `--constraint kind=max_duration,seconds=<n>` also cancels in-flight waits, including replay sleeps.
+  For replay, `time_scale` on the profile is a time scale for the intervals between trace events. Use `--data-loader kind=pytorch,samples=1000` to limit how many trace rows are loaded and replayed. Use `--constraint kind=max_requests,count=<n>` only as a runtime completion constraint; it does not limit the trace rows loaded from the file. `--constraint kind=max_duration,seconds=<n>` also cancels in-flight waits, including replay sleeps.
 
 - **JSON files (`.json`)**: Where the entire dataset is represented as a JSON array of objects nested under a specific key. To surface the correct key to use, a `--data-column-mapper` argument must be passed in of `"field": "NAME"` for where the array exists. The objects should include `prompt` or other common names for the prompt which will be used as the prompt column. Additional fields can be included based on the previously mentioned aliases for the `--data-column-mapper` argument.
 
--- a/docs/guides/datasets.md
+++ b/docs/guides/datasets.md
@@ -218,7 +218,7 @@ GuideLLM supports various file formats for datasets, including text, CSV, JSON, 
     --data kind=trace_synthetic,path=replay.jsonl,timestamp_column=timestamp,prompt_tokens_column=input_length,output_tokens_column=output_length
   ```
 
-  For replay, `time_scale` on the profile is a time scale for the intervals between trace events rather than requests per second. Use `--data-loader kind=pytorch,samples=1000` to limit how many trace rows are loaded and replayed. Use `--constraint kind=max_requests,count=<n>` only as a runtime completion constraint; it does not limit the trace rows loaded from the file.
+  For replay, `time_scale` on the profile is a time scale for the intervals between trace events rather than requests per second. Use `--data-loader kind=pytorch,samples=1000` to limit how many trace rows are loaded and replayed. Use `--constraint kind=max_requests,count=<n>` only as a runtime completion constraint; it does not limit the trace rows loaded from the file. `--constraint kind=max_duration,seconds=<n>` also cancels in-flight waits, including replay sleeps.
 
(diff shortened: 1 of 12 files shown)


src/guidellm/scheduler/worker_group.py
```diff
@@ -751,11 +800,17 @@ def _update_state_request_counts(self, info: RequestInfo):
...
-            info.timings.request_start or float("inf"),
-            self._state.start_requests_time or float("inf"),
-        )
+        if info.timings.request_start is not None:
+            self._state.start_requests_time = min(
+                info.timings.request_start,
+                self._state.start_requests_time
+                if self._state.start_requests_time is not None
+                else info.timings.request_start,
+            )
```
sjmonson, line 811: It should be enough to just check `request_start` since if set it will always be less then infinity. That way this logic is a little easier to follow.

```suggestion
        if info.timings.request_start is not None:
            self._state.start_requests_time = min(
                info.timings.request_start,
                self._state.start_requests_time or float("inf"),
            )
```

src/guidellm/scheduler/worker_group.py
```diff
@@ -408,14 +408,29 @@ async def request_updates(
...
+                # workers are sleeping until a future target start, because no
+                # request updates arrive during that wait.
+                if self.state is not None:
+                    state_update = self.state.update_state()
+                    if (
+                        state_update.stop_processing
+                        and self.constraint_reached_event is not None
+                        and not self.constraint_reached_event.is_set()
+                    ):
+                        self.constraint_reached_event.set()
```
sjmonson, line 420: Checking `is_set` isn't needed here because `set` is idempotent; its also a race condition technically since this is a mp event (though it doesn't matter in this case). `self.constraint_reached_event is not None` is unnecessary too since its only None before we create the processes, but at least that makes the type checker happy.

```suggestion
                    if (
                        state_update.stop_processing
                        and self.constraint_reached_event is not None
                    ):
                        self.constraint_reached_event.set()
```

sjmonson (changes requested):
Couple nitpicks. One other thing, I think we might need some before/after comparisons for oversaturation detection since I am concerned this might make it trigger too early.

docs/guides/datasets.md
```diff
@@ -218,7 +218,7 @@ GuideLLM supports various file formats for datasets, including text, CSV, JSON,
     --data kind=trace_synthetic,path=replay.jsonl,timestamp_column=timestamp,prompt_tokens_column=input_length,output_tokens_column=output_length
   ```
 
-  For replay, `time_scale` on the profile is a time scale for the intervals between trace events rather than requests per second. Use `--data-loader kind=pytorch,samples=1000` to limit how many trace rows are loaded and replayed. Use `--constraint kind=max_requests,count=<n>` only as a runtime completion constraint; it does not limit the trace rows loaded from the file.
+  For replay, `time_scale` on the profile is a time scale for the intervals between trace events rather than requests per second. Use `--data-loader kind=pytorch,samples=1000` to limit how many trace rows are loaded and replayed. Use `--constraint kind=max_requests,count=<n>` only as a runtime completion constraint; it does not limit the trace rows loaded from the file. `--constraint kind=max_duration,seconds=<n>` also cancels in-flight waits, including replay sleeps.
```
dbutenhof: When I first read this, I thought it sounded odd. I had to see it again before the reason struck me. Basically the "rather than" phrase is actually a reference to the old "rate" aliasing -- that is, "in this case, rate means something other than requests per second". Since the aliasing is gone, I think this phrase becomes confusing. I think it'd be better to remove the "rather than requests per second" phrase.

src/guidellm/scheduler/worker_group.py
```diff
@@ -465,14 +468,50 @@ async def shutdown(self) -> list[Exception]:  # noqa: C901
         self.error_event = None
         if self.mp_manager is not None:
             try:
-                self.mp_manager.shutdown()
+                # Bound this: a live worker holding a Manager connection
```
dbutenhof: I assume "Bound this" refers to the timeout, but I don't think that's really obvious to a casual reader. "Enforce a time limit because..." would be a lot more clear.

dbutenhof (commented):
Looks good; aside from Sam's comments, just a few minor nits...

docs/guides/datasets.md
jaredoconnell: Changed.

src/guidellm/scheduler/worker_group.py
jaredoconnell: Changed.

dbutenhof (approved):

sjmonson (approved):

PR #1081, opened 2026-09-02
Added options to better manage trace request timing

jaredoconnell:
## Summary

Adds options to manage the timing of trace conversations.

## Details

- Added max gap between requests in conversations
- Added max gap between conversations
- Added ability to arrange conversations so that they run in parallel
- **I see some trade-offs with this approach** that are worth discussion. I have the old changes saved, if desired. The main trade-off is you can't just run a sweep of trace replays, each going faster, or each with smaller gaps, or more conversations in parallel, because all of that is pre-computed.

## Test Plan

Here is a dataset you can test it with that works without the subagent changes merged: https://huggingface.co/datasets/semianalysisai/cc-traces-weka-no-subagents-051226

```
guidellm run     --backend "kind=openai_http,target=http://localhost:8000,request_format=/v1/chat/completions,validate_backend=false,model=Qwen/Qwen3-0.6B"     --tokenizer "kind=huggingface_auto,model=Qwen/Qwen3-0.6B"     --profile kind=replay  --constraint "kind=max_requests,count=150" --data "kind=weka,path=weka_traces.jsonl,time_scale=0.1,max_wait=4,max_session_wait=10" --constraint kind=max_duration,seconds=240
```

## Related Issues

- Relates to #1071 

---

## Use of AI

---

# git log

commit 6630dd6c603e3cc5e3038e029e0ce4e1ee7e853a
Date:   Wed Sep 2 14:30:29 2026 -0400

    Added options to better manage trace request timing
    
    - Added max gap between requests in conversations
    - Added max gap between conversations
    - Added ability to arrange conversations so that they run in parallel
    - Added data side option for time_scale
    
commit 86464a70577285fdb8b306d30bcdf8262a42076e
Date:   Thu Sep 3 16:18:42 2026 -0400

    Revert removal of scale from scheduler
    
    Assisted-by: Cursor AI Grok 4.6

commit f978c506e50c454cc525c72ca87c8ec27671df86
Date:   Thu Sep 3 16:38:31 2026 -0400

    Address review comments
    
---------

Assisted-by: Cursor AI Grok 4.6


docs/getting-started/benchmark.md
```diff
@@ -253,11 +251,11 @@ Run with the `replay` profile:
...
-  --profile kind=replay,time_scale=1.0
+  --data kind=trace_synthetic,path=path/to/trace.jsonl,time_scale=1.0 \
+  --profile kind=replay
 ```
 
-The replay profile parameter `time_scale` acts as a scaling factor for the intervals between trace events: `1.0` preserves the original timing, `2.0` doubles the intervals and runs twice as long, and `0.5` halves the intervals and runs twice as fast.
+The data parameter `time_scale` acts as a scaling factor for the intervals between trace events after wait and pack caps: `1.0` preserves the original timing, `2.0` doubles the intervals and runs twice as long, and `0.5` halves the intervals and runs twice as fast. Wait caps (`max_wait`, `max_session_wait`, `min_concurrent_sessions`) are applied in original trace seconds before `time_scale`.
 
 GuideLLM orders trace rows by timestamp before scheduling and payload generation, so each scheduled event uses the token lengths from the same sorted row. Use `--data-loader kind=pytorch,samples=1000` to limit how many trace rows are loaded and replayed. `--constraint kind=max_requests,count=1000` remains a runtime completion constraint; it does not truncate the trace dataset.
```
dbutenhof: Hmm. Minor point, and not really a binding comment on this PR, but I wanted to capture it while I was thinking about it. We've talked about trying to avoid the pre-processing scan through the trace records for startup performance. In that vein, this "before scheduling and payload generation" struck me as a bit misleading; and while that's completely irrelevant here, the stray thought made me realize that it's already unnecessary complication. I think this might read better as simply "GuideLLM schedules trace rows in timestamp order" without the internal implementation suggestions. And that wording would remain literally true if/when we correct the implementation.

sjmonson (dismissed):
This turned out better then I though it would. Super easy to follow all of the timestamp logic. I didn't dig too deep into `trace_session_timing.py` algorithms but from a simple skim they look accurate. 

One minor nit is all of the trace helpers should probably go in a sub-directory but that issue predates this PR and can be solved later.

docs/getting-started/benchmark.md
```diff
@@ -253,11 +251,11 @@ Run with the `replay` profile:
 ```bash
 guidellm run \
   --backend kind=openai_http,target=http://localhost:8000 \
-  --data kind=trace_synthetic,path=path/to/trace.jsonl \
-  --profile kind=replay,time_scale=1.0
+  --data kind=trace_synthetic,path=path/to/trace.jsonl,time_scale=1.0 \
```
dbutenhof, line 258: While this is literally the same as before only migrated to the data specification, it seems silly to show `time_scale=1.0`, which is the default and kinda pointless. Should we be more dramatic with something like `time_scale=0.9`?? 😆

docs/getting-started/benchmark.md
```diff
@@ -210,12 +210,10 @@ guidellm run --profile kind=sweep,sweep_size=10,rampup_duration=10,strategy_type
...
 
 ```bash
-guidellm run --profile kind=replay,time_scale=1.0
+guidellm run --profile kind=replay
 ```
 
-| Profile parameter | Description                                   | Example                                |
-| ----------------- | --------------------------------------------- | -------------------------------------- |
-| `time_scale`      | Time scale for intervals between trace events | `--profile kind=replay,time_scale=2.0` |
+Timeline options such as `time_scale` and wait caps belong on `--data`, not the profile.
```
dbutenhof: This sounds like an issue description, not documentation. You're moving the time scale, so it's no longer documented under profile. (On the other hand, this should be added to the migration guide for 0.8!)

I'm also not entirely convinced it's true, since the "profile" defines the replay logic while the "data" supplies the steps ... logically, the timing controls seem more attached to the profile, though ultimately maybe it doesn't make that much difference.

dbutenhof (commented):
I think I mostly like this: architecturally, I see the concept of "timeline scaling" as more of a "scheduling thing" than a "dataset thing" and I might be slightly happier with the parameters on the profile... ultimately it's not a big deal.

On a whim, I did two AI reviews using my new vscode setup, with Continue/GPT-4.1 and Codex/5.6 Sol (those are the defaults coming up, and I just went with them).

1. I'm really unhappy with the Continue experience, and it may just mean I need to configure it with a better model. It completely failed to understand "PR 1081" or even "the change in the current branch", and to get anywhere sensible I needed to point it at the particular commit number. However, once it got there, it was utterly thrilled with the changes and only warned (as I already have) about clearly calling out the breaking CLI change in documentation. (FYI, if you'd added the other new parameters to the profile there would *be* no breaking change, and I would have preferred that... but again it's probably not a deal breaker for anyone.)
2. Codex/5.6 Sol had no trouble the the PR number, but really got bogged down on the fact that you'd changed the constraint calling signature to allow the `| None`; it *really* did not like that at all. I pushed it further and it reluctantly admitted that you adapted the constraint code to avoid breaking anything, but it still didn't much like it. It didn't even notice the breaking CLI change, apparently... so maybe GPT-4.1 was better (despite the difficulty with the "PR concept).

dbutenhof (changes requested):
I resolved the two conversational comments -- getting a bit more serious since Sam approved.

I do however want to be sure we have clear *migration* documentation about the breaking CLI change here, so please add that.

docs/getting-started/benchmark.md
jaredoconnell: Despite you resolving it, I addressed it.

docs/getting-started/benchmark.md
jaredoconnell: I set it to 0.5 for you 😄

docs/getting-started/benchmark.md
jaredoconnell: This part got reverted, and I wrote a new part without AI.

jaredoconnell:
I reverted it so that you can still adjust the scale on the scheduler side. I wanted to retain the ability to run multiple replays at different time scales.
The new options are retained on the data side.

sjmonson (dismissed):

dbutenhof (dismissed):
I'm actually not sure how I feel about having *both* profile and data scaling controls, though I suppose it's not that hard to understand. It's certainly one way to avoid breaking the CLI, at least.

dbutenhof (approved):

sjmonson (approved):

PR #1084, opened 2026-09-04
fix: don't start in-process backends in the main process

### LINKED ISSUE #1083: vLLM Python backends start up twice
### Bug Description

When using the `vllm_python` backend the initial validation triggers a vLLM startup on the main process. Then the spawned worker initializes the backend again.

### Expected Behavior

This likely broke due to the switch to spawn as the default context, since before we would have passed the initialized backend directly to the worker. The correct behavior is to not initialize the backend in the main process.

### Steps to Reproduce

Run a benchmark with the vllm_python backend an observe vLLM starting twice:

```sh
guidellm run \
    --backend kind=vllm_python,model=Qwen/Qwen3-0.6B \
    --profile kind=constant,rate=3 \
    --constraint kind=max_duration,seconds=20
```

### Operating System

Ubuntu 22.04.5 (Container)

### Python Version

3.12.13

### GuideLLM Version

main

### Installation Method

pip install guidellm

### Installation Details

_No response_

```shell

```

### Additional Context

_No response_

Pragadeesh122:
## Summary

When benchmarking with the `vllm_python` backend, vLLM starts up twice: `resolve_backend` starts and validates the backend in the main process to resolve the default model, and then the spawned worker starts it again. For an in-process backend this loads the entire engine twice.

As noted in the issue, this regressed with the switch to `spawn` as the default multiprocessing context — under `fork` the worker inherited the already-initialized backend, so the main-process startup was not duplicated work.

## Details

- Added a `Backend.requires_startup_for_resolution` property (default `True`). It indicates whether `resolve_backend` must start the backend up in the main process to validate it and resolve its default model.
- `VLLMPythonAsyncBackend` overrides it to `False` (inherited by `VLLMPythonBatchBackend`). Its `default_model` returns the configured model without an engine, so `resolve_backend` resolves the model directly and never loads the engine in the main process.
- `resolve_backend` branches on the flag: `True` keeps the existing startup → validate → resolve → shutdown behavior; `False` resolves the model without starting the backend. The worker (`scheduler/worker.py`) remains responsible for `process_startup`/`validate`, so validation is not skipped — only relocated to the worker for in-process backends.

Scope: only `vllm_python` / `vllm_python_batch` change behavior. All other backends default to `True` and are unchanged. `resolve_backend` already returned a shut-down (not-started) backend, so the returned object's state is identical either way; the only difference is the removed transient main-process engine load.

**One intended behavioral change:** for `vllm_python`, a misconfiguration/health-check error now surfaces at worker startup rather than in the main process. Validation still runs (in the worker); it is not skipped.

## Test Plan

- Added `resolve_backend` tests: it skips `process_startup`/`validate`/`process_shutdown` for a backend reporting `requires_startup_for_resolution=False` (resolving the model directly), and still performs the full startup/validate/shutdown for a backend reporting `True`. The existing validation-error test still passes.
- Added a test asserting `VLLMPythonAsyncBackend.requires_startup_for_resolution is False`.
- `tox -e lint-check`, `tox -e type-check`, and `tox -e test-unit` all pass (2979 passed).

Note: I was unable to reproduce end-to-end against a real vLLM engine locally (no GPU/vLLM install). The change is confined to `resolve_backend`; the vLLM startup path itself is untouched, and the behavior change (which lifecycle calls happen in the main process) is what the added tests verify.

## Related Issues

- Fixes #1083

---

# git log

commit 3215725517da345694b023f2dc96dde12ac1b3c3
Date:   Thu Sep 3 22:19:07 2026 -0500

    fix: don't start in-process backends in the main process (#1083)
    
    resolve_backend started up and validated the backend in the main process to
    resolve the default model, then the spawned worker started it again. For
    in-process backends like vllm_python this loads the entire engine twice.
    
    Add a Backend.requires_startup_for_resolution flag (default True). vllm_python
    overrides it to False so resolve_backend resolves the model from config without
    starting the engine; the worker remains responsible for startup and validation.
    
commit de2cc109f229640f1ffad369d8ef9052d7d4ff78
Date:   Sun Sep 6 14:36:58 2026 -0500

    refactor: rename flag to backend_defines_model per review (#1083)
    
    Address review feedback: rename requires_startup_for_resolution to
    backend_defines_model, make it a ClassVar constant instead of a property, and
    reframe the docstrings/comment around GuideLLM knowing the model from config
    (validation always runs in the worker). No behavior change.
    
---------

### DIFF
--- a/src/guidellm/backends/backend.py
+++ b/src/guidellm/backends/backend.py
@@ -110,6 +110,21 @@ class Backend(
         """
         return None
 
+    @property
+    def requires_startup_for_resolution(self) -> bool:
+        """
+        Whether ``resolve_backend`` must start the backend up in the main process
+        to validate it and resolve its default model.
+
+        In-process backends that initialize heavyweight resources on
+        ``process_startup`` (e.g. an inference engine) return ``False`` so the
+        resources are only initialized once, in the worker process; their
+        ``default_model`` must be resolvable without a startup.
+
+        :return: True to start up during resolution (default), False to skip it.
+        """
+        return True
+
     @abstractmethod
     async def default_model(self) -> str:
         """
--- a/src/guidellm/backends/backend.py
+++ b/src/guidellm/backends/backend.py
@@ -9,7 +9,7 @@ provide a standard interface for distributed execution across worker processes.
 from __future__ import annotations
 
 from abc import abstractmethod
-from typing import Any
+from typing import Any, ClassVar
 
 from guidellm.scheduler import BackendInterface
 from guidellm.schemas import GenerationRequest, GenerationResponse
@@ -110,20 +110,13 @@ class Backend(
         """
         return None
 
-    @property
-    def requires_startup_for_resolution(self) -> bool:
-        """
-        Whether ``resolve_backend`` must start the backend up in the main process
-        to validate it and resolve its default model.
-
-        In-process backends that initialize heavyweight resources on
-        ``process_startup`` (e.g. an inference engine) return ``False`` so the
-        resources are only initialized once, in the worker process; their
-        ``default_model`` must be resolvable without a startup.
-
-        :return: True to start up during resolution (default), False to skip it.
-        """
-        return True
+    #: Whether GuideLLM already knows this backend's default model from its
+    #: configuration, without querying a running engine. When ``True``,
+    #: ``resolve_backend`` reads the model directly and does not start the backend
+    #: in the main process. When ``False`` (the default), the backend must be
+    #: started to discover its model. This is independent of validation: workers
+    #: always start and validate the engine before making requests.
+    backend_defines_model: ClassVar[bool] = False
 
     @abstractmethod
     async def default_model(self) -> str:
--- a/src/guidellm/backends/vllm_python/vllm.py
+++ b/src/guidellm/backends/vllm_python/vllm.py
@@ -109,6 +109,15 @@ class VLLMPythonAsyncBackend(Backend):
         """
         return 1
 
+    @property
+    def requires_startup_for_resolution(self) -> bool:
+        """
+        VLLM Python loads an entire engine on ``process_startup``, so it must only
+        be initialized in the worker process. ``resolve_backend`` resolves the
+        model from configuration without starting the engine in the main process.
+        """
+        return False
+
     async def process_startup(self):
         """
         Initialize VLLM AsyncLLMEngine instance with configured parameters.
--- a/src/guidellm/backends/vllm_python/vllm.py
+++ b/src/guidellm/backends/vllm_python/vllm.py
@@ -14,7 +14,7 @@ import time
 import uuid
 from collections.abc import AsyncIterator
 from pathlib import Path
-from typing import Any, cast
+from typing import Any, ClassVar, cast
 
 import jinja2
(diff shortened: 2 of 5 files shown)


src/guidellm/backends/vllm_python/vllm.py
```diff
@@ -109,6 +109,15 @@ def processes_limit(self) -> int | None:
         """
         return 1
 
+    @property
+    def requires_startup_for_resolution(self) -> bool:
```
dbutenhof, line 113: I don't think the "resolution" in the name is an overly general term for a very specific constraint. The actual criterion implemented is whether GuideLLM *defines* the default model rather than requiring access to the backend engine to make a query. This method wouldn't make sense in any other context. Maybe even something as specific as "backend defines model".

Furthermore, there's nothing about this method that could ever require an instance method -- it can be a static method or even a class constant.

src/guidellm/backends/backend.py
```diff
@@ -110,6 +110,21 @@ def requests_limit(self) -> int | None:
         """
         return None
 
+    @property
+    def requires_startup_for_resolution(self) -> bool:
+        """
+        Whether ``resolve_backend`` must start the backend up in the main process
```
dbutenhof, line 116: This comment is subtly misleading, as you're suggesting a `False` response would allow `validate` -- it doesn't, and you've simply skipped the validation on the in-process backends.

The real criterion is that, as the in-process backends are currently defined (we're implicitly hoping/assuming here that will always hold true), GuideLLM already *knows* the model it will select when initializing the backend engine and doesn't have to ask anybody anything.

src/guidellm/benchmark/entrypoints.py
```diff
@@ -98,18 +98,30 @@ async def resolve_backend(
...
+                    title="Resolving default model from backend.default_model",
+                    status_level="info",
+                )
+            model = await backend_instance.default_model()
+        finally:
+            await backend_instance.process_shutdown()
+    else:
+        # In-process backends (e.g. vllm_python) initialize heavyweight resources
+        # on startup and must only do so in the worker process. Resolve the model
+        # from configuration without starting the backend here; validation runs in
```
dbutenhof, line 117: Maybe even "workers always validate the engine before making requests" rather than implying that's a change for the in-process backends due to this PR.

dbutenhof (changes requested):
Thanks ... I think that'll work, and my comments are relatively minor, but I think would improve clarity.

jaredoconnell (commented):
This is the type of design I expected when I talked to Sam about the issue he created. I agree with Dave's comments.

Pragadeesh122:
Thanks @dbutenhof @jaredoconnell — addressed all three: renamed the flag to `backend_defines_model` and made it a `ClassVar` constant instead of a property, and reframed the base docstring + the resolve_backend comment around GuideLLM knowing the model from config (with validation always happening in the worker, not something this PR changes). No behavior change.

dbutenhof (approved):
It generally seems more clean/clear for class variables to be defined before any methods in the class.

But apparently our lint/type checks don't care, so maybe that's just me. 😆

jaredoconnell (approved):
Designing it this way basically sets the precedent that the primary reason for doing the pre-validation is to query the model from the server. Based on the discussion we've had, this is okay.

Issue #197, opened 2025-06-24
Add new benchmark mode to search for peak goodput under an SLO


dagrayvid:
Often when we benchmark a new model or hardware, the goal is to determine the max RPS or tokens per second that the server can sustain under a certain SLO. We should add a new feature similar to the "sweep" but instead of doing linearly spaced constant RPS runs, it should do something like a binary search to try to find the peak load which the server can handle while meeting a defined latency SLO. 

We would need to support some config options for the SLO, to support p99 or p95 ITL and TTFT. 

I have a rough PoC of this in progress on this branch: https://github.com/dagrayvid/guidellm/tree/goodput, but wanted to open this issue to discuss the idea further and track progress.

QHarshil:
I have this implemented and tested, opened as two PRs: #1085 adds the goodput metric, #1086 adds the search on top of it.

The starting point was the SLO guide in this repo, which lists targets like "TTFT ≤ 200ms for 99% of requests" but has no tooling to measure against them. Objectives are per-request thresholds named after `benchmark_serving` so they carry over. `tpot_ms` maps to inter-token latency rather than time per output token, and it is close to but not identical with vLLM's `tpot`, which measures to request completion.

The search varies concurrency rather than request rate, as your PoC branch also did. Every concurrency level settles into a steady state, whereas a rate above capacity grows a backlog and the measurements then describe the backlog.

Pass/fail uses attainment rather than the goodput rate. Attainment is a ratio, so it does not vary with measurement window length. The throughput-phase rate in #93 does: it is low by ramp/window, which I measured separately at 26% for a 30s run and under 1% at 300s.

I validated by predicting the knee analytically from the mock server's service-time distribution and then running the search:

| objective | predicted | found |
| --- | --- | --- |
| `e2el_ms=1200` | 19 | 19 |
| `e2el_ms=1500` | 26 | 26 |
| `e2el_ms=2000` | 38 | 38 |

Past the knee, throughput stays flat while goodput falls. At `e2el_ms=1500`, concurrency 16 gives 24.6 req/s throughput and 24.6 goodput; concurrency 32 gives 24.3 and 19.3. At `e2el_ms=2000`, concurrency 64 posts the highest throughput of any level tested, with goodput of 1.4.

Each probe carries a Wilson score interval on its attainment. When the interval straddles the target, the probe was too short to decide and the search says so rather than returning a number that looks precise.

[...]

PR #1085, opened 2026-09-04
Add goodput measurement against per-request latency objectives

QHarshil:
## Summary

The SLO guide in this repo documents targets such as "TTFT ≤ 200ms for 99% of requests" across eight use cases, but GuideLLM has no way to measure a run against them. This adds per-request latency objectives and two metrics derived from them.

## Details

- [x] Add `GoodputSLO` with `ttft_ms`, `tpot_ms` and `e2el_ms` thresholds, named after vLLM's `benchmark_serving` so objectives carry between the two tools. `tpot_ms` is compared against inter-token latency, not time per output token. That is the closest GuideLLM metric to vLLM's `tpot` but not identical: vLLM measures to request completion, inter-token latency to the last token received.
- [x] Compile `slo_attainment` (share of requests meeting every objective) and `request_goodput` (rate of those requests). Both are `None` when no objectives are configured.
- [x] Errored requests count as non-conforming, since a request that failed did not deliver a response within its objective. Requests cancelled when a run hits its duration limit are excluded, since the run ended them rather than the server. Requests whose objectives cannot be evaluated, such as time to first token on a non-streaming backend, are excluded from both numerator and denominator so the metric reads as inapplicable rather than as zero.
- [x] Thread objectives from `--metrics` through `Benchmarker.run` onto `BenchmarkConfig`.
- [x] Add console and CSV columns, emitted only when objectives are configured.
- [x] Document both metrics and how to declare objectives.

## Test Plan

- `tox -e lint-check`, `tox -e type-check` and `tox -e test-unit` pass.
- Attainment verified against independent recounts on live runs against the bundled mock server: 0.88 (22/25) for an `e2el_ms` objective, and 0.7107 (506/712) for combined `ttft_ms` and `tpot_ms`, matching to four decimal places.
- Mutation testing over the objective mapping, the attainment denominator and the configuration wiring; each mutation is killed by a test.
- A report written before these fields existed still validates and renders without the new columns.

## Related Issues

Groundwork for #197.

---

## Use of AI

---

# git log

commit 1de51b70f2cb5e505cd70f19e63e2b8ab38a5e79
Date:   Fri Sep 4 00:51:55 2026 -0700

    Add goodput measurement against per-request latency objectives
    
    The SLO guide documents targets such as "TTFT under 200ms for 99% of
    requests" across eight use cases, but GuideLLM had no way to measure a
    run against them.
    
    Add GoodputSLO, a set of per-request latency thresholds named after
    vLLM's benchmark_serving (ttft_ms, tpot_ms, e2el_ms) so objectives carry
    between the two tools. Compile two metrics from it: SLO attainment, the
    share of requests meeting every objective, and request goodput, the rate
    of those requests. Both are None when no objectives are configured.
    
    Errored requests count as non-conforming, since a request that failed did
    not deliver a response within its objective. Requests cancelled when a run
    hits its duration limit are excluded, since the run ended them rather than
    the server. Requests whose objectives cannot be evaluated, such as time to
    first token on a non-streaming backend, are excluded from both numerator
    and denominator so the metric reads as inapplicable rather than as zero.
    
    Objectives are set through --metrics and surface in the console and CSV
    outputs only when configured.
    
---------

### DIFF
--- a/src/guidellm/benchmark/benchmarker.py
+++ b/src/guidellm/benchmark/benchmarker.py
@@ -33,7 +33,7 @@ from guidellm.scheduler import (
     Scheduler,
     SchedulingStrategy,
 )
-from guidellm.schemas.benchmark import TransientPhaseConfig
+from guidellm.schemas.benchmark import GoodputSLO, TransientPhaseConfig
 from guidellm.utils.mixins import InfoMixin
 from guidellm.utils.singleton import ThreadSafeSingletonMixin
 
@@ -69,6 +69,7 @@ class Benchmarker(
         progress: (
             BenchmarkerProgress[BenchmarkAccumulatorT, BenchmarkT] | None
         ) = None,
+        slo: GoodputSLO | None = None,
     ) -> AsyncIterator[BenchmarkT]:
         """
         Execute benchmark runs across scheduling strategies in the profile.
@@ -87,6 +88,8 @@ class Benchmarker(
         :param prefer_response_metrics: Whether to prefer response metrics over
             request metrics, defaults to True
         :param progress: Optional tracker for benchmark lifecycle events
+        :param slo: Per-request latency objectives defining which requests count
+            toward goodput, or None to disable goodput measurement
         :yield: Compiled benchmark result for each strategy execution
         :raises Exception: If benchmark execution or compilation fails
         """
@@ -122,6 +125,7 @@ class Benchmarker(
                     warmup=warmup,
                     cooldown=cooldown,
                     prefer_response_metrics=prefer_response_metrics,
+                    slo=slo,
                     profile=InfoMixin.extract_from_obj(profile),
                     requests=InfoMixin.extract_from_obj(requests),
                     backend=InfoMixin.extract_from_obj(backend),
--- a/src/guidellm/benchmark/outputs/console.py
+++ b/src/guidellm/benchmark/outputs/console.py
@@ -485,6 +485,13 @@ class GenerativeBenchmarkerConsole(GenerativeBenchmarkerOutput):
         :param report: The benchmark report containing throughput metrics
         """
         columns = ConsoleTableColumnsCollection()
+        # Widen the table whenever objectives were configured, not only when
+        # they could be evaluated. A workload that cannot measure an objective,
+        # such as time to first token without streaming, then shows empty cells
+        # instead of the table silently looking as though none were set.
+        report_has_goodput = any(
+            benchmark.config.slo is not None for benchmark in report.benchmarks
+        )
 
         for benchmark in report.benchmarks:
             columns.add_value(
@@ -528,6 +535,22 @@ class GenerativeBenchmarkerConsole(GenerativeBenchmarkerOutput):
                 name="Per Sec",
                 types=("mean",),
             )
+            if report_has_goodput:
+                attainment = benchmark.metrics.slo_attainment
+                columns.add_value(
+                    None if attainment is None else attainment * 100.0,
+                    group="Goodput",
+                    name="Attainment",
+                    units="%",
+                    precision=1,
+                )
+                columns.add_stats(
+                    benchmark.metrics.request_goodput,
+                    status="total",
+                    group="Goodput",
+                    name="Per Sec",
+                    types=("mean",),
+                )
 
         headers, values = columns.get_table_data()
         self.console.print("\n")
--- a/src/guidellm/benchmark/outputs/csv.py
+++ b/src/guidellm/benchmark/outputs/csv.py
@@ -499,6 +499,22 @@ class GenerativeBenchmarkerCSV(GenerativeBenchmarkerOutput):
             "Server Throughput",
             "Concurrency",
         )
+        # Emitted whenever objectives were configured, so a workload that
+        # cannot evaluate them still reports attainment as empty rather than
+        # dropping the columns and reading as though none were set. Every
+        # benchmark in a run shares one objective set, so the columns stay
+        # aligned across rows.
+        if benchmark.config.slo is not None:
+            # Written directly rather than through _add_stats_for_metric,
+            # which drops any status whose total is 0.0. A run that conforms to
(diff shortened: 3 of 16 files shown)


src/guidellm/benchmark/benchmarker.py
```diff
@@ -66,6 +66,7 @@ async def run(
         cooldown: TransientPhaseConfig,
         sample_size: int | None = None,
         prefer_response_metrics: bool = True,
+        slo: GoodputSLO | None = None,
```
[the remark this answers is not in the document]
QHarshil: Confirmed and fixed: `slo` now comes after `progress`, preserving the existing positional order.

src/guidellm/benchmark/outputs/console.py
```diff
@@ -485,6 +485,12 @@ def print_server_throughput_table(self, report: GenerativeBenchmarksReport):
         :param report: The benchmark report containing throughput metrics
         """
         columns = ConsoleTableColumnsCollection()
+        # Only widen the table when latency objectives were actually configured;
+        # every benchmark then contributes a value so the columns stay aligned.
+        report_has_goodput = any(
+            benchmark.metrics.slo_attainment is not None
+            for benchmark in report.benchmarks
```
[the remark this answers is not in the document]
QHarshil: Fixed in both console and CSV: they now gate on `config.slo`, so an objective the workload cannot evaluate shows empty cells instead of rendering as though none were set. Every benchmark in a run shares one objective set, so this also keeps the CSV columns aligned across rows. Covered by `test_shows_columns_when_objectives_cannot_be_evaluated`.

tests/unit/benchmark/test_console_output.py
```diff
@@ -87,3 +96,103 @@ def test_retains_existing_latency_columns(self):
...
+    )
+    output.print_server_throughput_table(SimpleNamespace(benchmarks=benchmarks))
+
+    return (
+        " ".join(str(header) for header in captured["headers"]),  # type: ignore[union-attr]
+        captured["values"],  # type: ignore[return-value]
+    )
+
+
+class TestServerThroughputTableGoodput:
```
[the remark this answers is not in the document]
QHarshil: `tests/unit/benchmark/test_console_output.py` is an existing file from #1053 covering this same module, so I added to it rather than relocating tests unrelated to this change. Happy to move the whole file separately if the convention should be applied retroactively.

QHarshil:
Rebased onto main and CI is green. Mind taking a look when you get a chance?

sjmonson (dismissed):
Yeah this seems reasonable

QHarshil:
Rebased to resolve a conflict with #1084 in `test_entrypoints.py`, which dismissed the approval. The only change is keeping both sets of tests in that file and merging the import; no source changes. Mind re-approving?

src/guidellm/benchmark/schemas/metrics.py
```diff
@@ -879,6 +880,111 @@ class GenerativeMetrics(StandardBaseDict):
         description="Tool call metrics for tokens and call counts"
     )
 
+    # Goodput stats, populated only when latency objectives are configured
+    slo_attainment: float | None = Field(
+        default=None,
+        description=(
+            "Fraction of successful requests meeting every configured latency "
```
dbutenhof: Isn't this really "fraction of requests", since `_compile_goodput` includes errored requests in the total count?

src/guidellm/benchmark/outputs/csv.py
```diff
@@ -499,6 +499,23 @@ def _add_server_throughput_metrics(
...
             "Concurrency",
         )
+        # Emitted whenever objectives were configured, so a workload that
+        # cannot evaluate them still reports attainment as empty rather than
+        # dropping the columns and reading as though none were set. Every
+        # benchmark in a run shares one objective set, so the columns stay
+        # aligned across rows.
+        if benchmark.config.slo is not None:
+            if benchmark.metrics.request_goodput is not None:
+                self._add_stats_for_metric(
```
dbutenhof: This is a potential issue -- `_add_stats_for_metric` skips output on `None` or `== 0.0`, which means that when you fail all SLO criteria, the console and JSON will show 0.0 goodput values but the CSV will omit the goodput columns completely. (Verified with a test run.)

dbutenhof (changes requested):
I didn't see any obvious implementation problems, but I did a Codex review. Codex pointed out two issues, one a minor documentation inconsistency, and the other a problem in the CSV output (which I've experimentally verified) -- both of which I've logged in the respective code.

I actually went back and forth a few times when I first saw this on whether I liked putting the SLO objectives under `--metrics`; but it's "metrics-related" and arguably better than inventing new CLI syntax or a completely new registry object.

I really dislike the term "goodput", but it's a fairly well established term and inventing something else would be silly. I'm also not really thrilled with "slo" as a keyword. While the SLO acronym is well known, as a three-letter (lowercase) keyword I think it lacks clarity ... however short of spelling the whole thing out, which would be awkward, we probably won't do better: so I'm letting both go (almost 😆) without comment...

src/guidellm/benchmark/schemas/metrics.py
QHarshil: Good catch. That wording went stale when errored requests started counting against attainment. Fixed on both descriptions.

src/guidellm/benchmark/outputs/csv.py
QHarshil: Reproduced, and thanks for verifying it. `_add_stats_for_metric` skips any status with a 0.0 total, so a run that conforms to nothing lost the columns while the console and JSON still showed 0.0. Writing the mean directly now, with tests that fail against the old code.

dbutenhof (approved):
Operationally, I would have preferred a second commit for the review changes, which makes it a little easier to review.

The new comments in the CSV code are also perhaps a bit more wordy than necessary, but I'm not inclined to object.

sjmonson (approved):

PR #1088, opened 2026-09-04
Fix inconsistent dataset split resolving

sjmonson:
## Summary

Thought we fixed this at some point but apparently not.

## Details

If a dataset has multiple splits (train, test, etc) the dataset object is of type DatasetDict or IterableDatasetDict. Each deserializer handled split datasets differently, file based handlers would automatically detect a split dataset and atempt to determine the default split. Other deserializers would pass the dataset as-is and fail later when iteration was attempted.

To solve this issue, the DatasetDeserializerFactory now uses the resolve_dataset_split function on all deserialized datasets. A user can still manually select a split using load args.

## Test Plan

Run with any dataset which is split. Pretty much any huggingface loaded dataset will do.

---

## Use of AI

---

# git log

commit c5c14d78900f689b3a844dc95c23fb34071e6c0b
Date:   Fri Sep 4 14:24:04 2026 -0400

    Fix inconsistent dataset split resolving
    
    If a dataset has multiple splits (train, test, etc) the dataset object
    is of type DatasetDict or IterableDatasetDict. Each deserializer handled
    split datasets differently, file based handlers would automatically
    detect a split dataset and atempt to determine the default split. Other
    deserializers would pass the dataset as-is and fail later when iteration
    was attempted.
    
    To solve this issue, the DatasetDeserializerFactory now uses the
    resolve_dataset_split function on all deserialized datasets. A user can
    still manually select a split using load args.
    
---------

### DIFF
--- a/src/guidellm/data/deserializers/deserializer.py
+++ b/src/guidellm/data/deserializers/deserializer.py
@@ -5,7 +5,8 @@ from typing import Protocol, runtime_checkable
 
 from transformers import PreTrainedTokenizerBase
 
-from guidellm.data.schemas import DataNotSupportedError, DatasetType
+from guidellm.data.schemas import DataNotSupportedError, DatasetDictType, DatasetType
+from guidellm.data.utils import resolve_dataset_split
 from guidellm.schemas.data import DataArgs
 from guidellm.utils.registry import RegistryMixin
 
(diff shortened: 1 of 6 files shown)


jaredoconnell (approved):
Looks good. Tested with a basic dataset.

dbutenhof (commented):
> Thought we fixed this at some point but apparently not.

I was absolutely certain I saw an approach for that -- and it was Jared's #906.

This is probably better ... that one was just taking the dataset's first split.

PR #1092, opened 2026-09-04
Trace dataset deserialization cleanup

sjmonson:
## TODO

- [x] Map trace `load_kwargs` to `source.load_kwargs`
- [ ] ~Verify mooncake and trace_synthetic still work~
  ~Edit 1: They don't even work on main right now.~
  Edit 2: The issue has been identified and fix will be separate, for now just test with shorter mooncake datasets

## Summary

Various cleanups of the trace deserialization pathway to improve usability and performance. See individual commit comments bellow for more information.

## Test Plan

Test various sources for data such as WEKA JSONL:

```sh
guidellm run ... \
    --data "kind=weka,source.kind=json_file,source.path=weka.jsonl"
```

WEKA Huggingface:

```sh
guidellm run ... \
    --data "kind=weka,source.kind=hf,source.src=semianalysisai/cc-traces-weka-062126-256k"
```

Mooncake traces (Note, there is currently a bug on main which caused long Mooncake datasets to stall so this example limits the dataset load to the first 100 rows):

```sh
guidellm run ... \
    --data '{ \
                  "kind": "mooncake", \
                  "source": { "kind": "hf", "src": "valeriol29/mooncake-traces" }, \
                  "load_kwargs": { "name": "conversation", "split": "train[:100]" } \
                }'
```

Also test running non-trace or malformed datasets as the source. So far any I tried fail pretty cleanly.

---

- [x] "I certify that all code in this PR is my own, except as noted below."

## Use of AI

- [x] Includes code generated or substantially modified by an AI agent
- [x] Includes tests generated or substantially modified by an AI agent

> NOTE: the `Generated-by` or `Assisted-by` trailers should be used in git commit messages when code or tests were generated or substantially modified by an AI agent, as described in the project's [`DEVELOPING.md`](https://github.com/vllm-project/guidellm/blob/main/DEVELOPING.md) file.

---

# git log

commit 70a96cdaf7ae7eb4c573605aa7a8046260e91c08
Author: Samuel Monson <smonson@redhat.com>
Date:   Fri Sep 4 15:35:24 2026 -0400

    Reuse other deserializers for trace datasets
    
    The trace deserializers are unique in that they are more about the
    dataset format than the file format. It was probabaly a mistake to
    implement them as a deserializer rather than a column-mapper but we have
    to live with that decision for now.
    
    To allow for better reuse of the other deserializers' functionality this
    commit replaces the `path` argument in TraceDataArgs with a `source`
    argument which takes another deserializer's *Args. This allows
[...]

PR #1093, opened 2026-09-05
Address review comments from WEKA Subagent PR

jaredoconnell:
## Summary

Addresses review comments from Dave in #1071 
Also avoids overlapping with #1092

## Test Plan

Review the code and run a WEKA trace.

## Related Issues

#1071 

---

## Use of AI

---

# git log

commit ee1adbaac984f06b284aa0528ad218f5ab3edddc
Date:   Fri Sep 4 20:08:52 2026 -0400

    Address review comments from prior PR
    
commit fcb517a909459bc122fb3a175a042c97b5f81336
Date:   Wed Sep 9 13:26:40 2026 -0400

    Address review comments
    
commit 2a5c4da025daabcfc0d9b294f57671fb5bdfda7e
Date:   Wed Sep 9 13:42:14 2026 -0400

    Address failing tests
    
commit 1c0e09be0c5a13837f60645153a3e858cb2bb8b7
Date:   Wed Sep 9 15:31:27 2026 -0400

    Address review comments
    
---------

### DIFF
--- a/src/guidellm/data/deserializers/trace_weka.py
+++ b/src/guidellm/data/deserializers/trace_weka.py
@@ -91,7 +91,10 @@ def _first_api_request(requests: list[Any]) -> dict[str, Any] | None:
         if not isinstance(row, dict):
             continue
         if _is_subagent_entry(row):
-            found = _first_api_request(list(row.get("requests") or []))
+            inner = list(row.get("requests") or [])
+            if not inner:
+                continue
+            found = _first_api_request(inner)
             if found is not None:
                 return found
             continue
@@ -418,6 +421,18 @@ class WEKATraceFormat(TraceFormatBase):
     def _unpack_conversation(
         self, conversation: Dataset
     ) -> tuple[str, list[dict[str, Any]]]:
+        """Resolve a stub Dataset yielded by ``__iter__`` to the nested request list.
+
+        HuggingFace Arrow cannot store mixed API rows and ``type: "subagent"``
+        groups in one table, so ``__iter__`` queues
+        ``(conversation_id, requests)`` in ``self._conversation_queue`` and
+        yields a one-row Dataset whose only column is ``_weka_index``.
+        ``conversation[0]`` is that stub row (not request index 0);
+        ``_weka_index`` is the queue position of the original conversation.
+
+        :param conversation: One-row stub Dataset from ``__iter__``.
+        :return: ``(conversation_id, requests)`` from the queue.
+        """
         index = int(conversation[0]["_weka_index"])
         return self._conversation_queue[index]
 
@@ -447,11 +462,11 @@ class WEKATraceFormat(TraceFormatBase):
         pending_join_ids: list[str] = []
         last_chain_id: str | None = None
         chain_idx = 0
-        chain_events: list[tuple[float, float | None]] = []
         # Previous API row's stop on this chain only. Subagent children are
         # interleaved in the flattened spec list, so classification cannot
         # use that list's adjacency.
         prev_stop: str | None = None
+        prev_event: tuple[float, float | None] | None = None
 
         for item in requests:
             if _is_subagent_entry(item):
@@ -526,7 +541,9 @@ class WEKATraceFormat(TraceFormatBase):
             )
             last_chain_id = node_id
             chain_idx += 1
-            chain_events.append((abs_t, _optional_float(item.get("api_time"))))
+            event = (abs_t, _optional_float(item.get("api_time")))
+            self._warn_chain_overlap(conversation_id, agent_id, prev_event, event)
+            prev_event = event
             prev_stop = _weka_stop_reason(item)
 
         if pending_join_ids:
@@ -537,14 +554,14 @@ class WEKATraceFormat(TraceFormatBase):
                 agent_id,
             )
 
-        self._warn_chain_overlap(conversation_id, agent_id, chain_events)
         return specs, last_chain_id
 
     def _warn_chain_overlap(
         self,
         conversation_id: str,
         agent_id: str,
-        events: list[tuple[float, float | None]],
+        prev_event: tuple[float, float | None] | None,
+        event: tuple[float, float | None],
     ) -> None:
         """Warn when consecutive turns of one agent overlap in time.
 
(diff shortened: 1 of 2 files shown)


src/guidellm/data/deserializers/trace_weka.py
```diff
@@ -434,6 +437,18 @@ def _tool_response_text(
...
         self, conversation: Dataset
     ) -> tuple[str, list[dict[str, Any]]]:
+        """Resolve a stub Dataset yielded by ``__iter__`` to the nested request list.
+
+        HuggingFace Arrow cannot store mixed API rows and ``type: "subagent"``
+        groups in one table, so ``__iter__`` queues
+        ``(conversation_id, requests)`` in ``self._conversation_queue`` and
+        yields a one-row Dataset whose only column is ``_weka_index``.
+        ``conversation[0]`` is that stub row (not request index 0);
+        ``_weka_index`` is the queue position of the original conversation.
```
sjmonson: I don't really understand this explanation but it might be related to datasets not supporting JSON columns in the version locked here, which is fixed in #1092. Not that having this method hurts, I think it works either way.

src/guidellm/data/deserializers/trace_weka.py
```diff
@@ -557,7 +572,9 @@ def _emit_chain(
             )
             last_chain_id = node_id
             chain_idx += 1
-            chain_events.append((abs_t, _optional_float(item.get("api_time"))))
+            event = (abs_t, _optional_float(item.get("api_time")))
+            self._warn_chain_overlap(conversation_id, agent_id, prev_event, event)
```
sjmonson, line 556: Have you seen this issue on the existing WEKA datasets or is this a proactive check?

sjmonson (dismissed):
Could questions but not blocking.

src/guidellm/data/deserializers/trace_weka.py
```diff
@@ -434,6 +437,18 @@ def _tool_response_text(
     def _unpack_conversation(
         self, conversation: Dataset
     ) -> tuple[str, list[dict[str, Any]]]:
+        """Resolve a stub Dataset yielded by ``__iter__`` to the nested request list.
+
+        HuggingFace Arrow cannot store mixed API rows and ``type: "subagent"``
+        groups in one table, so ``__iter__`` queues
+        ``(conversation_id, requests)`` in ``self._conversation_queue`` and
+        yields a one-row Dataset whose only column is ``_weka_index``.
+        ``conversation[0]`` is that stub row (not request index 0);
```
dbutenhof: Thanks for the docstring.

Do you mean "*IN* that stub row"? What does "request index 0" have to do with this? Do you mean index 0 in a JSON list `requests` column? Row 0? I don't see enough context here to be sure what I'm seeing.

And if `_weka_index` is a "position", e.g., random access, I don't think we're really talking about a "queue" but more like a list, so the name "queue" (which, I know, you didn't add here) seems a bit misleading.

src/guidellm/data/deserializers/trace_weka.py
```diff
@@ -90,7 +90,10 @@ def _first_api_request(requests: list[Any]) -> dict[str, Any] | None:
         if not isinstance(row, dict):
             continue
         if _is_subagent_entry(row):
-            found = _first_api_request(list(row.get("requests") or []))
+            inner = list(row.get("requests") or [])
```
dbutenhof: I think you no longer need the `or []` ... the `get` will return `None` or a value, which might be an empty list, and in either case you'll `continue` below before making the recursive call...

dbutenhof (changes requested):
Overall, the description points to my unresolved comments on #1071; in fact, it addresses several (three, I think) and ignores a few others, plus apparently introducing a few additional changes. None of which is necessarily a problem, but I think maybe the description of scope could be tightened up a bit.

I think the docstring needs to be cleaned up a bit, and there's a minor code tweak ...

src/guidellm/data/deserializers/trace_weka.py
jaredoconnell: I can't just remove it, because if it's None then it could error out. `list(None)` causes an exception.

That said, I could rewrite it like this. What do you think?
```
            inner = row.get("requests")
            if not inner:
                continue
            found = _first_api_request(list(inner))
```

src/guidellm/data/deserializers/trace_weka.py
jaredoconnell: I have seen it on the WEKA datasets I've tested, yes.

src/guidellm/data/deserializers/trace_weka.py
jaredoconnell: Does the latest comment address it?

src/guidellm/data/deserializers/trace_weka.py
jaredoconnell: I redid the comment. How is it now?

src/guidellm/data/deserializers/trace_weka.py
sjmonson: Sorry I am late to ask this, why not just `row.get("requests", [])` or can `row["requests"]` actually be `None`?

src/guidellm/data/deserializers/trace_weka.py
sjmonson: Still a little confusing in terms of the terminology used, but I get the gist now so yeah its acceptable.

src/guidellm/data/deserializers/trace_weka.py
dbutenhof: I guess that's OK, Jared. Passing a `None` to the function is a no-no (because it blindly iterates), but making the call with an empty list we just built just seems silly.

But on second thought, I'm kinda curious why you need the `list` conversion? Essentially, something that's not inherently iterable but becomes iterable via `list`??

src/guidellm/data/deserializers/trace_weka.py
```diff
@@ -418,8 +428,24 @@ def _tool_response_text(
...
+        """Look up the nested request list for a conversation stub from ``__iter__``.
+
+        The ``requests`` column mixes API request records with
+        ``type: "subagent"`` groups, which cannot share one Arrow schema, so
+        ``__iter__`` does not put that list into the yielded Dataset. It
+        appends ``(conversation_id, requests)`` to ``self._conversations``
+        and yields a Dataset with a single row and a single column,
+        ``_weka_index``.
+
+        ``conversation[0]`` is HuggingFace row access for that only Dataset
```
dbutenhof: Better, though "that only Dataset row" sounds pretty odd. Maybe "that [single | one] Dataset row"?

dbutenhof (commented):
I could probably resolve the running conversation, but I think there's still some confusion around the actual predicate here. Can it be `None`? Why, if not, would it need/benefit-from `list`?

Basically this seems down to minor nits...

src/guidellm/data/deserializers/trace_weka.py
jaredoconnell: I reworded it.

jaredoconnell:
> I could probably resolve the running conversation, but I think there's still some confusion around the actual predicate here. Can it be `None`? Why, if not, would it need/benefit-from `list`?
> 
> Basically this seems down to minor nits...

Yes, it can be None if not present in the JSON.

I simplified it a bit. No more list conversion.

dbutenhof (approved):

src/guidellm/data/deserializers/trace_weka.py
jaredoconnell: > Sorry I am late to ask this, why not just `row.get("requests", [])` or can `row["requests"]` actually be `None`?

It can be `None` if the value is missing from the dataset.

Specifically, a `type: "subagent"` group may omit requests or set the JSON as null.

sjmonson (approved):

PR #1094, opened 2026-09-05
fix: load SQLite dataset files through a database connection

git-jxj:
## Summary

Loading a local SQLite dataset with `kind=db_file` fails because its filesystem path is passed to `Dataset.from_sql` as a database URI. The existing success test mocks the database read and hides the failure.

## Details

Open the file using a SQLite connection and close it after loading. Replace the mocked read with a real SQLite query, and document how to supply that query through `load_kwargs.sql`.

## Test Plan

- The real SQLite regression fails before the fix and passes afterward.
- 407 data-module tests passed; 1 optional HDF5 test skipped.
- Independently loaded a database whose filename contains spaces and `#`, and verified SQL filtering returns the expected row.
- `tox -e lint-check,type-check` passed.
- Full unit suite: 3,015 passed, 31 skipped, 163 xfailed, 14 failed. The 14 failures are audio tests unable to load FFmpeg shared libraries through torchcodec. The same 14 test nodes fail with the unmodified upstream source in the same environment.

## Related Issues

None.

---

## Use of AI

> `Generated-by: Codex` and DCO trailers are retained in the commit and the generated squash data below.

---

# git log

commit 6c3c6d5a1c1335b95dfbd346fce31e03adc87453
Author: xinjun.jiang <xinjun.jiang@daocloud.io>
Date:   Sat Sep 5 20:30:02 2026 +0800

    fix: load SQLite dataset files through a database connection
    
    Generated-by: Codex
    Signed-off-by: xinjun.jiang <xinjun.jiang@daocloud.io>

---------

Generated-by: Codex
Signed-off-by: xinjun.jiang <xinjun.jiang@daocloud.io>

### DIFF
--- a/docs/guides/datasets.md
+++ b/docs/guides/datasets.md
@@ -27,6 +27,8 @@ In addition, you can specify additional arguments to the dataset loading with th
 
 - load_kwargs: Additional arguments to the dataset loading. For example, dataset splits can be specified with `--data '{"kind":"huggingface","source":"my/dataset","load_kwargs":{"split":"test"}}'`.
 
+For SQLite `.db` files, provide the query through `load_kwargs.sql`, for example `--data '{"kind":"db_file","path":"prompts.db","load_kwargs":{"sql":"SELECT text FROM samples"}}'`.
+
 ### Data Loader
 
 You can specify the data loader with `--data-loader` and a configuration string. The only current type is `pytorch`:
--- a/src/guidellm/data/deserializers/file.py
+++ b/src/guidellm/data/deserializers/file.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
+import sqlite3
 from collections.abc import Callable
+from contextlib import closing
 from pathlib import Path
 from typing import Any
 
@@ -200,7 +202,8 @@ class DBFileDatasetDeserializer(DatasetDeserializer):
                 f"expected str or Path to a local .db file, got {path}"
             )
 
-        return Dataset.from_sql(con=str(path), **config.load_kwargs)
+        with closing(sqlite3.connect(path)) as connection:
+            return Dataset.from_sql(con=connection, **config.load_kwargs)
 
 
 @DatasetDeserializerFactory.register("tar_file")
--- a/tests/unit/data/deserializers/test_file.py
+++ b/tests/unit/data/deserializers/test_file.py
@@ -345,10 +345,10 @@ def test_hdf5_file_deserializer_success(tmp_path):
 
 
 @pytest.mark.sanity
-def test_db_file_deserializer_success(monkeypatch, tmp_path):
+def test_db_file_deserializer_success(tmp_path):
     """DBFileDatasetDeserializer reads .db file into Dataset.
 
-    ### WRITTEN BY AI ###
+    ## WRITTEN BY AI ##
     """
 
     def create_sqlite_db(path: Path):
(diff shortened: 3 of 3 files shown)


git-jxj:
I rebased this PR onto the current main branch and moved the dataset guide addition to its new `docs/en/guides/datasets.md` path. GitHub now reports the branch mergeable. The targeted deserializer tests passed (13 passed, 1 optional HDF5 test skipped), and `lint-check` and `type-check` passed locally. The new head is 31f94734; its CI is running.

dbutenhof (changes requested):
My big problem with this solution is that `Datasets.from_sql` is not strictly limited to sqlite3, and you've forced that restriction. I did some independent experimentation: *in theory* we want to pass a SQLAlchemy URI for the dataset `path` string, which `Datasets.from_sql` would read and process using your query string.

There are a couple of problems with this:

1. You can't actually pass a SQLAlchemy URI through the `kind=db_file,path=` parameter, because `DataFileArgs.path` is typed as `Path`. The `Path` processing removes the "redundant" `/` characters.
2. The `DBFileDatasetDeserializer` checks that `config.path` is a file that exists: and that won't work with a URI.
3. Passing a `Path` to `Datasets.from_sql` won't work: it's expecting a `str` or a `Connection`.
4. Even once we've gotten past those issues, while `Datasets.from_sql` is prepared to use SQLAlchemy to open and access the database (of various kinds, like PostgreSQL, etc.), it requires that "someone else" have imported SQLAlchemy, and we don't.

SQLAlchemy is a complicated ecosystem, not a single Python module ... for example to actually use PostgreSQL we'd also need psycopg2, pymysql, etc.

Clearly nobody has used `--data kind=db_file`, because there's no way to make it work. I don't think we're interested in tackling the entire SQLAlchemy ecosystem to handle the full potential range of databases through `Datasets.from_sql` at this time, so the core change of your PR is probably fine.

__HOWEVER__:

I want it implemented in a way that can be extended to support the full range later, and that means we can't blindly *assume* a sqlite3 database as you've done.

We want to support the full allowed *syntax*, and that means taking a SQLAlchemy URI to specify the database; something like:

`--data kind=db_file,path=sqlite:///file.db,load_kwargs.sql="SELECT text FROM samples"`

This implies a few additional changes:

1. A new `class SqlDataArgs(DataArgs)` ... since we can be pretty sure nobody's successfully *used* `--data kind=file_db`, it probably makes sense to change `path` to `uri` to clarify the meaning. And it'll be typed as a `str`... or perhaps as Pydantic `AnyUrl` (which provides parsing and minimal syntax checking).
2. In the `DbFileDatasetDeserializer`, we can check that `config.uri.scheme == "sqlite"` and raise a `ValueError` if something else is specified. (Note, that's the main advantage of using `AnyUrl` vs `str`).
3. Obviously, we remove the file exists & is_file checks from the deserializer.
4. If we imported SQLAlchemy (without any additional drivers), we could just pass `str(config.uri)` to `Datasets.from_sql`; but I think that, at least for now, it makes more sense to skip that new dependency and use your simple `sqlite3.connect` code, with `str(config.uri.path)[1:]` as the file name. (Because `.path` includes the leading `/`.)
5. The `db_file` format should be documented as a bullet in the "Accepted types" list, including the information you added in your separate paragraph. It should say that GuideLLM currently only supports loading from a sqlite3 database.

PR #1097, opened 2026-09-07
fix: export tool call metrics to CSV

tangming1996:
CSV output defined tool call metrics but skipped the tool_call modality during export, causing benchmark reports to silently omit token and call count statistics. Include the modality and regression coverage so CSV reports remain consistent with the collected benchmark data.

Assisted-by: Codex

## Details

- [ ]

## Test Plan

-

## Related Issues

- Resolves #

---

## Use of AI

---

# git log

commit d974f2e422703b67e2033c4c9dc80911df352f87
Date:   Mon Sep 7 16:46:31 2026 +0800

    fix: export tool call metrics to CSV
    
    CSV output defined tool call metrics but skipped the tool_call modality during export, causing benchmark reports to silently omit token and call count statistics. Include the modality and regression coverage so CSV reports remain consistent with the collected benchmark data.
    
    Assisted-by: Codex
    
---------


sjmonson (dismissed):

tangming1996:
@Mergifyio queue

tangming1996:
@sjmonson Thanks for reviewing and approving this PR. All checks have passed, but I’m unable to add it to the Mergify queue because the command requires write permission. Could you please help queue and merge PR #1097 when convenient? Thank you!

src/guidellm/benchmark/outputs/csv.py
```diff
@@ -138,7 +138,13 @@ async def finalize(self, report: GenerativeBenchmarksReport) -> Path:
                 self._add_server_throughput_metrics(
                     benchmark, benchmark_headers, benchmark_values
                 )
-                for modality_name in ["text", "image", "video", "audio"]:
+                for modality_name in [
```
dbutenhof: Maybe it would be better to use `for modality_name in MODALITY_METRICS.keys():`? Presumably what happened here is that when `tool_call` was added, the iterator here wasn't fixed. Unless someone can think of a good reason we might *not* want to iterate over all defined modalities here, it seems better to avoid duplicating the keys unnecessarily...

dbutenhof (commented):
This seems like a proper fix, but I question the original logic that enabled the omission when adding the new key ... I think that could have been easily avoided.

Just a comment -- I don't actually object to letting this go with the point fix and perhaps generalizing the code later.

src/guidellm/benchmark/outputs/csv.py
jaredoconnell: So the values are repeated here, in the Console output type, and in the literal.

We could do that by setting them in the literal, and extract them as follows:
```
Status = Literal[
                    "text",
                    "image",
                    "video",
                    "audio",
                    "tool_call",
]
STATUSES = get_args(Status)
```

Or, as you suggested, by using MODALITY_METRICS. That's probably the way to go.

tangming1996: Good point. I updated the CSV exporter to iterate directly over MODALITY_METRICS, making it the single source of truth for supported modalities. I also removed the duplicated Literal declaration and the related type: ignore.
The CSV unit tests and tox -e lint-check both pass.

dbutenhof (approved):
👍🏻 

(And I love getting rid of type-check overrides!)

jaredoconnell (approved):

PR #1098, opened 2026-09-07
Fix shared mutable fill values in argument parsing

tangming1996:
ArgStringParser evaluated the fill-value factory once during construction and reused that object for every sparse list position. With fill_value=list or dict, mutating one placeholder changed other positions and results from subsequent decodes using the same parser.

Retain the factory and invoke it separately for every added list position. Keep a separate reference value for the existing overwrite checks. Add regression coverage for mutable list and dictionary fills, nested paths, repeated decode calls, and successive set calls.

Assisted-by: Codex

## Details

- [ ]

## Test Plan

-

## Related Issues

- Resolves #

---

## Use of AI

---

# git log

commit 2153ab1a17a2a77058db58775000d1759ef59ba1
Date:   Mon Sep 7 17:17:02 2026 +0800

    Fix shared mutable fill values in argument parsing
    
    ArgStringParser evaluated the fill-value factory once during construction
    and reused that object for every sparse list position. With fill_value=list
    or dict, mutating one placeholder changed other positions and results from
    subsequent decodes using the same parser.
    
    Retain the factory and invoke it separately for every added list position.
    Keep a separate reference value for the existing overwrite checks. Add
    regression coverage for mutable list and dictionary fills, nested paths,
    repeated decode calls, and successive set calls.
    
    Assisted-by: Codex

---------

Assisted-by: Codex

### DIFF
--- a/src/guidellm/utils/arg_string.py
+++ b/src/guidellm/utils/arg_string.py
@@ -67,12 +67,16 @@ class ArgStringParser:
 
         :param skip_invalid: If True, skips invalid key=value pairs rather than raising
         :param fill_value: Callable that returns fill value for sparse lists.
-            Defaults to lambda: None.
+            Called separately for each fill position. Defaults to lambda: None.
         :param allow_overwrite: If True, allows overwriting existing values.
             Defaults to False, which raises an error on overwrites.
         """
         self.skip_invalid = skip_invalid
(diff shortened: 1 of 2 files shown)


sjmonson (commented):
I guess this is technically fixing a bug in the utility but it is impossible to hit this problem in the normal use of GuideLLM. Neither approving or disapproving for now.

PR #1102, opened 2026-09-08
fix(metrics): preserve all cancellations in final scheduler counts

DivyamTalwar:
## Summary

Scheduler request counts must include all cancellations, including requests cancelled before dispatch. The `incomplete` field represents cancelled requests because the scheduler reuses `StatusBreakdown`; filtered benchmark totals belong in `metrics.request_totals`.

This revision addresses the review by preserving those semantics and fixing the remaining event-order bug: a queued cancellation returns before updating the cached scheduler metrics, so a final queued cancellation can be missing from the exported scheduler counts. Compile those counts from the final `SchedulerState`, and document the distinction between scheduler outcomes and filtered measurement-window totals.

## Review follow-up

The original change incorrectly derived scheduler counts from the filtered request accumulators. That implementation has been replaced in commit `db6ec4ee7d7b78ed7f1029c5977f2cd17492b842`. Serialized field names remain compatible.

## Validation

- With the original cached-count implementation restored, the new tests produced 5 failures: four cases where queued cancellation was processed last, plus the case where every request was cancelled before dispatch.
- All six terminal-event orderings are tested for both completion and error, alongside started and queued cancellations. The tests check both scheduler counts and filtered benchmark totals.
- `uv run --frozen tox -e tests -- tests/unit/benchmark tests/unit/scheduler/test_worker.py -k 'not run_async_lifecycle and not run_with_timings' --tb=short`: 153 passed, 12 deselected.
- `uv run --frozen tox -e lint-check,type-check`: passed; mypy checked 222 source files.
- Broader benchmark/scheduler suite: 700 passed, 16 xfailed, 5 failed during `multiprocessing.Manager` startup. A representative failure was reproduced with the original metrics implementation: the sandbox rejects socket creation with `PermissionError`, followed by `EOFError`. This is not a clean full-suite result.
- `git diff --check`: passed.

## Use of AI

Implementation, tests, and documentation were assisted by OpenAI Codex and remain subject to maintainer review. The follow-up commit includes `Generated-by: OpenAI Codex` and the author's sign-off.

---

# git log

commit 7c2d3ed629b79ad3cb1e23d6f772ab2b45521525
Author: Divyam Talwar <divyamtalwar0@gmail.com>
Date:   Tue Sep 8 14:35:51 2026 +0530

    fix(metrics): exclude queued cancellations from requests made
    
    Compile request outcome counts from the request accumulators so cancellations that never reached dispatch do not leak into report totals after a later terminal update.
    
    Generated-by: OpenAI Codex
    
    Signed-off-by: Divyam Talwar <divyamtalwar0@gmail.com>

commit db6ec4ee7d7b78ed7f1029c5977f2cd17492b842
Author: Divyam Talwar <divyamtalwar0@gmail.com>
Date:   Wed Sep 9 03:02:43 2026 +0530

    fix(metrics): preserve all cancellations in final scheduler counts
    
    Address PR #1102 review feedback: scheduler requests_made includes every
    cancellation, with the shared incomplete field representing cancelled.
    Filtered measurement-window counts remain in metrics.request_totals.
    
    Compile scheduler outcomes from the final SchedulerState. The cached
    accumulator can be stale when the last event is a queued cancellation,
    because that event returns before updating scheduler metric counts.
    Document the distinction without changing the serialized field names.
    
    Regression evidence: restoring the original cached-count implementation
    produced 5 failures (four trailing-cancellation permutations and the
    all-queued-cancelled case). The corrected path passes all permutations
    of queued cancellation, started cancellation, and completion/error.
    
    Validation:
    - tox focused benchmark and worker checks: 153 passed, 12 deselected.
    - tox lint-check and type-check: passed (222 source files).
    - Broader benchmark/scheduler run: 700 passed, 16 xfailed, 5 failed.
      The five failures occur at multiprocessing.Manager startup in this
      sandbox. A representative case also fails with the original metrics
      implementation: PermissionError creating a socket, followed by EOFError.
    - git diff --check: passed.
    
    Generated-by: OpenAI Codex
    
    Signed-off-by: Divyam Talwar <divyamtalwar0@gmail.com>

---------

Signed-off-by: Divyam Talwar <divyamtalwar0@gmail.com>

### DIFF
--- a/src/guidellm/benchmark/schemas/metrics.py
+++ b/src/guidellm/benchmark/schemas/metrics.py
@@ -139,7 +139,16 @@ class SchedulerMetrics(StandardBaseDict):
             request_end_time=accumulator.timings.finalized_request_end,
             end_time=scheduler_state.end_time or -1.0,
             # Request details tracked by the scheduler
-            requests_made=accumulator.scheduler_metrics.requests_made,
+            requests_made=StatusBreakdown(
+                successful=len(accumulator.completed.requests_stats),
+                incomplete=len(accumulator.incomplete.requests_stats),
+                errored=len(accumulator.errored.requests_stats),
+                total=(
+                    len(accumulator.completed.requests_stats)
+                    + len(accumulator.incomplete.requests_stats)
+                    + len(accumulator.errored.requests_stats)
+                ),
+            ),
             # Scheduler internal performance timings
             queued_time_avg=accumulator.scheduler_metrics.queued_time.mean or -1.0,
             resolve_start_delay_avg=(
--- a/src/guidellm/benchmark/schemas/metrics.py
+++ b/src/guidellm/benchmark/schemas/metrics.py
@@ -84,7 +84,10 @@ class SchedulerMetrics(StandardBaseDict):
 
     # Request details tracked by the scheduler
     requests_made: StatusBreakdown[int, int, int, int] = Field(
-        description="Request counts by status: successful, incomplete, errored, total"
+        description=(
+            "Scheduler request counts: successful, cancelled (stored as incomplete), "
+            "errored, total. Includes requests cancelled before dispatch."
+        )
     )
 
     # Scheduler internal performance timings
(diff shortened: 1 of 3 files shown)


src/guidellm/benchmark/schemas/metrics.py
```diff
@@ -139,7 +139,16 @@ def compile(
             request_end_time=accumulator.timings.finalized_request_end,
             end_time=scheduler_state.end_time or -1.0,
             # Request details tracked by the scheduler
-            requests_made=accumulator.scheduler_metrics.requests_made,
```
sjmonson, line 142: `scheduler_metrics.requests_made` is supposed to include all cancelled requests, the final tallies after filtering are in metrics.request_totals. Technically "incomplete" is the wrong term here and it should be successful, cancelled, errored. Its only incomplete because we are reusing the StatusBreakdown object.

sjmonson (changes requested):

PR #1103, opened 2026-09-08
Improve Determinism for Integration Tests

jaredoconnell:
## Summary

This ensures that cross-worker cancellations work.

In #1080 it was changed to properly stop on cancellation. This goes further to poll because it wasn't propagated to other workers.

This should theoretically make it handle cancellation better, and it should satisfy the failing test.

## Test Plan

This should mainly apply to trace formats with long gaps, just like in #1080 

## Related Issues

- Follow up to #1080 

---

## Use of AI

---

# git log

commit 3c40e11ed9c2e17f65b694b12854aa5e645aa2d2
Date:   Tue Sep 8 11:43:09 2026 -0400

    Ensure workers poll for cancellation
    
commit 7375e275c889025545989bd1308192020c5c2883
Date:   Tue Sep 8 14:04:48 2026 -0400

    Revert polling, and improve cancellation process
    
commit 356d15a753e270915214fc68215462e09515807a
Date:   Tue Sep 8 15:54:20 2026 -0400

    Pre-determine cancellation time to improve determism.
    
---------

### DIFF
--- a/src/guidellm/scheduler/worker.py
+++ b/src/guidellm/scheduler/worker.py
@@ -128,10 +128,6 @@ class WorkerProcess(Generic[RequestT, ResponseT]):
         self.backend_started = False
         self.messaging_started = False
         self.turns_queue: list[DAGExecutionState[RequestT, ResponseT]] = []
-        # In-flight ``_process_next_graph_node`` tasks. These are created with
-        # ``asyncio.create_task`` and are not cancelled merely by cancelling
-        # ``_process_requests_loop``; the stop-event handler must cancel them.
-        self._pending_request_tasks: set[asyncio.Task[None]] = set()
 
     def run(self):
         """
@@ -242,13 +238,7 @@ class WorkerProcess(Generic[RequestT, ResponseT]):
                 self.constraint_reached_event,
                 poll_interval=self.messaging.poll_interval,
             )
-            # Request tasks are create_task children of the loop, so they keep
-            # running (including asyncio.sleep on a replay target) until they
-            # are cancelled. Cancel them here, when the stop event is observed,
-            # rather than waiting for the loop's CancelledError handler.
             processing_task.cancel()
-            for task in list(self._pending_request_tasks):
-                task.cancel()
             # Let in-flight nodes finish reporting their own terminal update
             # before the sweep below, so a node is never reported twice.
             with contextlib.suppress(asyncio.CancelledError):
@@ -300,12 +290,13 @@ class WorkerProcess(Generic[RequestT, ResponseT]):
         Schedules and processes requests according to the timing strategy while
         maintaining the configured concurrency limit through semaphore coordination.
         """
+        pending_tasks: set[asyncio.Task] = set()
         try:
             # Run request processing
             async_semaphore = asyncio.Semaphore(self.async_limit)
 
-            def _task_done(task: asyncio.Task[None]):
-                self._pending_request_tasks.discard(task)
+            def _task_done(task: asyncio.Task):
+                pending_tasks.discard(task)
                 async_semaphore.release()
 
                 if not task.cancelled() and (exception := task.exception()):
@@ -326,13 +317,12 @@ class WorkerProcess(Generic[RequestT, ResponseT]):
                 request_task = asyncio.create_task(
                     self._process_next_graph_node(target_start=request_time)
                 )
-                self._pending_request_tasks.add(request_task)
+                pending_tasks.add(request_task)
                 request_task.add_done_callback(_task_done)
         except asyncio.CancelledError as err:
-            in_flight = list(self._pending_request_tasks)
-            for task in in_flight:
+            for task in pending_tasks:
                 task.cancel()
-            await asyncio.gather(*in_flight, return_exceptions=True)
+            await asyncio.gather(*pending_tasks, return_exceptions=True)
 
             raise err
 
@@ -566,8 +556,8 @@ class WorkerProcess(Generic[RequestT, ResponseT]):
         self, request: RequestT, request_info: RequestInfo, target_start: float
     ):
         request_info.timings.scheduled_at = request_info.timings.dequeued
-        if target_start > time.time():
-            await asyncio.sleep(target_start - time.time())
+        if target_start > (current_time := time.time()):
+            await asyncio.sleep(target_start - current_time)
             # Adapt delay so that scheduled at reflects the sleep time
             request_info.timings.scheduled_at = target_start
 
--- a/src/guidellm/scheduler/worker.py
+++ b/src/guidellm/scheduler/worker.py
@@ -309,10 +309,10 @@ class WorkerProcess(Generic[RequestT, ResponseT]):
                     worker_index=self.worker_index
                 )
 
-                if (
-                    time_until := request_time - time.time()
-                ) >= self.fut_scheduling_time_limit:
-                    await asyncio.sleep(time_until - self.fut_scheduling_time_limit)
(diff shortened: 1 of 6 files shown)


src/guidellm/scheduler/worker.py
```diff
@@ -552,12 +552,36 @@ def _abort_remaining_nodes(
         if state in self.turns_queue:
             self.turns_queue.remove(state)
 
+    async def _sleep_until_target(self, target_time: float) -> None:
+        """Sleep until ``target_time``, aborting if the worker should stop.
+
+        A single ``asyncio.sleep`` for a far-future replay timestamp is only
+        interrupted if the parent task is cancelled. Under multiprocessing that
+        cancel can lag ``max_duration``, so poll stop events on the same
+        interval used elsewhere in the worker.
```
sjmonson: This comment makes no sense.

sjmonson (commented):
If I am reading this correctly this makes every single waiting request poll. I don't love adding `n` more places where we poll in userspace, it puts more stress on the async scheduler. To me, this is masking another issue since these task should already be canceled by their parent when the stop event occurs.

jaredoconnell:
> If I am reading this correctly this makes every single waiting request poll. I don't love adding `n` more places where we poll in userspace, it puts more stress on the async scheduler. To me, this is masking another issue since these task should already be canceled by their parent when the stop event occurs.

I see. Yeah I guess that was a workaround. I updated it to just change how it handles cancellation itself.

src/guidellm/scheduler/worker.py
```diff
@@ -238,7 +242,13 @@ async def _process_requests(self):
                 self.constraint_reached_event,
                 poll_interval=self.messaging.poll_interval,
             )
+            # Request tasks are create_task children of the loop, so they keep
+            # running (including asyncio.sleep on a replay target) until they
+            # are cancelled. Cancel them here, when the stop event is observed,
+            # rather than waiting for the loop's CancelledError handler.
             processing_task.cancel()
+            for task in list(self._pending_request_tasks):
```
sjmonson: Why cast to a list, you can iterate a set?

src/guidellm/scheduler/worker.py
```diff
@@ -317,12 +326,13 @@ def _task_done(task: asyncio.Task):
                 request_task = asyncio.create_task(
                     self._process_next_graph_node(target_start=request_time)
                 )
-                pending_tasks.add(request_task)
+                self._pending_request_tasks.add(request_task)
                 request_task.add_done_callback(_task_done)
         except asyncio.CancelledError as err:
-            for task in pending_tasks:
+            in_flight = list(self._pending_request_tasks)
```
sjmonson: Another unnecessary list cast.

src/guidellm/scheduler/worker.py
```diff
@@ -238,7 +242,13 @@ async def _process_requests(self):
                 self.constraint_reached_event,
                 poll_interval=self.messaging.poll_interval,
             )
+            # Request tasks are create_task children of the loop, so they keep
+            # running (including asyncio.sleep on a replay target) until they
+            # are cancelled. Cancel them here, when the stop event is observed,
+            # rather than waiting for the loop's CancelledError handler.
             processing_task.cancel()
+            for task in list(self._pending_request_tasks):
+                task.cancel()
```
sjmonson: This still doesn't make sense, if you call `processing_task.cancel()` then `_process_requests_loop` should catch the `asyncio.CancelledError` and move to cancel all of the requests in its own loop. The only time that wouldn't happen is if `_process_requests_loop` is awaiting another function that captures `CancelledError` and blocks, which would be the real issue.

sjmonson (changes requested):

src/guidellm/scheduler/worker.py
```diff
@@ -128,6 +128,10 @@ def __init__(
         self.backend_started = False
         self.messaging_started = False
         self.turns_queue: list[DAGExecutionState[RequestT, ResponseT]] = []
+        # In-flight ``_process_next_graph_node`` tasks. These are created with
+        # ``asyncio.create_task`` and are not cancelled merely by cancelling
+        # ``_process_requests_loop``; the stop-event handler must cancel them.
+        self._pending_request_tasks: set[asyncio.Task[None]] = set()
```
dbutenhof: `asyncio.Task[None]` looks odd. Isn't it technically `asyncio.Task[Future[Awaitable[<type>]]]`? I guess nobody's complaining about the `None`, but why did you even need to add it? Was the generic no longer good enough?

dbutenhof (commented):
It feels like this is adding layers & complication to work around a more basic problem in our cancellation handling; but the overall logic is complicated... does this really help?

src/guidellm/scheduler/worker.py
jaredoconnell: Apparently the reason was to prevent modifying the set, but it didn't matter in this case. This has been replaced with a new design.

jaredoconnell:
A new design is proposed now. The other attempts were attempting to make the cancellation pathway more reliable. The new design pre-determines when it needs to stop, which is more reliable. However, this comes at the small expense of some additional data being passed around, making it slightly less generic.

dbutenhof (commented):
Can you describe exactly where this is supposed to help? I'm not seeing any obvious change in mid-benchmark cancellation responsiveness... and in particular if I let a constant@100 benchmark run to about 50% completion and then ^C, everything (still) seems to effectively "hang" ... at least for a significant period of time.

What mode of operation, specifically, has this improved for you?

jaredoconnell:
> Can you describe exactly where this is supposed to help? I'm not seeing any obvious change in mid-benchmark cancellation responsiveness... and in particular if I let a constant@100 benchmark run to about 50% completion and then ^C, everything (still) seems to effectively "hang" ... at least for a significant period of time.
> 
> What mode of operation, specifically, has this improved for you?

The specific situation this affects is a trace dataset with a long duration between requests, with an end constraint during that long gap.

For example, if a trace has requests within the first minute, then some requests 10 minutes later, if you set the max duration constraint to 2 minutes, the benchmark will take 11 minutes because it would have waited until the next request to cancel.

This was mostly fixed in #1080, but the test `test_max_duration_cancels_long_replay_sleep` is unreliable, at least with the durations used in the test (currently max-seconds 0.4s, with requests scheduled at 5 seconds). So this would make that test reliable without polling. The test checks to make sure it properly ended the benchmark and cancelled the requests before it hit 5 seconds, but it's failing because on CI there are no cancelled requests. So there's either a race, or it's just so slow that it isn't cancelling. It appears to be reliable with this PR's changes.

dbutenhof:
Ah. I usually run integration tests, and I haven't seen a failure locally. Have there been a lot of CI failures I failed to notice? Is it failing a lot for *you* locally?

This is an interesting adaptation; but if cancellation worked right, I don't really see how it'd be necessary. Seems there's some key we're still missing here ... but maybe this is a reasonable mitigation. I'll look at the code again.

jaredoconnell:
> Ah. I usually run integration tests, and I haven't seen a failure locally. Have there been a lot of CI failures I failed to notice? Is it failing a lot for _you_ locally?
> 
> This is an interesting adaptation; but if cancellation worked right, I don't really see how it'd be necessary. Seems there's some key we're still missing here ... but maybe this is a reasonable mitigation. I'll look at the code again.

Yes, main has been flaky recently. This is an attempt to make it less flaky.

https://github.com/vllm-project/guidellm/actions?query=branch%3Amain

The test seems valuable, so I'd rather make the process more consistent and easier to test than to remove the test.

PR #1070, opened 2026-08-31
Fix a bunch of issues with logging

sjmonson:
## Summary

Console logging was completely broken when used with the Live progress. This PR fixes a bunch of issues that prevented it from working. Additionally logging due to external packages in workers should also be handled properly (e.g. vLLM)

## Details

There were a bunch of reasons logging was not working correctly:

### 1. `loguru` grabbed an instance of stderr before Rich could redirect it

`loguru`, package we use for logging, is initialized early into the package load and by default will grab a copy of `sys.stderr`. Later when Rich would load, it would replace `sys.stderr` (and `sys.stdout`) with a custom class called `FileProxy` which would redirect writes though itself. Since `loguru` loaded earlier it did not have this redirect in place.

### 2. Spawned worker processes did not receive Rich's `FileProxy`

 TODO Write explaination

### 3. `loguru` not properly configured for multiprocess logging

When initializing non-fork workers `loguru` requires that its state be passed as an argument to the worker in order to share logging queues. This does not matter as much for console logs since the logger can just be initialized again with the stderr pipe in (2). However for log files this causes each worker to open the same file separately. 

## Test Plan

The best way to test this PR is with the following patch which will add a bunch of extra logging from both the main process and workers:

```diff
diff --git a/src/guidellm/scheduler/worker.py b/src/guidellm/scheduler/worker.py
index ece7f587..1819ba81 100644
--- a/src/guidellm/scheduler/worker.py
+++ b/src/guidellm/scheduler/worker.py
@@ -327,6 +327,7 @@ class WorkerProcess(Generic[RequestT, ResponseT]):
 
             # Main loop; loop until canceled
             while True:
+                logger.info("Getting next request to process for worker {}", self.worker_index)
                 await async_semaphore.acquire()
                 request_time = await self.strategy.next_request_time(
                     worker_index=self.worker_index
diff --git a/src/guidellm/scheduler/worker_group.py b/src/guidellm/scheduler/worker_group.py
index 2a3caa2e..21f879f5 100644
--- a/src/guidellm/scheduler/worker_group.py
+++ b/src/guidellm/scheduler/worker_group.py
@@ -746,6 +746,7 @@ class WorkerGroupState(Generic[RequestT, ResponseT]):
     def _update_state_request_counts(self, info: RequestInfo):
         finalized = time.time()
 
+        logger.info("Received request update: {} (status={})", info.request_id, info.status)
         if info.status == "queued":
             self._queued_request_ids.add(info.request_id)
[...]

PR #1115, opened 2026-09-09
Log benchmark progress by default alongside the Rich display

### LINKED ISSUE #704: Display benchmark progress in non-interactive consoles
### Problem Statement

When running GuideLLM in a non-interactive environment such as a K8s Pod there is no indication of benchmark progress until the entire run completes.

### Proposed Solution

Create a system for emitting status updates in non-interactive mode.

### Alternatives Considered

_No response_

### Usage Examples

```markdown

```

### Additional Context

_No response_

provoke210:
## Summary

Log benchmark progress automatically through the existing logger, including in non-interactive shells where the Rich display is unavailable. Users do not need any additional progress configuration. Interactive users keep the Rich display and receive progress logs at the same time.

## Behavior

- Emit INFO records at strategy start, at most once every ten seconds during progress updates, and at completion, including short benchmarks.
- Include elapsed time, successful/errored/incomplete request counts, request throughput and output-token throughput.
- Enable logging in the benchmark entrypoint for both CLI and Python callers. Console-disable flags affect displays, not logging.
- Reuse existing logger levels and sinks, including stderr redirection and the built-in structured file logger. No new logger configuration is introduced.
- Reuse main-process lifecycle callbacks and metrics; periodic updates are not an independent heartbeat during a stalled callback.
- Keep logging and Rich as independent `BenchmarkerProgress` implementations. `Benchmarker.run` accepts a list and concurrently notifies every observer for each lifecycle event, using `_notify_progress` with `asyncio.gather`; each event waits for all callbacks before propagating the first failure. Finalization runs only after normal benchmark execution. The logging observer releases its own state without calling the process-wide `logger.complete()` method.

The final implementation has no new console-mode selector or progress-logging CLI switch.

## Validation

Latest revision on macOS arm64 / Python 3.12.14:

- 179 related tests passed: benchmark unit tests, CLI display flags, queued logs sharing a Rich terminal, and three end-to-end cases using the built-in mock HTTP server.
- Observer tests use synchronization barriers to verify concurrent dispatch and cover observer-update, initialization, and scheduler failures. Initialization and scheduler failures propagate without finalization. Normal execution still finalizes both observers.
- All 12 focused observer and progress tests also passed under Python 3.10.21 without a TaskGroup backport.
- End-to-end cases use no progress-enabling options: default Rich plus logs, displays disabled, and interactive progress disabled. Each crosses the default ten-second interval and checks intermediate records plus final counts against the JSON report. Existing file-logger configuration is also exercised.
- `tox -e lint-check` passed; configured type checks passed across 224 source files.
- No GPU is required for these reporting tests.

Earlier full-suite validation: 3024 unit tests passed, 31 skipped, 163 xfailed; 28 integration tests passed, 25 xfailed; 14 end-to-end tests passed. Fourteen audio unit failures and one realtime audio setup error caused by unavailable local torchcodec/FFmpeg libraries were also reproduced on unchanged upstream commit `d469d0e`. The full suite has not been rerun for this revision.

## Related Issues

Fixes #704.

---

## Use of AI

Implementation and tests were developed with OpenAI Codex assistance. Commits include `Assisted-by: OpenAI Codex` and DCO sign-offs; generated tests carry the repository's required AI markers.

---

# git log

commit 88faa28dd235e35e1a39a6ad2cd8a76f32ff2ba7
Author: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>
Date:   Thu Sep 10 03:31:45 2026 +0800

    feat: add opt-in plain text benchmark progress
    
    Assisted-by: OpenAI Codex
    Signed-off-by: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>

commit 199696fe61c7e5b21f3ebf292f62581a41e7c73e
Author: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>
Date:   Sat Sep 12 00:52:19 2026 +0800

    docs: clarify console progress defaults
    
    Use reference wording and remove the legacy progress alias mention as requested in review.
    
    Assisted-by: OpenAI Codex
    Signed-off-by: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>

commit 16bb46b5f74d917846dded03f6d560cc8364c7eb
Author: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>
Date:   Sat Sep 12 02:35:38 2026 +0800

    feat: log benchmark progress alongside Rich display
    
    Add independent interval-limited INFO progress logging with queued-log cleanup and coverage for simultaneous Rich rendering, file logging, and disabled displays.
    
    Assisted-by: OpenAI Codex
    Signed-off-by: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>

commit f0e1220502591917590e281ccf3bab48c076e092
Author: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>
Date:   Sat Sep 12 02:40:40 2026 +0800

    fix: log benchmark progress by default without extra configuration
    
    Enable interval-limited progress logging in the benchmark entrypoint alongside optional Rich output. Remove the opt-in console registry, simple display and logging CLI switch.
    
    Assisted-by: OpenAI Codex
    Signed-off-by: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>

commit 174f301daa89f6da4706703ef32c1ab46e3ea750
Author: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>
Date:   Sat Sep 12 03:33:17 2026 +0800

    refactor: notify independent benchmark progress trackers concurrently
    
    Make Benchmarker.run accept a list of progress observers. Separate logging from Rich rendering, await all callbacks before propagating failures, and finalize observers on exit.
    
    Assisted-by: OpenAI Codex
    Signed-off-by: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>

commit 8646b8fefcb33c5addb2a02c8ac0577703973357
Author: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>
Date:   Sat Sep 12 04:02:24 2026 +0800

    refactor: use task groups for benchmark progress callbacks
    
    Replace manual callback exception collection with TaskGroup lifecycle scopes. Preserve Python 3.10 support with the taskgroup backport and verify sibling cancellation and finalization.
    
    Assisted-by: OpenAI Codex
    Signed-off-by: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>

commit 7a786593d609c9807f307217cc2197e8e76950e1
Author: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>
Date:   Sat Sep 12 04:07:08 2026 +0800

    refactor: finalize progress only after normal benchmark execution
    
    Remove the outer try/finally as requested in review and update observer tests to expect initialization and scheduler failures to skip finalization.
    
    Assisted-by: OpenAI Codex
    Signed-off-by: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>

commit 0cfa0fdb6ad7ad4706f77585dca2827014573752
Author: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>
Date:   Sat Sep 12 04:09:05 2026 +0800

    fix: leave logger completion to the process lifecycle
    
    Only release observer state during progress finalization, without calling the process-wide logger.complete method.
    
    Assisted-by: OpenAI Codex
    Signed-off-by: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>

commit cd6844ab4dacc49a057e7bf5e20b6a1728b3e085
Author: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>
Date:   Sat Sep 12 04:18:59 2026 +0800

    refactor: restore progress notification helper
    
    Revert only TaskGroup dispatch and its backport dependencies. Keep finalization on normal completion and leave process logger completion unchanged.
    
    Assisted-by: OpenAI Codex
    Signed-off-by: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>

---------

Assisted-by: OpenAI Codex
Signed-off-by: Haozhe Jiang <162801044+provoke210@users.noreply.github.com>


jaredoconnell (commented):
This does work for me. It seems like a reasonable design. I was just imagining adding log messages rather than printing to the console. I'll let my teammates have a say before approving.

provoke210:
> This does work for me. It seems like a reasonable design. I was just imagining adding log messages rather than printing to the console. I'll let my teammates have a say before approving.

Thanks for testing! I followed the earlier discussion in #704, which suggested an explicit --console kind=simple option with readable stdout output. Happy to adjust based on the team's feedback.

docs/getting-started/benchmark.md
```diff
@@ -300,6 +300,23 @@ guidellm run \
...
+  --backend kind=openai_http,target=http://localhost:8000 \
+  --data kind=synthetic_text,prompt_tokens=256,output_tokens=128 \
+  --profile kind=synchronous \
+  --constraint kind=max_duration,seconds=60 \
+  --console kind=simple,interval=10
+```
+
+Each strategy prints a start line, periodic statistics, and a completion line. Lines include elapsed time, successful/errored/incomplete request counts, request throughput, and output token throughput. They are flushed immediately and contain no terminal cursor controls, so they can be read through `kubectl logs` or a pipe. Periodic lines are emitted on benchmark updates, at most once per `interval` seconds (default: 10); they are not an independent heartbeat if execution stalls. Completion is always printed, including for benchmarks shorter than the interval.
+
+The default remains `--console kind=rich`. `--disable-console-interactive` (legacy alias: `--disable-progress`) suppresses Rich progress but allows an explicitly selected simple display. `--disable-console` suppresses both modes as well as other console output. Console progress does not change the saved result files.
```
dbutenhof: "The default remains" sounds like a migration guide, which this isn't. We do maintain a migration guide, and an entry for GuideLLM 0.8 CLI changes would be a good idea -- but *this* section should just say "The default is". Also, let's drop the "legacy alias" bit: the point of labeling it "legacy" is that it's an old option we plan to drop, so let's not add new documentation about it.

dbutenhof (changes requested):
Sorry: but just one little documentation detail ...

dbutenhof (dismissed):

docs/getting-started/benchmark.md
provoke210: > "The default remains" sounds like a migration guide, which this isn't. We do maintain a migration guide, and an entry for GuideLLM 0.8 CLI changes would be a good idea -- but _this_ section should just say "The default is". Also, let's drop the "legacy alias" bit: the point of labeling it "legacy" is that it's an old option we plan to drop, so let's not add new documentation about it.

Updated as suggested. Thanks for the review!

sjmonson (changes requested):
Sorry not a fan of this design. The way I saw progress updates working was through logging. The user should not have to choose between having the nice rich progress and having progress logged.

Edit: A little more clarification, #704 is specifically for cases where the calling shell is non-interactive and the rich console is not available. The user should not have to configure anything extra, progress updates should just happen and since #1070 logging does not interfere with the console. Doing it with logging also is more intuitive, the user could redirect just the logs to another sink with a stderr pipe or they could configure our built-in file logger.

provoke210:
> Sorry not a fan of this design. The way I saw progress updates working was through logging. The user should not have to choose between having the nice rich progress and having progress logged.
> 
> Edit: A little more clarification, #704 is specifically for cases where the calling shell is non-interactive and the rich console is not available. The user should not have to configure anything extra, progress updates should just happen and since #1070 logging does not interfere with the console. Doing it with logging also is more intuitive, the user could redirect just the logs to another sink with a stderr pipe or they could configure our built-in file logger.

Thanks for the clarification! I've updated the PR so progress is logged automatically through the existing logger, with no extra configuration required. Rich remains available alongside logging, and I've removed the console-mode selector and opt-in logging flag. Progress logs use the existing stderr/file handlers. Could you take another look?

src/guidellm/benchmark/progress.py
```diff
@@ -236,6 +243,122 @@ def _sync_run_progress(self):
             )
 
 
+class GenerativeLoggingBenchmarkerProgress(
+    BenchmarkerProgress[GenerativeBenchmarkAccumulator, GenerativeBenchmark]
+):
+    """Log interval-limited progress alongside an optional display."""
```
sjmonson: I don't love how this class is a wrapper around the console progress class. Maybe the best thing would be to change `Benchmarker.run`'s `progress` argument to a list and have it try to update all `BenchmarkerProgress` objects in parallel.

sjmonson (changes requested):

src/guidellm/benchmark/progress.py
provoke210: > I don't love how this class is a wrapper around the console progress class. Maybe the best thing would be to change `Benchmarker.run`'s `progress` argument to a list and have it try to update all `BenchmarkerProgress` objects in parallel.

Thanks for the suggestion. I've updated Benchmarker.run to accept a list of progress trackers and notify them concurrently. Logging and Rich are now independent implementations, with no wrapper relationship.

src/guidellm/benchmark/benchmarker.py
```diff
@@ -87,100 +88,120 @@ async def run(
...
+                strategies_generator = profile.strategies_generator()
+                strategy: SchedulingStrategy | None
+                constraints: dict[str, Constraint] | None
+                strategy, constraints = next(strategies_generator)
+
+                while strategy is not None:
+                    logger.info("Starting benchmark for strategy: {}", strategy)
+                    await _notify_progress(
+                        *(tracker.on_benchmark_start(strategy) for tracker in trackers)
+                    )
```
sjmonson: Instead of adding all of this messy new error handling please just use TaskGroups. E.g.

```python
async with asyncio.TaskGroup() as tg:
    for tracker in trackers:
        tg.create_task(tracker.on_benchmark_start(strategy))
```

sjmonson (changes requested):

src/guidellm/benchmark/benchmarker.py
sjmonson: Oh sorry, I misunderstood what the change was because GitHub's diff viewer is terrible. Either your `_notify_progress` method or a `TaskGroup` is fine, but I prefer TaskGroups here.  The outer try/finally loop is unnecessary. If an exception occurs here the whole benchmark crashes.

src/guidellm/benchmark/progress.py
```diff
@@ -236,6 +243,102 @@ def _sync_run_progress(self):
...
+        """
+        if self._state:
+            self._state.complete(benchmark)
+            self._log_update("completed")
+            self._state = None
+
+    async def on_finalize(self):
+        """Release progress state and drain queued log records."""
+        self._state = None
+        await logger.complete()
```
sjmonson: Sorry just noticed this. Do not call `logger.complete()`. That is the method which is called when **all** logging is complete in the process.

sjmonson (changes requested):

src/guidellm/benchmark/progress.py
provoke210: > Sorry just noticed this. Do not call `logger.complete()`. That is the method which is called when **all** logging is complete in the process.

How about now?

sjmonson (changes requested):
Sorry one last thing. I forgot that TaskGroups aren't officially supported in Python 3.10. Could you just revert the TaskGroup change to remove the new dependency.

provoke210:
> Sorry one last thing. I forgot that TaskGroups aren't officially supported in Python 3.10. Could you just revert the TaskGroup change to remove the new dependency.

Sure, I'll switch back to _notify_progress.I added a backport for Python 3.10 compatibility, but happy to revert to _notify_progress to avoid the extra dependency.

sjmonson (approved):
Perfect, thanks for putting up with all of my demands :). Assuming CI passes this will land after @dbutenhof re-approves.

provoke210:
> Perfect, thanks for putting up with all of my demands :). Assuming CI passes this will land after @dbutenhof re-approves.

Thanks for the guidance and review!

dbutenhof (approved):
> after @dbutenhof re-approves.

Ha ... not entirely sure this is "re"-anything since it's completely changed. But, yeah -- this looks like a valid transparent approach with no CLI change, and I've verified that it works.

jaredoconnell (commented):
Looks good to me.

PR #1153, opened 2026-09-16
OTeL trace support

### LINKED ISSUE #968: Implement OTEL trace format replay
Support OTEL replaying OTEL trace format

References include

* https://github.com/open-telemetry/semantic-conventions-genai
* https://github.com/kubernetes-sigs/inference-perf/blob/main/docs/otel_trace_replay.md

jaredoconnell:
## Summary

 OTEL traces are a useful addition because they typically include the actual text of the requests, rather than needing to mess with hash IDs to synthetically generate data.

This PR adds support for OTeL, with options for both constructed history or raw trace history. Trace-offs are documented.

## Details

- Adds OTeL Support
- Adds history=trace to use history context new rather than history context inherit.
- Adds history=dag for built history, and uses the diff between requests to find just the new messages, rather than including all of them.
- Due to the history=trace option not using DAG history, and using history context = new, I added another option to see the conversation history, called `preceding_nodes` 
- Does not support sub-agents. It appears that subagents are not as common in OTeL traces
- Has more fruitful tool call support than WEKA due to more tool information being included in the trace.
- The trace data options added alongside WEKA work for this, too.

## Test Plan

- The documentation includes some commands that are worth testing for the permutations of the settings.

## Related Issues

Resolves #968 

---

## Use of AI

---

# git log

commit 99b5247aaa37b153ca889e493f188d75517bac98
Date:   Fri Sep 11 13:13:52 2026 -0400

    MVP of OTeL Trace Support
    
    Currently minimal support. No tool call. Incomplete hash support.
    
commit ca08ad452303cbf8f0cbf472bbfeda88ed9042dc
Date:   Tue Sep 15 17:54:58 2026 -0400

    Add preceeding_nodes request metric
    
    Useful for multiturn dags that embed their own history, rather than using DAG history.
    
    Generated-by: Cursor AI Composor 2.5

commit f9cde77b686cb49766b237f90aeaf28ca9d73780
Date:   Tue Sep 15 17:57:58 2026 -0400

    Complete OTel Support
    
    Properly utilizes the content from the OTeL traces.
    Also adds multiple settings to change how GuideLLM uses the OTeL trace.
    
commit 0fb0fbb0d9996bed546de89ae1e6f192427d417f
Date:   Wed Sep 16 12:24:38 2026 -0400

    Remove synthetic option to simplify initial implementation
    
commit a314fe9683a4c99066e1a73bd8591449bb2bdfd1
Date:   Fri Sep 18 23:42:15 2026 -0400

    Address review comments
    
    Remove speculative defensive code
    Document actually needed defensive code
    Remove responses support for future consideration
    And more.
    
    Assisted-by: Cursor AI Grok 4.6

---------

Assisted-by: Cursor AI Grok 4.6
Generated-by: Cursor AI Composor 2.5


src/guidellm/schemas/data/deserializers/trace_otel.py
```diff
@@ -0,0 +1,60 @@
...
+
+from pydantic import Field
+
+from guidellm.schemas.data.deserializers.trace_common import TraceDataArgs
+from guidellm.schemas.data.entrypoints import DataArgs
+
+__all__ = ["OTELTraceFormatArgs"]
+
+
+@DataArgs.register(["otel", "opentelemetry", "otel_trace"])
```
sjmonson: Maybe drop `"otel_trace"` from the names. We don't really want to add multiple options without good reason. E.g. Legacy name or full name outside of GuideLLM.

src/guidellm/scheduler/dag.py
```diff
@@ -80,6 +80,10 @@ def __init__(self, graph: ConversationGraph[_RequestT]):
         self._available_after: dict[str, float] = dict.fromkeys(graph.nodes, 0.0)
         self._in_progress: set[str] = set()
         self._aborted: bool = False
+        order = self.topological_order()
```
sjmonson: If we are pre-computing `topological_order` to use it here, then maybe turn it into a cached_property so when we use it in the worker_group we don't need to re-compute it twice.

src/guidellm/scheduler/dag.py
```diff
@@ -293,6 +297,20 @@ def _depth(nid: str) -> int:
...
+        Independent of ``history_context``; reflects DAG structure only.
+
+        :param node_id: The node to look up.
+        :return: Number of nodes before this one in topological order (0-based).
+        :raises KeyError: If ``node_id`` is not in the graph.
+        """
+        if node_id not in self._preceding_nodes:
+            raise KeyError(f"Unknown node_id '{node_id}'")
+        return self._preceding_nodes[node_id]
+
```
sjmonson: Not really `compute`, maybe `count_preceding_nodes` or maybe it would be better to also make this a cached property called `preceding_nodes` and just `return { node_id: index for index, node_id in enumerate(self.topological_order)}` to avoid calculating it unless we actually use this method.

sjmonson: Ah ok, I realizing after reading a bit more code that this will always be called not just for otel, but I still think a cached property for both this and `topological_order` is a good idea.

sjmonson (commented):
Not finished reviewing, but here are a couple small things to start.

docs/guides/datasets.md
```diff
@@ -20,7 +20,7 @@ The following arguments configure datasets and their processing:
...
-  - `trace_synthetic`, `mooncake`, `weka` — replay trace data with `--profile kind=replay`. Required field: `source` (a nested dataset config such as `json_file` or `huggingface`), see other supported sources for details. Optional: `timestamp_column`, `prompt_tokens_column`, `output_tokens_column`, `time_scale`, and other format-specific options. `weka` uses different column defaults (`t`, `in`, `out`). See [Trace File Formats](./trace_replay.md).
+  - `trace_synthetic`, `mooncake`, `weka`, `otel` — replay trace data with `--profile kind=replay`. Required field: `source` (a nested dataset config such as `json_file` or `huggingface`), see other supported sources for details. Optional: `timestamp_column`, `prompt_tokens_column`, `output_tokens_column`, `time_scale`, and other format-specific options. `weka` uses different column defaults (`t`, `in`, `out`). `otel` (aliases `opentelemetry`, `otel_trace`) reads GenAI span attributes and ISO-8601 timestamps. Replay sends recorded `gen_ai.input.messages` (`history=trace` by default). See [Trace File Formats](./trace_replay.md).
```
dbutenhof: I think this bullet item is getting way too long and detailed with the weka and otel details. Can we strip these down to just the high-level summary and the link? Although a bit *more* explanation of the nested source might be helpful, at least with a simple example of the structure?

src/guidellm/data/deserializers/trace_otel.py
```diff
@@ -0,0 +1,952 @@
...
+_LLM_OPERATIONS = frozenset({"chat", "generate", "text_completion"})
+_NON_LLM_OPERATIONS = frozenset({"invoke_agent", "execute_tool", "embeddings"})
+_FAILED_STATUS_CODES = frozenset({2, "2", "ERROR", "STATUS_CODE_ERROR"})
+_TOOL_FINISH_REASONS = frozenset(
+    {"tool_calls", "tool_call", "tool_use", "function_call"}
+)
+_NANOSECONDS = 1e16
+_MILLISECONDS = 1e11
+
+_OTEL_KINDS = ["otel", "opentelemetry", "otel_trace"]
```
sjmonson: See other comment

```suggestion
_OTEL_KINDS = ["otel", "opentelemetry"]
```

src/guidellm/data/deserializers/trace_otel.py
```diff
@@ -0,0 +1,952 @@
...
+    if isinstance(value, (int, float)):
+        number = float(value)
+        if abs(number) >= _NANOSECONDS:
+            return number / 1e9
+        if abs(number) >= _MILLISECONDS:
+            return number / 1e3
+        return number
+    raise InvalidRowError(
+        f"OTEL format: unsupported timestamp type {type(value).__name__}: {value!r}"
+    )
```
sjmonson, line 311: Surely this is overkill. In the example dataset I only see ISO-8601 formatted times. Even in the case where this is actually needed it should live in `utils`, or better yet use one of the many dependencies we have that can parse timestamps.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -471,6 +471,107 @@ def _constrain_tool_call_body(
...
+            if not isinstance(text, str):
+                text = "" if text is None else str(text)
+            items.append(
+                {"role": role, "content": [{"type": "input_text", "text": text}]}
+            )
+        else:
+            items.append({"role": role, "content": "" if text is None else text})
+    return items
+
+
```
sjmonson: Can we get this in a separate PR? Converting chat to responses is pretty niche and I see a lot of issues with doing it this way. For instance this completely breaks multi-modal.

src/guidellm/data/deserializers/trace_otel.py
```diff
@@ -0,0 +1,952 @@
...
+        attributes = span_attributes(turn["span"])
+        input_messages = parse_gen_ai_messages(attributes.get("gen_ai.input.messages"))
+        if not input_messages:
+            raise InvalidRowError(
+                "OTEL format: span missing gen_ai.input.messages. "
+                "Use kind=trace_synthetic for token-count-only traces."
+            )
+        if self.config.history != "runtime" or turn_idx == 0:
+            return input_messages
+        return self._runtime_raw_delta(input_messages, rows[turn_idx - 1])
```
sjmonson, line 920: When we initial discussed this I assumed that the `raw` column would be the entire request and not just the messages/content section. Oh well, I guess not as reusable as I thought.

docs/guides/trace_replay.md
```diff
@@ -42,6 +43,16 @@ guidellm run \
...
 
+**OTEL dataset from `huggingface`:**
+
+```bash
+guidellm run \
+  --backend kind=openai_http,target=http://localhost:8000 \
+  --profile kind=replay \
+  --data kind=otel,source.kind=huggingface,source.source=ibm-research/synthetic-conversations-traces,load_kwargs.split=train \
+  --data-loader kind=pytorch,samples=2
+```
```
sjmonson, line 58: If we are treating these as full examples they should probably all have constraints set. Also `--data-loader kind=pytorch,samples=2` should be removed. Also technically a lot of these are setting the default value. Maybe change `time_scale=2.0` on the `trace_synthetic` example, remove the split option from WEKA and OTEL.

docs/guides/trace_replay.md
```diff
@@ -121,3 +132,74 @@ Modified defaults:
...
+Start from Hugging Face. `samples` counts conversations (`trace_id`s), not spans. IBM traces have 30–50 LLM calls each, so keep `samples` small at first.
+
+**Default (`history=trace`):** send each span's recorded messages in full. Later turns wait on the DAG but use `history_context=new`, so live completions are not spliced into the next prompt.
+
+```bash
+guidellm run \
+    --backend kind=openai_http,target=http://localhost:8000 \
+    --profile kind=replay \
+    --data kind=otel,source.kind=huggingface,source.source=ibm-research/synthetic-conversations-traces,load_kwargs.split=train \
+    --data-loader kind=pytorch,samples=2
```
sjmonson: Replace with constraint

docs/guides/trace_replay.md
```diff
@@ -121,3 +132,74 @@ Modified defaults:
...
+```
+
+**Recorded messages with DAG history (`history=runtime`):** send only the new messages; prior turns come from live completions (`history_context=full`). Requires each span's input to continue the previous span's input plus output.
+
+```bash
+guidellm run \
+    --backend kind=openai_http,target=http://localhost:8000 \
+    --profile kind=replay \
+    --data kind=otel,source.kind=huggingface,source.source=ibm-research/synthetic-conversations-traces,load_kwargs.split=train,history=runtime \
+    --data-loader kind=pytorch,samples=2
```
sjmonson: Replace with constraint

sjmonson (changes requested):
Well with the most basic test it seems to work. A lot of the conditional paths in `trace_otel.py` seem unlikely to ever occur but such is true for most parsing code, especially when its all written by AI. Few more nits and I would like to see the chat completions -> responses code moved outside this PR to give it a little more time in the oven.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -471,6 +471,107 @@ def _constrain_tool_call_body(
...
+        return None
+    first = recorded[0]
+    if isinstance(first, list):
+        return [item for item in first if isinstance(item, dict)]
+    if isinstance(first, dict):
+        return [item for item in recorded if isinstance(item, dict)]
+    return None
+
+
+def _json_to_str(value: Any) -> str:
```
dbutenhof: The name is a bit misleading since the JSON dump is the *third* return.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -471,6 +471,107 @@ def _constrain_tool_call_body(
...
+        return value.decode()
+    if isinstance(value, str):
+        return value
+    dumped = json.dumps(value)
+    if isinstance(dumped, bytes):
+        return dumped.decode()
+    return dumped
+
+
+def _chat_messages_to_responses_input(
```
dbutenhof: Do we really need to support replaying raw chat completions as responses? Is this something inference-perf or aiperf do?

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -471,6 +471,107 @@ def _constrain_tool_call_body(
...
+
+    :param data: Current generation request.
+    :return: ``required`` or ``auto``.
+    """
+    values = data.columns.get("tool_choice_column") or []
+    if values:
+        choice = values[0]
+        if choice in ("required", "auto"):
+            return str(choice)
+    return "required"
```
dbutenhof, line 485: So anything other than "required" or "auto" is "required"?? Are there other realistic possibilities? Ignoring an unrecognized value and assuming "required" seems potentially bad ...

src/guidellm/data/deserializers/trace_otel.py
```diff
@@ -0,0 +1,952 @@
...
+def span_attributes(span: dict[str, Any]) -> dict[str, Any]:
+    """Return the span ``attributes`` mapping.
+
+    Nested JSON columns are already decoded by ``datasets``. Non-dict values
+    are treated as missing so the span is skipped rather than parsed here.
+
+    :param span: One OTEL span dict.
+    :return: Attribute mapping, or an empty dict when absent or not a dict.
+    """
+    attrs = span.get("attributes") or {}
```
dbutenhof: Not that it hurts much, but the `or {}` seems pointless when anything but a `dict` (including `None`) will already fail the `if isinstance` below and return `{}`.

src/guidellm/data/deserializers/trace_otel.py
```diff
@@ -0,0 +1,952 @@
...
+    return attrs if isinstance(attrs, dict) else {}
+
+
+def first_attribute(attributes: dict[str, Any], keys: list[str]) -> Any:
```
dbutenhof, line 66: Technically, return "an arbitrary non-null attribute value", since dict keys are unordered. More like `def an_attribute` or `non_null_attribute`; but then, the afternoon got crazy and I'm back looking at this late at night and I'm probably a bit punchy...

dbutenhof (commented):
Checkpoint ... I tried, but I'm not getting any further today and I might as well drop what I've got so far.

docs/guides/datasets.md
jaredoconnell: Shortened. This is going to cause a merge conflict with the change I made to the other PR, but that's okay.

docs/guides/trace_replay.md
jaredoconnell: I didn't add too many options to these since they're just basic examples of using huggingface as the data source. I added time scale to the trace synthetic example. I added a constraint.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: Auto and required are the only options we'd want to see. Required just means since a took call is expected, we tell the model server it must generate one. Otherwise it will fail to generate the tool call. There are tradeoffs, though. Those should be documented elsewhere.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: I removed this from this PR.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: I removed it from this PR. I don't know how important it is.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: Removed from this PR. I have it stashed locally if we find it important.

docs/guides/trace_replay.md
jaredoconnell: Done.

docs/guides/trace_replay.md
jaredoconnell: Done. The original logic is that the datasets can be really large, but your earlier PR did optimize it greatly.

src/guidellm/data/deserializers/trace_otel.py
jaredoconnell: Done.

src/guidellm/data/deserializers/trace_otel.py
jaredoconnell: Removed.

src/guidellm/data/deserializers/trace_otel.py
jaredoconnell: There's now a clarifying docstring. Let me know if that helps or not.

src/guidellm/data/deserializers/trace_otel.py
jaredoconnell: This part was speculative:
```
-    if isinstance(value, (int, float)):
-        number = float(value)
-        if abs(number) >= _NANOSECONDS:
-            return number / 1e9
-        if abs(number) >= _MILLISECONDS:
-            return number / 1e3
-        return number
```

The rest are used in the real world. I added documentation specifically mentioning which known use cases use each.

src/guidellm/data/deserializers/trace_otel.py
jaredoconnell: Yeah, that's why I mentioned that to do a pure raw column we'd need another column to describe what part is raw. I don't think it makes sense to go full `raw` for this.

src/guidellm/scheduler/dag.py
jaredoconnell: Done.

src/guidellm/scheduler/dag.py
jaredoconnell: Let me know if the latest changes address this.

src/guidellm/schemas/data/deserializers/trace_otel.py
jaredoconnell: Done.

jaredoconnell (commented):
Thank you for the reviews. I hopefully addressed everything. I also had AI go through it and remove speculative parts, since AI adds a lot of defensive speculative parts. That simplified it a bit.

docs/guides/tool_calling.md
```diff
@@ -156,7 +156,7 @@ guidellm run \
...
+**3. OTEL traces** -- tool-call turns come from the file: assistant `tool_call` parts on span output, or `gen_ai.response.finish_reasons` of `tool_calls` / `tool_call` / `tool_use` / `function_call`. Schemas come from `gen_ai.tool.definitions` when present, otherwise the default synthetic tool. Recorded tool result strings are injected and rebound to live `tool_call_id`s. Pass `tool_choice=required` (default) or `tool_choice=auto` on `--data kind=otel`. Recorded messages are chat-completions dicts; replay is `/v1/chat/completions` only for now. `/v1/completions` has no tool loop. Backend `tool_call_missing_behavior` is unchanged. See [Trace replay](trace_replay.md#otel).
```
dbutenhof: > `/v1/completions` has no tool loop.

Even with your translator, OTEL wasn't supporting text completions traces, right? This sounds particularly odd, though, now that it immediately follows "replay is `/v1/chat/completions` only for now."

docs/guides/trace_replay.md
```diff
@@ -152,15 +156,15 @@ Output length always uses the span's completion token count (`max_tokens` + `ign
...
 
 If the next span cannot be parsed as tool results, a placeholder injection is synthesized (using a following `execute_tool` span's result when present) and that next span is still replayed.
 
 Completed request stats merge response usage over the request's expected token counts, so a dedicated expected-vs-actual MAE is not reported. Compare `request.input_metrics` / `output_metrics` (span counts) with response usage before that merge if you need the deviation.
 
-ISO-8601 `start_time` values, unix seconds, milliseconds, and nanoseconds are converted to epoch seconds before scheduling. Token counts are read from span `attributes`, trying current GenAI names first and then the deprecated aliases. OTel `parts` (`text`, `tool_call`, `tool_call_response`) are converted to OpenAI chat dicts.
+ISO-8601 `start_time` values (naive, `Z`, or offset) and HuggingFace-decoded `datetime` objects are converted to epoch seconds before scheduling. Token counts are read from span `attributes`, trying current GenAI names first and then the deprecated aliases. OTel `parts` (`text`, `tool_call`, `tool_call_response`) are converted to OpenAI chat dicts.
```
dbutenhof, line 167: Naive, UTC, or timezone offset only? The parser doesn't handle general timezone syntax??

src/guidellm/backends/openai/request_handlers.py
dbutenhof: My point was more about error handling. If the dataset contains a bad value in the column, we're just ignoring it and silently choosing "required". I'd hesitate to say we should log or something, because if this is in every row it'll get noisy ... but just ignoring it, even though "it should (probably) never happen", after we've gone out of our way to check for it, seems ... odd?

src/guidellm/data/deserializers/trace_otel.py
dbutenhof: Yeah, yeah, OK -- I think I *was* a bit punchy here; you're traversing the "list of things we'd like to find" in order for the first that appears among the (unordered) dict keys, so the lack of clarity was mine. But, yeah; I think the comment helps.

dbutenhof (commented):
On a whim, I had Codex do a local code review, and it brought up a few interesting points.

First, OTEL replay only supports /v1/chat/completions, but there's no validation. I confirmed that if I specify `request_format=` either `/v1/completions` or `/v1/responses`, the GuideLLM run completes very quickly, with no output tokens (plus a request err and 407 input token errs). This is implied in the documentation, and the output statistics are certainly enough to suggest a problem... but while I'm a bit reluctant to add a cross-component validator, the Goodput PR already introduced that concept with a cross-check between profile and metrics, and we *could* add a cross-check between the replay profile and the backend request format.

A bit more subtle and arcane (and I haven't tried to verify), so I'm passing these on verbatim:

* Medium: fallback tool-response injections lose their original timing. When the next LLM span cannot be consumed as the injection, _placeholder_or_execute_results() correctly searches for later execute_tool spans, but _append_tool_loop() still stamps the synthetic injection with the tool-call span’s timestamp. In practice that collapses tool latency to zero even when the trace recorded a later tool execution, which distorts replay pacing for agentic workloads. The injection timestamp should come from the matched execute_tool span (or at least the earliest matched fallback), not the original call time. trace_otel.py (line 817) trace_otel.py (line 831)
* Medium: execute_tool attribution is based only on time windows, not span relationships, so overlapping tool activity inside one trace_id can be attached to the wrong tool loop. The code explicitly ignores parent_span_id and collects every execute_tool span between two LLM timestamps. That is fragile for concurrent subagents or nested tool execution, which is exactly where OTEL traces tend to be valuable. I’d want parent/child matching here before trusting multi-agent traces. trace_otel.py (line 425) trace_otel.py (line 841)

sjmonson:
> we could add a cross-check between the replay profile and the backend request format.

The problem here is that replay can support other request_formats for weka and mooncake. So I guess you could check if `DataArgs.type in ("otel", "opentelemetry")`.

src/guidellm/scheduler/dag.py
sjmonson: Yep!

src/guidellm/data/deserializers/trace_otel.py
```diff
@@ -0,0 +1,942 @@
...
+
+def is_execute_tool_span(span: dict[str, Any]) -> bool:
+    """Return whether ``span`` is a tool-execution span, not an LLM request.
+
+    Matches ``gen_ai.operation.name == execute_tool``.
+
+    :param span: One OTEL span dict.
+    :return: ``True`` when this span records a tool execution.
+    """
+    return span_attributes(span).get("gen_ai.operation.name") == "execute_tool"
```
sjmonson: This is only used in one place. Just inline it with a comment:

```python
        # if `span` is not a tool-execution span
        if not span_attributes(span).get("gen_ai.operation.name") == "execute_tool":
```

sjmonson (changes requested):

docs/guides/trace_replay.md
jaredoconnell: The goal was to not over-engineer this function, so to support the ones actually used instead of speculatively do all of them, which the prior commit did.

src/guidellm/data/deserializers/trace_otel.py
jaredoconnell: Done.

src/guidellm/backends/openai/request_handlers.py
jaredoconnell: I see that point. I'll switch it to raise a ValueError.

docs/guides/tool_calling.md
jaredoconnell: Oh wow. Yeah, that does seem very random. I'll remove that part.

jaredoconnell (commented):
Addressed review comments, and attempted to make the recommended improvements for tool calls.

src/guidellm/backends/openai/request_handlers.py
```diff
@@ -506,6 +511,38 @@ def _recorded_chat_messages(columns: dict[str, Any]) -> list[dict[str, Any]] | N
...
+    :param field: Body key that must be non-empty (``prompt``, ``messages``,
+        ``input``).
+    :param sources: Human-readable column names that can fill ``field``.
+    :param columns: Request columns, listed in the error when present.
+    :raises ValueError: If ``field`` is missing or empty.
+    """
+    body = arguments.body or {}
+    if body.get(field):
+        return
+    present = sorted(key for key, value in columns.items() if value)
```
dbutenhof: According to the docstring, the `if value` is correct because you really want "non-empty" whereas the *name* here implies that the test ought to be `if Value is not None`. I'm assuming the implementation makes sense, but the `present` name is misleading and arguably should be changed.

This applies to the error message, too, because you could be omitting fields that were actually *present* in the row, but falsey.

dbutenhof (approved):
I'm going to approve, although I'm making one minor naming comment here "for your consideration".

sjmonson (approved):

PR #1155, opened 2026-09-16
Implement `copies` feature for synthetic traces

jaredoconnell:
## Summary

This allows you to increase the apparent size of the dataset for long-running benchmarks.

## Details

- Creates a new faker with a new salt for each iteration of the dataset
- Adds each new copy after the prior copy (including the original) ends
- The parallelism option is still the option that should be used to increase the parallelism of the dataset, but if your benchmark runs so long that it runs out, you can use the new `copy` option.

## Test Plan

You can test with all synthetic trace formats. Here is a command I used to test it, along with the attached file I used.
`guidellm run     --backend "kind=openai_http,target=http://localhost:8000,request_format=/v1/chat/completions,validate_backend=false,model=Qwen/Qwen3-0.6B"     --tokenizer "kind=huggingface_auto,model=Qwen/Qwen3-0.6B"     --profile kind=replay --data "kind=trace_synthetic,source.kind=json_file,source.path=replay_stall_delayed_requests.jsonl,max_wait=1,max_session_wait=1,copies=4,min_concurrent_sessions=4" --constraint kind=max_duration,seconds=60`

```jsonl
{"timestamp": 0.0, "input_length": 16, "output_length": 8}
{"timestamp": 0.5, "input_length": 16, "output_length": 8}
{"timestamp": 60.0, "input_length": 16, "output_length": 8}
{"timestamp": 60.5, "input_length": 16, "output_length": 8}
{"timestamp": 61.0, "input_length": 16, "output_length": 8}
{"timestamp": 65.0, "input_length": 16, "output_length": 8}
{"timestamp": 70.0, "input_length": 16, "output_length": 8}

```

## Related Issues

- Resolves #

---

## Use of AI

---

# git log

commit 53518f1013eb92e3390a7228790deb405c47b0b3
Date:   Wed Sep 16 18:06:23 2026 -0400

    Implement `copies` feature for synthetic traces
    
    This allows you to increase the apparent size of the dataset for long-running benchmarks.
    
commit 75d05ea0f0d4c132d29895b61fa6dbffc2af9377
Date:   Wed Sep 16 18:43:08 2026 -0400

    Fix min_concurrent_sessions flattening all requests in trace_synthetic
    
commit 9d97329ed8b63cc3c16dbce5aad0438873c0ef42
Date:   Fri Sep 18 21:47:35 2026 -0400

    Preserve relative timestamps of conversations in WEKA traces
    
commit ee4947cc4ff5074d2a4b3c8198f35c9db19d06f7
Date:   Fri Sep 18 21:54:29 2026 -0400

    Add copy offset setting
    
commit 84bcabf913136f067999114280824bf1c8624140
Date:   Fri Sep 18 22:28:24 2026 -0400

    Address review comments
    
---------


docs/guides/datasets.md
```diff
@@ -20,7 +20,7 @@ The following arguments configure datasets and their processing:
...
   - `json_file`, `csv_file`, `text_file`, `parquet_file`, `arrow_file`, `hdf5_file`, `db_file`, `tar_file` — loads from a local file. Required field: `path`.
-  - `trace_synthetic`, `mooncake`, `weka` — replay trace data with `--profile kind=replay`. Required field: `source` (a nested dataset config such as `json_file` or `huggingface`), see other supported sources for details. Optional: `timestamp_column`, `prompt_tokens_column`, `output_tokens_column`, `time_scale`, and other format-specific options. `weka` uses different column defaults (`t`, `in`, `out`). See [Trace File Formats](./trace_replay.md).
+  - `trace_synthetic`, `mooncake`, `weka` — replay trace data with `--profile kind=replay`. Required field: `source` (a nested dataset config such as `json_file` or `huggingface`), see other supported sources for details. Optional: `timestamp_column`, `prompt_tokens_column`, `output_tokens_column`, `time_scale`, `copies`, and other format-specific options. `weka` uses different column defaults (`t`, `in`, `out`). See [Trace File Formats](./trace_replay.md).
```
dbutenhof: Not new, but the "weka uses different column defaults" sounds odd here, since we haven't yet mentioned any other format's column defaults. I think, at this point, sometime more generic like "see [Trace File Formats](...) for more information on replay datasets."??

src/guidellm/data/deserializers/trace_common.py
```diff
@@ -158,6 +163,26 @@ def fill_hash_id_table(
             sibling_token_blocks[prev_id].add(block)
 
 
+def copy_faker_seed(random_seed: int, copy_index: int) -> int:
```
dbutenhof: 1. Called only from `_seeded_faker` ... does it need to be a separate method?
2. I always look with some suspicion on inline literal values ... can we conceive of any reason someone might want a different stride? Seems unlikely to ever matter, but ...
3. If this is only called from a "private" `_` routine, maybe it should also be an `_` routine?

src/guidellm/data/deserializers/trace_common.py
```diff
@@ -169,6 +194,15 @@ def __iter__(self) -> Iterable[Dataset]:
     def reset(self) -> None:
         pass
 
+    def set_copy_index(self, copy_index: int) -> None:  # noqa: ARG002
```
dbutenhof: Why the `ARG002` override? I verified that nothing's complaining, for me, without it.

(Although I do kinda think that this and some others should likely be marked `@abstract` or at least `...`, our checkers don't seem to care... which actually surprised me a bit.)

docs/guides/trace_replay.md
```diff
@@ -67,7 +68,9 @@ guidellm run \
...
+`trace_synthetic` and `mooncake` replay each row as an independent, single-request conversation. Rows are sorted by timestamp and keep their offsets from the first request in the trace. Prompts are generated as rows are consumed, and Mooncake hash IDs remain shared across rows within one `copies` pass. Use `max_session_wait` to cap gaps between these independent requests; `max_wait` only caps gaps within multi-request conversations, such as WEKA sessions.
+
+Raise parallelism with `min_concurrent_sessions`, wait caps, and `time_scale` first. Use `copies` only when that packed pass is too short for the benchmark (`max_duration` / `max_requests`). `copies` replays the entire packed dataset back-to-back; it does not run duplicate conversations at the same timestamp. The next pass's first request is scheduled at the previous pass's last request timestamp (timeline concatenation, not a wait for the last generated token). Hash-id formats (`mooncake`, `weka`) use a separately salted global token-block table per pass so later passes do not reuse earlier tokens and inflate prefix-cache hits.
```
dbutenhof: > The next pass's first request is scheduled at the previous pass's last request timestamp

So essentially there's no gap between copies ... one copy begins scheduling immediately after the previous ends. (Not simultaneous with the last event simply because we have to go back for the next iteration.) Makes me wonder if there should be a configurable gap?

src/guidellm/data/deserializers/trace_common.py
```diff
@@ -286,36 +320,48 @@ def __init__(
...
-                samples_count,
-                {
-                    "conversation_turns": (
-                        payload.decode() if isinstance(payload, bytes) else payload
-                    )
-                },
+        last_end = 0.0
+        for copy_index in range(self.config.copies):
+            self.format.set_copy_index(copy_index)
+            copy_faker = self._copy_fakers[copy_index]
```
dbutenhof: Minor nit, but "copy_faker" sounds like an action phrase (which normally signals a *call*), rather than a passive noun (variable)... "faker_copy" would read more naturally.

src/guidellm/data/deserializers/trace_common.py
```diff
@@ -286,36 +320,48 @@ def __init__(
...
+                    samples_count,
+                    {
+                        "conversation_turns": (
+                            payload.decode() if isinstance(payload, bytes) else payload
+                        )
+                    },
+                )
+                self.format.reset()
+                pass_end = graph_max_timestamp(graph_data)
+                if pass_end is not None:
```
dbutenhof: Why would a replay graph have no turns with timestamps (because that's what this means) and do we really just want to leave `last_end` where it was (meaning the next copy, if any, will overlap, as we said they won't)?

src/guidellm/data/deserializers/trace_mooncake.py
```diff
@@ -66,9 +66,23 @@ class MooncakeTraceFormat(SingleTurnTraceFormat):
     def __init__(self, config: MooncakeTraceFormatArgs, dataset: Dataset) -> None:
         self.config = config
         self.dataset = dataset
+        self._copy_index = 0
+        self._hash_id_tables: list[dict[int, tuple[int, ...]]] = [
+            {} for _ in range(config.copies)
+        ]
+        self._sibling_tables: list[dict[Any, set[tuple[int, ...]]]] = [
+            {} for _ in range(config.copies)
+        ]
```
dbutenhof: Not really an issue as long as the copy count is relatively small -- but do we really need to pre-allocate the entire set instead of just re-allocating as we iterate from one copy to the next?

dbutenhof (commented):
Just comments, really -- nothing's a blocker; but I'll hold off a bit on approving just so in case Sam approves we don't auto-merge before you see the comments/questions here. 😁

src/guidellm/data/deserializers/trace_common.py
sjmonson: I'm in agreement on (1), does not need to be its own function. (2) I don't think we ever need a different stride. This is basically what we do in `synthetic_text` though I don't think it really matters to use a big prime.

src/guidellm/data/deserializers/trace_common.py
```diff
@@ -286,36 +320,48 @@ def __init__(
...
+            # Fresh packing state per copy so inner min_concurrent_sessions
+            # does not carry session-end lanes across sequential passes.
+            timing = TraceSessionTiming(
+                max_wait=self.config.max_wait,
+                max_session_wait=self.config.max_session_wait,
+                min_concurrent_sessions=self.config.min_concurrent_sessions,
+                time_scale=self.config.time_scale,
             )
-            self.format.reset()
+            pass_offset = 0.0 if copy_index == 0 else last_end
```
sjmonson: Won't this always shift the copied conversation? We definitely do not want to do that. The whole point of copying conversations is to put more concurrent load on the server.

sjmonson (changes requested):

src/guidellm/data/deserializers/trace_common.py
jaredoconnell: Correct, but then you can shift multiple conversations so that they run in parallel. I think there's value in not using duplicate, only hash-salt-shifted, data at the same time. The logic is to use `min_concurrent_sessions` so that you have other conversations shifted earlier. But I could see the argument either way. I'll make it configurable. I have an idea. An offset value, where 0 puts the following one at the beginning of the prior one. 1 puts it exactly at the end. 1.1 adds a tiny gap between them. 0.5 means half overlap. I think this option has a lot of potential.

sjmonson: For reference: we met and discussed this. The conclusion was either leave it as is or use the proposed "offset" value.

jaredoconnell: I checked and it does already act the way I originally intended for it to work, with it shifting relative to the entire dataset. So it should stack the data as needed for long benchmarks.

I implemented the copy offset, though. It just scales the offset.

docs/guides/trace_replay.md
jaredoconnell: With the offset setting, it allows values greater than 1 for a gap. But I don't think a gap is that important.

docs/guides/datasets.md
jaredoconnell: I removed it, but kept the link to the trace file format for more details.

src/guidellm/data/deserializers/trace_common.py
jaredoconnell: I inlined this now. That removed a lot of boilerplate.

src/guidellm/data/deserializers/trace_common.py
jaredoconnell: Removed the QA override.

src/guidellm/data/deserializers/trace_common.py
jaredoconnell: Renamed.

src/guidellm/data/deserializers/trace_common.py
jaredoconnell: Yeah, I think this is too much defensiveness. Removed a lot of the no timestamp checks.

src/guidellm/data/deserializers/trace_mooncake.py
jaredoconnell: Changed to not pre-allocate.

src/guidellm/data/deserializers/trace_mooncake.py
```diff
@@ -66,23 +66,16 @@ class MooncakeTraceFormat(SingleTurnTraceFormat):
...
-        ]
+        self._hash_id_table: dict[int, tuple[int, ...]] = {}
+        self._sibling_table: dict[Any, set[tuple[int, ...]]] = {}
 
-    def set_copy_index(self, copy_index: int) -> None:
-        """Select the hash-table slot for a sequential dataset copy.
-
-        Index 0 is the original global table. Each later copy has its own
-        independently salted slot in the same lists.
+    def set_copy_index(self, copy_index: int) -> None:  # noqa: ARG002
```
dbutenhof, line 72: Since you're no longer pre-allocating all the hash tables and indexing through, is there any need for the `copy_index` parameter? I'd think this would be better now as maybe "reset_hash_tables` without the parameter ... which would also remove the annoying noqa.

src/guidellm/data/deserializers/trace_mooncake.py
```diff
@@ -109,8 +102,8 @@ def create_prompt(
         """Before generating the prompt, this first generates a block of tokens for
         each hash ID that has not already been seen."""
         ids = row[self.config.hash_ids_column]
-        hash_id_table = self._hash_id_tables[self._copy_index]
-        sibling_token_blocks = self._sibling_tables[self._copy_index]
+        hash_id_table = self._hash_id_table
+        sibling_token_blocks = self._sibling_table
```
dbutenhof, line 106: Is there really a reason for the local pointers now? You're referencing the first twice, but I doubt that's really worth the extra local variable.

dbutenhof (commented):
Cleaner ... a couple of minor follow-ups on the hash table rework.

src/guidellm/data/deserializers/trace_mooncake.py
jaredoconnell: Removed.

src/guidellm/data/deserializers/trace_mooncake.py
jaredoconnell: Good catch. Removed.

dbutenhof (dismissed):

dbutenhof (approved):

sjmonson (approved):

PR #1159, opened 2026-09-17
Pass a backend's per-request metrics through to the compiled stats

wenxuan-elastix:
## Summary

vLLM 0.29 can report a request's own server-side timings and speculative-decoding
acceptance on the response, in a `metrics` object next to `usage`. GuideLLM drops it:
`extract_choices_and_usage` reads only `choices` and `usage`, so the object is gone
before a `GenerationResponse` is built and never reaches the report.

That matters most for speculative decoding. Today the only way to get acceptance out of
a GuideLLM run is to scrape `/metrics` separately and difference two counter snapshots
taken around it. Those counters are server-wide and cumulative, so the delta also counts
anything else that hit the endpoint during the run, and a per-engine counter reset can
hide inside the sum. The numbers on the response describe exactly the requests GuideLLM
made, and arrive with them.

This PR carries that object through unchanged, as `response_metrics`.

What the server sends when started with `--per-request-spec-decode-metrics summary`:

```json
{
  "choices": [ ... ],
  "usage": { ... },
  "metrics": {
    "time_to_first_token_ms": 12.5,
    "speculative_decoding": {
      "mean_acceptance_length": 1.23,
      "draft_acceptance_rate": 0.078,
      "acceptance_histogram": [39, 1, 0, 3],
      "num_spec_steps": 43,
      "num_accepted_draft_tokens": 10,
      "num_draft_tokens": 129,
      "num_spec_tokens": 3
    }
  }
}
```

In streaming it rides the closing usage chunk, which GuideLLM already requests via
`stream_options.include_usage`, so nothing about how requests are sent changes.
Reference: https://docs.vllm.ai/en/v0.29.0/features/speculative_decoding/acceptance_metrics/

## Details

- [x] Add `GenerationResponse.response_metrics: dict[str, Any] | None`, and the same
      field on `GenerativeRequestStats`, passed through by `compile_stats`.
- [x] Record the object in `TextCompletionsRequestHandler.extract_choices_and_usage`.
      That is the one place every completions handler already reads the raw response
      dict, on both the streaming and the non-streaming path, so recording it there
      rather than returning it keeps that method's signature as it is.
- [x] Thread it through `_compile_streaming_response` for the chat handler.

Notes on the choices made:

- **Passed through, not modelled.** The contents belong to the backend, and vLLM marks
  this shape experimental, so GuideLLM validating it would just be a second thing to
  keep in sync. A `dict` costs nothing and stays correct when the shape moves.
- **Which handlers.** Completions and chat completions populate it, and the audio and
  pooling handlers inherit that. The `/v1/responses` handler does not read the field and
  keeps the default `None`.
- **Sampling.** `clear_stats_data` clears `request_args`, `output`, `reasoning_output`
  and `tool_calls` on non-sampled requests; it does not touch metrics, so a non-sampled
  request keeps `response_metrics` the same way it keeps its usage.
- **Backends that report nothing** are unaffected: the field stays `None`.

## Test Plan

`TestResponseMetricsPassthrough` in `tests/unit/backends/openai/test_request_handlers.py`
covers the three cases:

- non-streaming: the object on the response reaches `GenerationResponse.response_metrics`
- streaming: the object on the closing usage event reaches it
- a backend that reports no `metrics` leaves the field `None`

```
pytest tests/unit/backends/openai/test_request_handlers.py -k ResponseMetricsPassthrough
```

Ran against `main` as a baseline and with this change, over
`tests/unit/backends/openai/test_request_handlers.py`, `tests/unit/schemas` and
`tests/unit/benchmark`: same failures before and after, plus the three new tests passing.
`ruff check` reports the same findings as `main` over the changed files.

## Related Issues

None open that I could find; happy to file one if you would rather track it separately.

---

## Use of AI

---

# git log

commit 21df4897a5cf7ce461c35f8a359390fee528f3b8
Author: wenxuan-elastix <wenxuan@elastix.ai>
Date:   Thu Sep 17 16:03:41 2026 -0700

    Pass a backend's per-request metrics through to the compiled stats
    
    vLLM reports per-request server-side timings and speculative-decoding
    acceptance in a `metrics` object beside `usage`, served behind
    `--per-request-spec-decode-metrics` since 0.29. `extract_choices_and_usage`
    reads only `choices` and `usage`, so the object is dropped before a
    `GenerationResponse` is built and never reaches the report.
    
    Record it on the handler where the raw response is already read, and carry
    it on `GenerationResponse.response_metrics` into `GenerativeRequestStats`.
    It is passed through as received rather than modelled, since its contents
    are the backend's own and vLLM marks the shape experimental.
    
    The completions and chat-completions handlers populate it; the responses
    handler leaves the default. `clear_stats_data` does not touch it, so a
    non-sampled request keeps its metrics like it keeps its usage.
    
    Signed-off-by: wenxuan-elastix <wenxuan@elastix.ai>

---------

Signed-off-by: wenxuan-elastix <wenxuan@elastix.ai>


sjmonson (approved):
Yeah seems reasonable. Adding summary metrics for this was on my TODO list for v0.9.0 but is somewhat blocked by refactoring the metrics handler to better support optional metrics. This is good for now since at least the user can get the raw data.

wenxuan-elastix:
@mergifyio queue

wenxuan-elastix:
@sjmonson thanks! I don't have write access, so `@mergifyio queue` is rejected on my end — could you queue it, or should I just wait for a second approval?

dbutenhof (approved):
OK -- this is relatively simple, and appears to work when I run vLLM with `--enable-per-request-metrics`.

Issue #243, opened 2025-07-27
Margin Of Errors (MOE) in output report


AlonKellner-RedHat:
Currently, the output report includes many aggregated measured metrics, essentially the following matrix:
| \ | µ | σ | min | max | p001 | p01 | p05 | p10 | p25 | p50 | p75 | p90 | 095 | p99 | p999 |
| :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| E2E |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |
| TTFT |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |
| ITL |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |
| TPUT |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |

This matrix has one big flaw - it does not include any measure of reliability for the measurements.
For example, given the values of two identical benchmarks, but the first one ran for 10 seconds, and the second one ran for 10 minutes. The statistics measured in the second benchmark are much more reliable, because it has x60 more samples.

Since this is basically an estimation task, we should probably calculate a MOE value (e.g ±3.3%) for each measured metric in the matrix, it's a well accepted, simple and interpretable reliability measure.

## Algorithmic Requirements
There are many reasons for measurements to be unreliable, that should reflect in the MOE values.
Here are the 4 top ones:
1. Amount of samples
2. STD of the measured metric
3. Precision of the RPS (scheduling discrepancies)
4. Out of Distribution samples (anomalies / temporal interferences)

## Proposed Algorithm
Here is the outline of an algorithm that achieves that MOE estimation given all of the requirements:
1. Define a weight per sample, all samples start with weight 1
2. Multiply the weights by a RPS precision penalty, using a gaussian similarity (`RPS_penalty=gaussian(target_RPS-acutal_RPS)`)
3. Multiply the weights by an Out of Distribution penalty, using autocorrelation (`OOD_penalty=mutual_information(current_sample, adjacent_sample)`)
4. Then, proceed to calculate MOE normally with a t-student distribution (given the weights).
That way, samples with non precise RPS, or Out of Distribution samples with temporal correlation will be discounted from the final estimation.

Here are a few cases that the proposed algorithm was designed to handle:
[...]

PR #1163, opened 2026-09-19
Add confidence intervals for request-level metrics

QHarshil:
## Summary

Every number in a benchmark report is a point estimate with no statement of how precisely it was measured. That matters most for the percentiles, which are order statistics: `Percentiles.from_pdf` puts p99 at rank `ceil(0.99n)`, so below 100 successful requests the reported p99 is the slowest single request. I confirmed this on a ten second run: n=71, reported p99 128.42 ms, max 128.42 ms. Across twelve identical runs of that length the reported p99 ranged 115.83 to 146.66 ms, a 23.5% spread, while the mean moved 8.5%.

This adds a confidence interval for the mean and for each percentile, for supported unweighted request-level distributions only. Token-weighted and derived rate distributions stay point estimates, for the reason below.

## Details

- [x] Add `guidellm/utils/statistics.py` with `normal_ppf`, an accurate `t_ppf`, `wilson_interval`, `mean_confidence_interval` and `quantile_confidence_interval`. `wilson_interval` moves here from `goodput.py` and `approx_t_ppf` from `saturation.py`, which removes the cross-layer import at `goodput.py:18`.
- [x] Add `ConfidenceInterval`, `PercentileIntervals` and `SampleUncertainty`, and `mean_ci` / `percentile_cis` on `DistributionSummary`. Both are nullable, so reports written before this load unchanged.
- [x] Add `confidence` to `--metrics kind=generative`, default 0.95, `null` to disable. Recorded once per benchmark as `config.confidence`.
- [x] Attach intervals at the eleven compile sites that record one unweighted value per request. Time per output token, inter-token latency and the derived rate distributions are excluded, and the estimator independently rejects any weight other than one.
- [x] Console shows the mean with the half width of its interval, for example `84.1 ±1.7`, and marks a percentile the sample cannot bound with `*` plus a footnote.
- [x] CSV appends `Mean CI`, `Percentile CIs` and the confidence level at the end of each row. Verified against `main`: all 238 pre-existing columns keep their exact position.

Two behavioural notes:

- **No change to existing metric values or scheduler and profile decisions.** `wilson_interval` keeps its current arithmetic rather than moving to the exact normal quantile. The two differ by 2.2e-4, which moves a Wilson upper bound of exactly 1.0 to just below it and would turn a goodput probe meeting a 100% attainment target into one that missed it. Five bounds are pinned by exact equality against `main`. Console and CSV output do change. Correcting the over-saturation detector's quantile is left as a separate deliberate change.
- **Percentile intervals are absent by design on short runs.** Below the sample size that supports two order statistics the field is `null` rather than a bound the observations cannot support: at the default 95% confidence level, 72 requests for p95, 368 for p99 and 3688 for p999.

## Notes for review

At 100k observations, uncertainty compilation adds about 13.5 ms per distribution on my machine, or about 0.15 s across the eleven enabled metrics, once during final report compilation.

#1101 and #1102 also touch `benchmark/schemas/metrics.py` and its tests. The regions differ, but whichever lands second will need a rebase.

## Test Plan

- `tox -e lint-check`, `tox -e type-check`: pass.
- `tox -e test-unit`: 3382 passed. The 14 `tests/unit/utils/test_audio.py` failures are a missing local FFmpeg; `python -c "import torchcodec"` fails identically in the same environment with no GuideLLM code involved.
- `tox -e test-integration`: 28 passed, 25 xfailed.
- `t_ppf` is checked against published Student's t tables for df 1 to 1000 at five confidence levels, plus symmetry, monotonicity and domain errors. Both interval estimators are checked for coverage over repeated samples.
- Each guard was checked by reverting it and confirming a test fails, so none of the new coverage is vacuous.
- End to end against the mock server, 438 requests, 0 errored: p95 `[109.56, 115.92]`, p99 `[119.13, 142.08]`, p999 `null`.

## Related Issues

- Addresses #243. Inter-token latency and time per output token are left as point estimates for now, for the reason above; I would like to handle those separately.

---

## Use of AI

---

# git log

commit fe087289d3b2b201286223d76e41c78ab631bd00
Date:   Fri Sep 18 18:48:45 2026 -0700

    Report confidence intervals for request-level metrics
    
    Every statistic in a report is an estimate from a finite number of requests,
    and nothing in the output says how precisely it was measured. The percentiles
    are order statistics, so p99 sits at rank ceil(0.99n). Below 100 successful
    requests that rank is n, and the reported p99 is the slowest single request,
    printed with the same authority as the mean. Across twelve identical ten
    second runs against the mock server the reported p99 varied by 23.5% while
    the mean varied by 8.5%.
    
    DistributionSummary now carries an interval for the mean and one for each
    percentile, at a level set through --metrics and recorded once per benchmark.
    A percentile interval is null where the sample cannot place two order
    statistics around it, which takes 72 requests for p95, 368 for p99 and 3688
    for p999. Reporting null there is the point: it marks the percentiles a short
    run cannot support. The console marks such a percentile rather than printing
    it as though it were measured.
    
    Intervals are attached only to distributions holding one unweighted
    observation per request. Time per output token and inter-token latency stay
    point estimates because, in a repeated-run variable-load experiment, their
    within-run intervals understated observed run-to-run variation by 1.8x and
    3.2x respectively. The measurements were associated with a shared run-level
    load condition that a single-run request-sampling interval cannot observe.
    Derived rate distributions are excluded separately: their time-weighted
    construction does not meet the estimator's request-sample assumption, and
    that construction is itself under discussion in #602. The estimator rejects
    any weight other than one, so a later caller cannot opt an unsupported
    distribution into inference.
    
    The quantile functions and interval estimators move to a shared module,
    which removes an import of approx_t_ppf from a scheduler constraint into a
    benchmark profile. approx_t_ppf moves unchanged with a re-export left behind,
    and wilson_interval keeps deriving its quantile from it rather than from the
    exact normal quantile. The two differ by 2.2e-4, enough to move a Wilson
    upper bound of exactly 1.0 to just below it and turn a goodput probe that met
    a 100% attainment target into one that missed it. New inference uses an
    accurate t quantile, since the approximation is 49% low at one degree of
    freedom and 6% low at three.
    
---------


dbutenhof (commented):
I like this, and it seems to work; however I'm uncomfortable with the model of dynamically shuffling CSV columns. While retaining column compatibility is a good thing, I'm not sure this is the best approach.

QHarshil:
Replaced the reordering with a separate section written after all existing columns, listing only the metrics that can carry an interval. That also cuts the CSV addition from 95 columns to 27. The old version wrote interval columns for every distribution, and 68 were always empty.

Rebased onto main for the docs move in #1170, hence the force-push. The CSV change is its own commit.

dbutenhof (approved):

PR #1172, opened 2026-09-22
fix: preserve resize dimensions for image URLs

git-jxj:
## Summary

Image URLs encoded as base64 silently ignored explicit `width` and `height` options. For example, an 80×40 image requested with `width=40` stayed 80×40, while the same image supplied as bytes became 40×20. Forward both dimensions after downloading the image so URL and local inputs use the same resizing behavior.

## Details

- [x] Preserve explicit width and height when processing downloaded images.
- [x] Cover width-only, height-only, both dimensions, and maximum-size constraints, checking decoded pixels and metadata against local input.

## Test Plan

- New regression cases: 3 failed / 2 passed before the fix; all 5 pass after.
- `tox -e tests -- tests/unit`: 3,273 passed, 31 skipped, 163 xfailed.
- `tox -e tests -- tests/integration tests/e2e`: 42 passed, 25 xfailed, using the in-tree mock server.
- `tox -e lint-check` and `tox -e type-check`: passed.
- A local HTTP server serving a real PNG reproduced the failure; decoded output dimensions are correct after the fix for width-only, height-only, and both dimensions. No real model/GPU benchmark was run.
- Suites emit existing dependency/fork warnings; the unit suite also reports semaphore cleanup at interpreter shutdown.

## Related Issues

Independent code inspection; no associated issue.

---

## Use of AI

Code and regression tests were generated with OpenAI Codex.

---

# git log

commit 82cdb1230afbb7a95dea2d2cb19fd319b1f7e717
Author: git-jxj <65210887+git-jxj@users.noreply.github.com>
Date:   Tue Sep 22 17:09:01 2026 +0800

    fix: preserve resize dimensions for image URLs
    
    Generated-by: OpenAI Codex
    Signed-off-by: git-jxj <65210887+git-jxj@users.noreply.github.com>

---------

Generated-by: OpenAI Codex
Signed-off-by: git-jxj <65210887+git-jxj@users.noreply.github.com>


src/guidellm/utils/vision.py
```diff
@@ -47,6 +47,7 @@ def encode_image(
     - PIL.Image.Image: Pillow image
     - datasets.Image: HuggingFace datasets Image object
 
+    width and height: resize dimensions, including downloaded URL images
```
dbutenhof: This function doesn't follow the `:param <name>:` docstring format. That's not your fault; but adding two parameters on one line definitely doesn't follow any parameter definition convention. Given the context, I'll let this go with a comment; but you could "take this opportunity" to fix the docstring to follow the `:param <name>: <desc>` and `:return: <desc>` style.

dbutenhof (approved):

jaredoconnell (dismissed):

src/guidellm/utils/vision.py
git-jxj: Thanks for the suggestion. I updated `encode_image` in commit `a29f5eb` to document each argument separately in the repository's reStructuredText style, and added the return and error descriptions. `uv run tox -e lint-check` passes locally; checks for the new head are queued. Could you confirm this addresses the comment and resolve the thread, then re-review the updated head when convenient?

src/guidellm/utils/vision.py
```diff
@@ -38,28 +38,26 @@ def encode_image(
...
+
+    :param image: Image input to encode.
+    :param width: Requested output width in pixels. Applies to downloaded URL
+        images when ``encode_type`` is ``"base64"``.
+    :param height: Requested output height in pixels. Applies to downloaded URL
+        images when ``encode_type`` is ``"base64"``.
+    :param max_size: Maximum length in pixels of the longest image edge.
+    :param max_width: Maximum output width in pixels.
+    :param max_height: Maximum output height in pixels.
+    :param encode_type: Output representation. ``"base64"`` encodes the image,
```
dbutenhof: Actually, *if* the input is a URL, `encode_type == "base64"` causes the function to download and process the image, returning the encoded BASE64 string while *any other value* just returns the URL. If the input isn't a URL, `encode_type` is ignored, so `None` has no special value.

I think I would say here that "if the `image` parameter is a URL string, then `encode_type` of `"base64"` downloads and processes the image file."

src/guidellm/utils/vision.py
```diff
@@ -38,28 +38,26 @@ def encode_image(
...
+        ``"url"`` returns the URL unchanged, and ``None`` selects a supported
+        representation for the input.
+    :return: A mapping with the image representation type, encoded image, and
```
dbutenhof, line 58: "optional" is generally used to describe inputs. I'd say "pixel and byte counts if known" -- or perhaps even "if the return type isn't `image_url`" (where they're not know, because we didn't download the image).

dbutenhof (changes requested):
My apologies.

I think I'm sorry for making the comment about the docstring, because once a PR touches human-targeted text, it's hard not to devolve into nit-picking.

But, with that said, since you took up the challenge (thanks for that), I think a few details should be clarified...

src/guidellm/utils/vision.py
git-jxj: Updated in `9b1a8c0`. The docstring now explains that `encode_type` controls URL inputs only: `base64` downloads and processes the image, while other values return the URL unless resize options are set. For non-URL inputs, this parameter is ignored and the image is encoded as base64. Please review the updated wording and resolve this thread if it addresses the comment.

src/guidellm/utils/vision.py
git-jxj: Updated in `9b1a8c0`: the return description says pixel and byte counts are included when known, and are `None` for an unprocessed URL. URL inputs encoded as base64 and all non-URL inputs are processed and include both counts. Please review the updated wording and resolve this thread if it addresses the comment.

dbutenhof (approved):

